PR #522, opened 2014-08-02
Add ability to get hydrogen bond lists from baker_hubbard.

leeping:
I tried the baker_hubbard criterion to get a listing of hydrogen bonds, and in some ways it's better than the wernet_nilsson criterion (the latter identifies more H-bonds, such that I end up with a mean that's higher than the normal by about 0.2).

This PR doesn't change the way `baker_hubbard` works but moves most of its code into `baker_hubbard_internal`.  The `baker_hubbard_list` function does what I want (i.e. get a list of hydrogen bonds for each frame).

I also fixed a bug where it would crash if there are no NH hydrogen bond donors (the case for water).

Thanks,
- Lee-Ping


ctk3b:
Would anyone be opposed to reviving this? I've had a few instances where I had to reach for other tools to get lists of hydrogen bonds.

ctk3b:
Related: https://github.com/mdtraj/mdtraj/issues/705

rmcgibbo:
Closing this because it's outdated.

PR #550, opened 2014-08-29
Fix some compiler warnings.

rmcgibbo:
It would be nice to port more of the cython code to use typed memoryviews instead of `cdef`d numpy arrays, too.


mpharrigan:
Can this branch be deleted?

PR #639, opened 2014-10-28
initial draft of lammpstrj file format

ctk3b:


janash:
Hi all, I realize this was merged quite a while ago, but we are trying to track down some changes we're seeing in MDTraj resulting from NumPy 2.0. Is there a reason we are using `np.float32` as the data type in `LAMMPSTrajectory`  (for example, [here](https://github.com/mdtraj/mdtraj/blob/5e2b85ccb1aa79591c87003f56a78b3c808b15d7/MDTraj/formats/lammpstrj.py#L246) ) instead of `np.float64`? @ctk3b

PR #679, opened 2015-01-08
Include nitrogen atoms as h-bond acceptors

gkiss:


rmcgibbo:
The docstrings for the two methods say that the only acceptors considered are "O", so they should be changed. And [this test](https://github.com/mdtraj/mdtraj/blob/master/MDTraj/geometry/tests/test_hbonds.py#L146) is failing, because it's now matching a nitrogen.

kyleabeauchamp:
So I think the combination of this and #681 have addressed all the issues, and tests are passing on #681.  I'm going to merge both and hope it works.

PR #684, opened 2015-01-08
PEP440 version number

rmcgibbo:
Changed 1.2.X to 1.3.0.dev0, which should be more appropriate now that pip has switched to following [PEP440](https://www.python.org/dev/peps/pep-0440/#developmental-releases).


kyleabeauchamp:
lgtm

Issue #695, opened 2015-01-27
Sidechain and backbone selection not working


synapticarbors:
Maybe I'm misunderstanding something about the selection DSL, but I think that the `sidechain` and `backbone` keywords are not working properly:

```python
In [3]: import mdtraj as md
In [4]: from mdtraj.testing import get_fn

In [5]: t = md.load(get_fn('2EQQ.pdb'))
In [6]: t.atom_slice(t.top.select('backbone'))
Out[6]: <mdtraj.Trajectory with 20 frames, 423 atoms, 28 residues, without unitcells at 0x10a74a390>

In [7]: t.top.select('backbone').shape
Out[7]: (423,)

In [8]: t.top.select('all').shape
Out[8]: (423,)

In [9]: t.top.select('sidechain').shape
Out[9]: (423,)

In [10]: t.top.select('name C or name CA or name N or name O').shape
Out[10]: (112,)
```

It appears as if with either keyword, you're just getting back all of the atom indices. I'm running this off of master but it was also in the 1.2 release on Omnia. I don't believe there are any tests that would have caught this.

kyleabeauchamp:
Looks like a bug to me

synapticarbors:
The following line also looks suspect:

https://github.com/mdtraj/mdtraj/blob/master/MDTraj/core/selection.py#L107

It appears as if `is_sidechain` is being mapped to the same expression as `is_backbone`.

kyleabeauchamp:
Looks like the culprit

synapticarbors:
That's definitely one problem

```
In [20]: t.topology.select_expression('backbone')
Out[20]: '[atom.index for atom in topology.atoms if atom.is_backbone]'

In [21]: t.topology.select_expression('sidechain')
Out[21]: '[atom.index for atom in topology.atoms if atom.is_backbone]'
```

but that's not the whole thing as:

```
[atom.index for atom in topology.atoms if atom.is_backbone]
```

also give you all of them atoms instead of just the backbone ones.

synapticarbors:
Something is going wrong in the list comprehension that I'm not entirely understanding:

```
In [28]: [(atom.index, atom.is_backbone()) for atom in t.topology.atoms if atom.is_backbone]
Out[28]:
[(0, True),
 (1, True),
 (2, True),
 (3, True),
 (4, False),
 (5, False),
 (6, False),
 (7, False),
 (8, False),
 (9, False),
 (10, False),
 (11, False),
 (12, False),
[...]

PR #696, opened 2015-01-27
Fixed bugs in backbone and sidechain selection

synapticarbors:
This PR fixes the bugs reported in https://github.com/mdtraj/mdtraj/issues/695

Also, when I tried to install mdtraj from source using `python setup.py develop` in a conda environment, I was getting a message that `mdtraj` was not installed, even after the install process looked ok. Do other devs use this method of working on mdtraj? This is how I've worked with MSMBuilder recently and there everything works as expected.

### DIFF
--- a/MDTraj/core/selection.py
+++ b/MDTraj/core/selection.py
@@ -104,7 +104,7 @@ class SelectionKeyword(object):
         (('all', 'everything'), ast.Name(id='True', ctx=ast.Load())),
         (('none', 'nothing'), ast.Name(id='False', ctx=ast.Load())),
         (('backbone', 'is_backbone'), _chain('is_backbone')),
-        (('sidechain', 'is_sidechain'), _chain('is_backbone')),
+        (('sidechain', 'is_sidechain'), _chain('is_sidechain')),
 
         ###--- Atom.residue.<attribute> ---###
         (('protein', 'is_protein'), _chain('residue', 'is_protein')),
--- a/MDTraj/core/topology.py
+++ b/MDTraj/core/topology.py
@@ -1132,10 +1132,12 @@ class Atom(object):
         return ilen(bond for bond in self.residue.chain.topology.bonds
                     if self in bond)
 
+    @property
     def is_backbone(self):
         """Whether the atom is in the backbone of a protein residue"""
         return self.name in set(['C', 'CA', 'N', 'O']) and self.residue.is_protein
 
+    @property
     def is_sidechain(self):
         """Whether the atom is in the sidechain of a protein residue"""
         return self.name not in set(['C', 'CA', 'N', 'O']) and self.residue.is_protein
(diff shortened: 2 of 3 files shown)


kyleabeauchamp:
For most of my packages, I've been avoiding `develop` in favor of forcing installs.  

To me, the issue with `develop` is that it makes it harder to guarantee that your a testing the same environment that end-users will see.

For example, for many of the projects I now do testing via `conda-build devtools/conda-recipe`, which creates a clean test environment and runs the nosetests outside of the source tree.  I agree that there are pros and cons with either way.

kyleabeauchamp:
LGTM

synapticarbors:
What I ended up doing was pretty similar actually, although more manual:

```
conda create -n mdtraj-test python cython numpy setuptools scipy pandas ipython nose
python setup.py install
```

The only problem is that when there is a small typo that needs to be corrected, you have to basically re-run everything which is pretty slow and makes iterating on problems difficult. Further, because of the twidling that setup.py does with names and script locations, you can't easily import mdtraj from the source tree without an install (although maybe there is an easy way to do this). 

It might be nice to add something to the docs about setting up a dev environment.

rmcgibbo:
I usually use `setup.py develop`, and haven't had too many problems. Sometimes its a little flakey, and interacts poorly with versions installed in site-packages.

synapticarbors:
@rmcgibbo strange that `setup.py develop` wasn't working for me with a clean conda environment:

```
git clone https://github.com/mdtraj/mdtraj.git mdtraj2
cd mdtraj2/
conda create -n mdtraj-test2 python ipython numpy pandas scipy nose cython
source activate mdtraj-test2
python setup.py develop
```

with the output of the last command ending with:

```
...
Creating /Users/lev/anaconda/envs/mdtraj-test2/lib/python2.7/site-packages/mdtraj.egg-link (link to .)
Adding mdtraj 1.3.0.dev0 to easy-install.pth file
Installing mdinspect script to /Users/lev/anaconda/envs/mdtraj-test2/bin
Installing mdconvert script to /Users/lev/anaconda/envs/mdtraj-test2/bin

Installed /Users/lev/Documents/projects/github/mdtraj2
Processing dependencies for mdtraj==1.3.0.dev0
Finished processing dependencies for mdtraj==1.3.0.dev0
```

Then trying to import mdtraj:

```
(mdtraj-test2) $ python -c "import mdtraj"
  File "<string>", line 1, in <module>
ImportError: No module named mdtraj
```

The same exact procedure works fine with MSMBuilder on the same machine.

rmcgibbo:
Try `mv MDTraj mdtraj` from the root project directory.

synapticarbors:
@rmcgibbo ok, that fixed things. Thanks

rmcgibbo:
I'll look at trying to fix this in the repo.

PR #697, opened 2015-01-27
Typo in mdcrd.py

mpharrigan:


mpharrigan:
wait there are more

mpharrigan:
quick shoutout to pycharm for having a nice comment spellchecker

Issue #637, opened 2014-10-20
`is_protein` fails for nonstandard residue names


mpharrigan:
For example, `ASH` = protenated Asp and `CYP` = palmitoylated Cys were not recognized as "protein" residues and were not included in my featurization. `HIP` = protenated HIS worked, however. I see there is an alias for this defined in `pdb/data/pdbNames.xml`

Possible fixes:
 1. Add aliases for every conceivable residue name in `pdbNames.xml`
 2. Switch logic of `is_protein`. e.g. VMD looks for any residue which contains atoms named `C, N, CA, O` ref: http://www.ks.uiuc.edu/Research/vmd/vmd-1.3/ug/node133.html

synapticarbors:
I'll add that I just stumbled across this issue when dealing with histidine naming conventions from NAMD/CHARMM (i.e. `HSD`, `HSE`).

kyleabeauchamp:
I vote to postpone this because it is related to a longer-term discussion of how to best maintain compatibility between code that is duplicated in OpenMM and mdtraj (e.g. topology and PDB parsers)

mpharrigan:
Sure.

This issue should be kept open, though

cxhernandez:
https://github.com/mdtraj/mdtraj/pull/871 might help with this to some extent, since `residues.py` has been updated to include all residues found in the PDB.

mpharrigan:
fixed in #871

PR #700, opened 2015-01-28
Added additional aliases for HIS

synapticarbors:
Added two more CHARMM/NAMD specific aliases for histidine to partially address https://github.com/mdtraj/mdtraj/issues/637. I believe that HSD, HSE and HSP are the standard naming schemes there.

### DIFF
--- a/MDTraj/formats/pdb/data/pdbNames.xml
+++ b/MDTraj/formats/pdb/data/pdbNames.xml
@@ -116,7 +116,7 @@
   <Atom name="HB3" alt1="HB1" alt2="1HB" alt3="3HB"/>
   <Atom name="HE1" alt1="HNE"/>
  </Residue>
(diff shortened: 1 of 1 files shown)


kyleabeauchamp:
This sound reasonable to me and continues the tradition that we are slightly looser wrt the RCSB formal specification as compared to OpenMM.

synapticarbors:
I think the big thing is to not surprise users. The selection DSL is really powerful, but it's potentially really frustrating if you ask for `protein` atoms and all of your histidines are missing because you don't understand how mdtraj is deciding these things and you expect things to behave fairly similarly to VMD or another tool.

rmcgibbo:
Cool. Yeah, we could switch to the even looser VMD convention for defining is_protein, as @mpharrigan suggested.

PR #708, opened 2015-01-30
Move scripts directory

rmcgibbo:
- move `scripts/` to `mdtraj/scripts`. this is required for `setup.py develop` to function properly.
- a couple simplifications to `setup.py` since setuptools isn't optional.


mpharrigan:
+1

PR #709, opened 2015-02-05
Update MANIFEST.in

ctk3b:
I think this should be changed with the folder being renamed. Not sure if it actually has an effect but it fixes this warning during `setup.py`:

```
reading manifest template 'MANIFEST.in'
warning: no directories found matching 'MDTraj'
```


mpharrigan:
Good catch

Issue #721, opened 2015-02-17
Unable to load a gro file


HubLot:
Hi,

I'm using the 1.2.0 version of mdtraj and each time I want to load a gro file, I've got this error :

```
In [1]: import mdtraj as md

In [2]: md.load("prod.gro")
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-2-672a3e1ebc2a> in <module>()
----> 1 md.load("prod.gro")

/home/hub/.local/lib/python2.7/site-packages/mdtraj/core/trajectory.pyc in load(filename_or_filenames, discard_overlapping_frames, **kwargs)
    360 
    361 
--> 362     value = loader(filename, **kwargs)
    363     return value
    364 

/home/hub/.local/lib/python2.7/site-packages/mdtraj/formats/gro.pyc in load_gro(filename, stride, atom_indices, frame)
     94             coordinates, time, unitcell_vectors = f.read(stride=stride, atom_indices=atom_indices)
     95 
---> 96         coordinates = in_units_of(coordinates, f.distance_unit, Trajectory._distance_unit, inplace=True)
     97         unitcell_vectors = in_units_of(unitcell_vectors, f.distance_unit, Trajectory._distance_unit, inplace=True)
     98 

NameError: global name 'in_units_of' is not defined
```

synapticarbors:
Looks like a simple missing import. In `mdtraj/format/gro.py`:

```python
from mdtraj.utils import cast_indices, ensure_type
```

should be:

```python
from mdtraj.utils import in_units_of, cast_indices, ensure_type
```

I'd need to look a little deeper into why the test isn't picking this up, but otherwise it should be a simple fix.

synapticarbors:
It looks like there were a few additional issues. I've submitted a PR that addresses these

HubLot:
Great, thanks for the quick reaction.

synapticarbors:
@HubLot -- @rmcgibbo just merged my PR that should fix your issue. It's now in the master branch.

HubLot:
Indeed, it fixed it. Thanks.

PR #722, opened 2015-02-17
Fix gro loading for issue 721

synapticarbors:
This a fix that should address https://github.com/mdtraj/mdtraj/issues/721 and several other issues that weren't immediately obvious from the reported traceback.

### DIFF
--- a/mdtraj/formats/gro.py
+++ b/mdtraj/formats/gro.py
@@ -91,9 +91,9 @@ def load_gro(filename, stride=None, atom_indices=None, frame=None):
         topology = f.topology
         if frame is not None:
             f.seek(frame)
-            coordinates, unitcell_vectors, time = f.read(n_frames=1, atom_indices=atom_indices)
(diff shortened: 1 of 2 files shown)


kyleabeauchamp:
I think this looks good.  Can you also fix the docstring in the `read()` function to match the mixed up return order.  E.g.

https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/gro.py#L200

synapticarbors:
@kyleabeauchamp or is it better to change the order of the return to keep with general convention? It's a little hard to tell if this is consistent across most formats. For instance for the `xtc` reader, `time` comes before box info.

synapticarbors:
Actually, looking through the different readers `time` always is returned before box info, so I'll change the return order rather than the docstring.

kyleabeauchamp:
Either way sounds good to me.

kyleabeauchamp:
LGTM

PR #724, opened 2015-02-18
[ENH] compute RDF

ctk3b:
Seems to be working. The current bottleneck is in the histogramming and could probably use some C++ love. 

I added some reference files produced by gromacs' `g_rdf` and outlined the tests to compare but the binning is done slightly differently right now.


rmcgibbo:
I'll take a look at this later tonight.

mdtraj/geometry/rdf.py
```diff
@@ -0,0 +1,92 @@
...
+def compute_rdf(traj, pair_names=None, r_range=None, bin_width=0.005,
+                periodic=True, opt=True):
+    """Compute radial distribution functions for pairs in every frame.
+
+    Parameters
+    ----------
+    traj : Trajectory
+    pair_names : array-like, shape=(2), optional, default=None
+        Pair of atomtypes to consider.
+    r_range : array-like, shape=(2), optional, default=(0.0, 1.0)
```
kyleabeauchamp: below it says default of 0.99

mdtraj/geometry/rdf.py
```diff
@@ -0,0 +1,92 @@
...
+    bins = np.arange(r_range[0], r_range[1], bin_width)
+    g_r, edges = np.histogram([0], bins=bins)
+    g_r[0] = 0
+
+    if not pair_names:
+        # all-all
+        pairs = np.array(list(itertools.combinations(range(traj.n_atoms), 2)))
+    else:
+        type_a = traj.top.select("name =~ '{0}'".format(pair_names[0]))
+        type_b = traj.top.select("name =~ '{0}'".format(pair_names[1]))
```
mpharrigan: Probably bad form to use string formatting + selection language internally when we can use the API.

Using regex matching might be overkill too? Maybe something like
```
[a.index for a in traj.top.atoms if a.name.startswtih(pair_names[0])]
```

mdtraj/geometry/rdf.py
```diff
@@ -0,0 +1,92 @@
...
+        # TODO: Probably a cleaner way to generate these pairs.
+        prod = itertools.product(type_a, type_b)
+        pairs = set()
+        for a, b in prod:
+            if a != b and (b, a) not in pairs:
+                pairs.add((a, b))
+        pairs = np.array(list(pairs))
```
mpharrigan: Could you refactor this into a generator
```
def unique_pairs(type_a, type_b):
    ...
    for a, b in prod:
        (logic)
        yield a, b
```

kyleabeauchamp:
I'm happy.

mdtraj/geometry/rdf.py
ctk3b: That's probably worth doing just to have the function segregated. But it wouldn't provide any significant performance benefit since it's being dumped into an array for the distance compute, right?

mdtraj/geometry/rdf.py
ctk3b: Would it make sense to still do the regex matching? Something like

    [a.index for a in traj.top.atoms if (re.match(pair_names[0], a.name) is not None)]

I suppose there aren't very many use cases that this would cover and your suggestion would not...

mdtraj/geometry/rdf.py
mpharrigan: You're right. I thought maybe compute_distances could leverage an iterator to save memory but it doesn't :)

mdtraj/geometry/rdf.py
mpharrigan: I don't have a strong opinion, but if you opt for the regex, I'd document it in the docstring.

kyleabeauchamp:
One more thing: right now the `opt` argument is really just being passed along to the distance calculation.  Maybe add a note in the docstring that `opt` refers to the optimized code for the distance calculation?

mdtraj/geometry/rdf.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def compute_rdf(traj, pair_names=None, r_range=None, bin_width=0.005,
+                periodic=True, opt=True):
+    """Compute radial distribution functions for pairs in every frame.
+
+    Parameters
+    ----------
+    traj : Trajectory
+    pair_names : array-like, shape=(2), optional, default=None
```
rmcgibbo: `(2)`->`(2,)`

mdtraj/geometry/rdf.py
```diff
@@ -0,0 +1,96 @@
...
+        Radial distribution function values at r.
+    """
+    if not r_range:
+        r_range = np.array([0.0, 1.0])
+    r_range = ensure_type(r_range, dtype=np.float64, ndim=1, name='r_range',
+                          shape=(2,), warn_on_cast=False)
+    bins = np.arange(r_range[0], r_range[1], bin_width)
+    g_r, edges = np.histogram([0], bins=bins)
+    g_r[0] = 0
+
```
rmcgibbo: check that `pair_names` is of length 2?

mdtraj/geometry/rdf.py
rmcgibbo: you can do it in one line, if you really want.

`np.array(list(set(((a,b) if a>b else (b,a) for (a,b) in product(type_a, type_b) if a != b))))`

mdtraj/geometry/rdf.py
```diff
@@ -0,0 +1,96 @@
...
+        Minimum and maximum radii.
+    bin_width : int, optional, default=0.005
+        Width of the bins in nanometers.
+    opt : bool, default=True
+        Use an optimized native library to calculate the RDF.
+
+    Returns
+    -------
+    r : np.ndarray, shape=(np.diff(r_range) / bin_width - 1), dtype=float
+        Radii values corresponding to bins.
```
rmcgibbo: From the code, the returned `r`s are the center of the bins, but either the left and right bin edges also be reasonable things the user might expect. Maybe the docstring should say that they're the center points.

rmcgibbo:
@ctk3b can you add an entry to the top section of `docs/whatsnew.rst` with your name and the function name to the appropriate spot in `docs/analysis.rst`? The code looks good to merge. Also, you can change the tag on the PR from `[WIP]` to `[ENH]`.

swails:
What is ENH?  Google and Urban Dictionary have failed me here...

rmcgibbo:
Enhancement.

rmcgibbo:
... at least that's how i've been using it all this time...

swails:
ack, I was looking for an acronym (like WIP) :)

rmcgibbo:
I have a question / concern about how the histogram is normalized.

Assume that the `\hat{g}(r)` is constructed from evenly spaced bins. Consider an arbitrary bin that runs over the interval `r_inner < r <= r_outer`. In this code, the estimator for `\hat{g}(r)` over this interval is

`(4/3 \pi n (r_outer^3 - r_inner^3))^{-1} * (\sum_{r \in R : r_inner < r <= r_outer} 1)`

where `R={r_1, r_2, ...}` is the set of all measured interatomic distances and `n` is the particle number density.

That is, we count up all the data points in `R` that fall within the bin and then the sum is normalized. If you pull it inside the sum, it's as if each data point within the bin is being normalized by an identical factor. But if we now look between multiple bins, it's obvious that the normalization factor is not constant. It's falling off like 1/r^2. Instead of using this step-function normalization, I suggest that a more accurate normalization is to compute our estimator as.

`n^{-1} * (\sum_{r \in R : r_inner < r <= r_outer} (4\pi r^2)^{-1})`

Here, instead of summing over the data points inside the histogram window weighted equally, they're each applied the appropriate normalization for their particular radius. (Note that here you could also easily substitute in any weighted kernel density estimator in place of the histogram.)

The difference between these two estimators is when the bin widths are large, since the first is really a sort of crude integral of the second. Assume for example that there are only two bins, one for `r < 10A` and then other for `r >= 10A`, and we're trying to build `\hat{g}(r)`. We consider a series of simulation systems of fixed density, but increasing volume and particle number. For the first estimator, `r_outer` is going to diverge to infinity as the boxes get bigger and bigger, because it needs to be at minimum the largest intermolecular distance in `R`. This means that estimated `\hat{g}(r)` will go to zero for `r>=10A`. For the second estimator, this won't happen, since each point in that window is normalized by its own `r`, as opposed to `r_outer`.

On the other hand, I looked at the source for for Ambertools cpptraj and gromacs g_rdf, and they both use the first estimator.

rmcgibbo:
Also, @ctk3b, thanks a ton for contributing this! It's awesome to have you working on mdtraj and I hope the comment above doesn't come off as rude or anything. It just got me thinking because I really like kernel density estimators, and it wasn't clear to me how to apply one to g(r) when the normalizer depends on `r_outer^3 - r_inner^3` which wouldn't make sense for a KDE.

mdtraj/geometry/rdf.py
```diff
@@ -0,0 +1,102 @@
+##############################################################################
+# MDTraj: A Python Library for Loading, Saving, and Manipulating
+#         Molecular Dynamics Trajectories.
+# Copyright 2012-2013 Stanford University and the Authors
```
swails: Probably worth fixing the date here.

kyleabeauchamp:
I'm no expert, but I wonder if people historically picked this particular "kernel" to eliminate the bias on some bulk observable like density?  I imagine Allan and Tildsley has discussion on this, I may poke around.

kyleabeauchamp:
p86 in Frenkel and smit has an RDF algorithm in case we need some references to cite.

ctk3b:
No worries at all! What you proposed sounds sensible at first glance but I'm on my phone right now and won't have time to take a closer look until probably Sunday.

Might also make sense to factor out some parts to give users different options for normalization/estimators?

rmcgibbo:
@kyleabeauchamp the fortran code on p86 is equivalent to the one implemented here.

kyleabeauchamp:
Working through it now.  Should probably add a note or reference to this textbook though, good to have a paper trail for these things.

rmcgibbo:
I have also been staring at eq 4.6.19 from Tuckerman, "Statistical Mechanics and Molecular Simulations" https://books.google.com/books?id=Lo3Jqc0pgrcC&pg=PA157&lpg=PA157&dq=radial+distribution+function+tuckerman&source=bl&ots=vBI3EWgqxC&sig=k6ImCr2e-ApWkx-RSE3UB0QZ9Hg&hl=en&sa=X&ei=8XDmVKTxJc3YoATE3YHwDg&ved=0CEoQ6AEwCg#v=onepage&q=radial%20distribution%20function%20tuckerman&f=false

leeping:
I think either way should be fine. To determine which method is better might require setting some criteria for how we expect it to behave. I'll suggest one below but let me know if you think it's wrong.

If I obtain a coarse RDF with a bin from r to r+2*dr, I might intuitively expect the height of that bin to be the average of two bins from a fine RDF computed from the same data set (r to r+dr, and r+dr to r+2dr.) It could be interesting to see which estimator performs better for this "criterion" in a nuerical test.

rmcgibbo:
That criteria might be something, but it's not sufficient. I could trivially return an estimate `\hat{g}(r) = 0` and pass with flying colors.

IMHO, the appropriate criterion is the expected mean squared error `E[ ||\hat{g}(r) - g(r)||_2^2 ]`. Other norms would be suitable too. This is also called the risk of the estimator. Obviously computing it requires knowing the true `g(r)` though. But the risk breaks down into [bias and variance terms](http://en.wikipedia.org/wiki/Bias%E2%80%93variance_tradeoff), and my contention is that these two estimators have basically the same variance, but that one has lower bias.

rmcgibbo:
Do you know any systems where the true g(r) is analytically known?

swails:
Pure Lennard-Jones liquid.

swails:
Or fluid. Gas. You know what I mean. It's actually in the same chapter of the tuckerman book you linked to earlier.

rmcgibbo:
I didn't see a full analytic expression for g(r) in there. I saw some plots, but it wasn't clear how they were made. From a quick googling, the implication seems to be that the 'analytic' expressions are just parametric fits to simulation data.
- http://www.sciencedirect.com/science/article/pii/S0301010404005452
- http://www.uic.edu/labs/trl/MatteoliMansooriRDFexpression.pdf

swails:
Hmm...  Seems I skimmed too quickly -- those articles do seem to suggest that the RDF has no closed-form solution.  I find that surprising given that a pure LJ fluid is a very simple pairwise potential that *should* uniquely define a pair correlation function.

Perhaps this is something like the DFT problem -- a unique functional *does* exist, we just don't know what it is.

Another thing that's probably reasonable is to either take one of those parametric solutions, or simply use a fine-bin-spacing on a heavily-sampled simulation as the "correct" ensemble (since both estimators should be equivalent in the limit of infinitely small bin sizes, right?), and see which estimator performs better as the bin size increases.

rmcgibbo:
Well it obviously does exist, at least for a fixed Hamiltonian, canonical ensemble. You can write it down with eq. 4.6.19. Less complex than Kohn-Sham theore.. More like saying that the wave function exists.

Although Usually when people say "The L-J RDF", they probably mean some kind of limit as the number of particles and volume goes to infinity, which eq. 4.6.19 doesn't prove exists.

> Another thing that's probably reasonable is to either take one of those parametric solutions, or simply use a fine-bin-spacing on a heavily-sampled simulation as the "correct" ensemble (since both estimators should be equivalent in the limit of infinitely small bin sizes, right?), and see which estimator performs better as the bin size increases.

Yes, true. It would just be a little more convenient if there was an unambiguous and easily-computable right answer to compare against. But in practice it doesn't matter much.

rmcgibbo:
OTOH, this is a huge rabbit hole and I think mostly as waste of time. My strong intuition is that the estimator I proposed is slightly less biased, but that in practice it makes very little difference.

If it can be proved that the estimator is less biased, it's perhaps suitable for a short note on arXiv. Otherwise, honestly, I think it doesn't matter.

swails:
I agree.  If your bin sizes are large enough that the estimator bias is significant, I suspect that your bin sizes are *way* too large (e.g., nobody uses only 2 bins :)).

OTOH, a bin-free estimator built from KDEs would indeed be nice. But I've never had problems with my RDFs not being smooth or being a challenge to integrate numerically to a high degree of precision.

kyleabeauchamp:
@rmcgibbo  maybe we should summarize the remaining to-do items for@ctk3b for the merge?

ctk3b:
This sounds like something interesting to look into but as @rmcgibbo mentioned, I suspect it's unlikely to be of great practical utility (for this immediate purpose). If it can be shown to reduce the bias, it would certainly make sense to adjust the normalization but for now I would suggest sticking with the method employed by the other frequently used tools, if nothing else, to produce the result that most users would expect.

rmcgibbo:
Sounds good. I'll merge this, and then add a comment to this file referring to this discussion.

PR #725, opened 2015-02-18
Refactor and fix .gitignore(s)

mpharrigan:


mpharrigan:
Things broke with the folder rename `MDTraj -> mdtraj`

If we switch to this scheme (`.gitignores` per directory), this will not happen again

rmcgibbo:
Nice.

Issue #728, opened 2015-02-19
[FR] Atom select check for valid keywords


synapticarbors:
I think it would improve user experience and reduce silent errors if the selection DSL did some checking for valid keywords:

```python 
In [43]: traj.top.select('protein').shape
Out[43]: (7705,)

In [44]: traj.top.select('protein and heavy').shape
Out[44]: (7705,)

In [45]: traj.top.select('protein and dog').shape
Out[45]: (7705,)
```

swails:
> In [45]: traj.top.select('protein and dog').shape
> Out[45]: (7705,)

I suspect all the atoms in your system *would* actually be present in a dog, though...

ctk3b:
> I suspect all the atoms in your system would actually be present in a dog, though...

Definitely a feature. Have you tried "protein and rock"?

swails:
> Definitely a feature. Have you tried "protein and rock"?

Psh.  Stop thinking small.

```Python
In [46]: traj.top.select('protein and mars').shape
Out[46]: (7705,)
```

Sounds like a Science paper to me.  "**Proteins found on Mars!**"

synapticarbors:
@swails Stop thinking small.

```python
In [46]: traj.top.select('protein and god').shape
Out[46]: (7705,)
```

swails:
Good luck publishing *that* in *Science*...

mpharrigan:
Nice find. It is parsing random strings as literals

rmcgibbo:
Fixed.

PR #729, opened 2015-02-19
[ENH] Don't allow bare literals in selection expressions

mpharrigan:
- [x] write test
- [x] fix

### DIFF
--- a/mdtraj/core/selection.py
+++ b/mdtraj/core/selection.py
@@ -353,6 +353,10 @@ class parse_selection(object):
         # name specified below (i.e. 'atom')
         astnode = self.transformer.visit(deepcopy(parse_result[0].ast()))
 
+        # Special check for a single literal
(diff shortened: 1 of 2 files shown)


mpharrigan:
I don't think there's a way to do what we want with just pyparsing grammar. I added in checks in the ParseAction objects.

I can't figure out a way to fail on single keywords
```
select('dog')
```
Will still work, whereas

```
select('protein and dog')
```
will fail because the `and` tree can tell that literals are no good

mpharrigan:
cc #728

mpharrigan:
I added a special case.

For the record: @rmcgibbo and I discussed a more [elegant, convoluted] scheme wherein we would parse strings and numbers to subclasses whose `__bool__` method raises an exception

rmcgibbo:
There also might be another way to solve this at the level of the grammar, by changing where literals are allowed to occur. But I don't think either @mpharrigan or myself understand the structure of the grammar well enough to do this reliably. This code is... nontrivial.

PR #734, opened 2015-02-23
[ENH] Faster histogram

ctk3b:
The current RDF code spends ~80% of its time in the histogram on my laptop so I figured I'd try my hand at writing a faster one (is this surprising/suspicious or is your distance code just that awesome?).

From some quick timings, this histogram is a ~10x faster for the RDF use case:
```python
In [8]: a = np.random.rand(10000000)

In [9]:  %timeit np.histogram(a, bins=100)
1 loops, best of 3: 928 ms per loop

In [10]: %timeit _numericutils._histogram(a, bins=100)
10 loops, best of 3: 83.6 ms per loop

In [11]: a = np.random.rand(100000000)

In [12]:  %timeit np.histogram(a, bins=100)
1 loops, best of 3: 8.91 s per loop

In [13]: %timeit _numericutils._histogram(a, bins=100)
1 loops, best of 3: 975 ms per loop
```

That being said, this is my first time dabbling with cython and I don't write a lot of C/C++ so I've got a few questions:
* First and foremost, do you want something like this in your code base?
* If yes, do my file structure and function names make sense to you? Any conventions that I'm doing a terrible injustice to?
* I've tried to reproduce the behavior of the numpy histogram with the exception of renaming the `range` arg to `bin_range`. The PR also includes all of the unit tests that numpy uses but modified to also check that the results are equal to those produced by numpy. Any objections to this approach and/or that particular change?

Other notes/TODO's:
* In `test_f32_rounding()` There is a slight difference (out to ~4-5 decimals) in the returned edges as compared to numpy. Still need to identify where this is coming from and fix it. The actual values in the histogram appear to be consistent, however, for the tests I've run so far.
* I have yet to implement the numpy functionality provided by the args `normed`, `weights` and `density` because there wasn't an immediate need.
* The current code could easily just be regular cython without further wrapping but I figured it may be worthwhile to keep the option open for some kind of SIMD acceleration (which is a bit [tricky for histograms](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.80.9041&rep=rep1&type=pdf)). Might also make it easier to add some OpenMP threading to accumulate sub-histograms but that can be done in cython too (some potential [code](https://github.com/kif/pyFAI/blob/master/src/histogram_omp.pyx) to borrow from).


kyleabeauchamp:
I guess my first question is whether this has already been implemented by, say, scipy, sklearn, or pandas, which might be considered for inclusion as an optional import.

ctk3b:
[Scipy uses numpy internally](https://github.com/scipy/scipy/blob/v0.14.0/scipy/stats/stats.py#L1666)
[Pandas has some cython](https://github.com/pydata/pandas/blob/9f439f0e53409fd48e5486a3acf84b844f200270/pandas/hashtable.pyx#L983)

I can't immediately find anything in sklearn. I'll benchmark the pandas one.

kyleabeauchamp:
Yeah, I thought the pandas folks have made some performance improvements over numpy / scipy for several "mission critical" tasks.  

Also, we already have some optional pandas dependencies scattered around the repo, so it wouldn't be a big burden to use that if it's fast.   

Finally, if your improved histogramming is a major performance improvement, I wonder if you're better off filing a PR directly to pandas.  Certainly, that would impact more users.

rmcgibbo:
> The current RDF code spends ~80% of its time in the histogram on my laptop 

How does this compare to the amount of time it takes, for example, to simply load the trajectory files from disk? Is this a big overall contributor to the time it takes to accomplish some task, or are we talking about accelerating 80% of something that already only takes milliseconds?

rmcgibbo:
The pandas histogram uses a hash table I think. Different implementation from numpy. At one point, I believe there was a discussion on numpy-dev or scipy-dev about this.

ctk3b:
@kyleabeauchamp  Well I haven't really done anything earth shattering here - it's a pretty bare bones implementation of only the part I wanted for the RDF. The pandas implementation doesn't seem to be any faster than numpy so maybe it'd be worth looking into improving their cython code.

@rmcgibbo for the tip3p trajectory, I get ~14s total, 11 spent in np.histogram, 2.5 in distances. With the PR histogram it's 4.1 total, 2.7 in distances and 1.0 in the histogram (note these are based on just a single run but I think the loading isn't limiting in either case).

ctk3b:
Using: 
```python
import cProfile, pstats, StringIO
import mdtraj as md
from mdtraj.testing import get_fn
pr = cProfile.Profile()
pr.enable()

traj = md.load(get_fn('tip3p_300K_1ATM.xtc'), top=get_fn('tip3p_300K_1ATM.pdb'))
r, g_r = md.compute_rdf(traj)

pr.disable()
s = StringIO.StringIO()
sortby = 'cumulative'
ps = pstats.Stats(pr, stream=s).sort_stats(sortby)
ps.print_stats()
print s.getvalue()
```

ctk3b:
Semi-related, how do you guys test across python versions? I'm using conda environments but from my understanding those don't play nicely with tox.

kyleabeauchamp:
Personally, I let the travis grid handle python versions for the most part.  When I need to test things locally, I end up using either conda environments or a separate (clean) miniconda install.

ctk3b:
Also if you guys aren't comfortable adding a re-invented wheel to your repo, that's more than understandable - it really does seem like it would belong in a package like numpy or pandas. This mostly started as a learning experience to fully understand how all the cython extensions work in mdtraj. 

I'll take a closer look at how accelerating the pandas code would work.

kyleabeauchamp:
I would look up the discussion on the listserv first, in case there are
some hard problems that we're not anticipating.

rmcgibbo:
Overall, I'm a little -1 for the following reasons:

1. In general, I think we should avoid having code in mdtraj that isn't "core" to our mission and competencies. We want to lean on numpy and friends as much as possible. They spend a lot of effort, which we don't have the manpower (or will) to reproduce, dealing with cross-platform compatibility, documentation, etc.

2. I think we should have a particularly high bar for including new C or C++ code in mdtraj. Two reasons:
  - A smaller percentage of the mdtraj developers or people who might in the future become mdtraj developers count hacking in C among their skills. The pool of people who can identify and fix bugs, do maintenance, and extend the code are simply smaller. This is worrisome if we want the code to continue to be maintained through a cycle of core-developer-graduates-and-a-new-team-steps-up.  Honestly, this is something I'm worried about already.
 - Cross platform support in C and C++ is just harder. Especially with C and Windows. There are pretty substantial differences between what's supported in C with MSVC and with gcc/clang. And the pool of developers who can work with windows and C/C++ is pretty small for us.

Obviously these considerations have to be balanced against performance. There is a balancing act. I'm not sure in general how to find the right balance, and we don't really have a project protocol laid out of this kind of thing.

kyleabeauchamp:
I agree that we need a high bar for additional C code.  

My vote is to first see whether this belongs in Numpy / pandas--in those places, it's easier to "hand off" code to the community.  If we merge it here, it means that we personally are going to be maintaining it.

ctk3b:
That all sounds very reasonable. I'll do some closer reading of the numpy and pandas lists to see what exactly people are working on. Specifically for the RDF, we don't really need a feature rich implementation but as mentioned, maintaining it across platforms will be the tricky part.

I was really surprised though that something like this doesn't exist already. It sounded so simple and wasn't hard to write but I am only looking at a pretty specific use case. From what I've been reading on the numpy PR's that worked on histogram performance, the main concern seems to be generality for all inputs sizes.

swails:
I'll interject here with a slightly different perspective, particularly with regards to:

> Obviously these considerations have to be balanced against performance. There is a balancing act. I'm not sure in general how to find the right balance, and we don't really have a project protocol laid out of this kind of thing.

I think a good philosophy is to have a software project like this be, first and foremost, useful to the people that develop it.  With that in mind, @ctk3b, how critical to your work is the RDF speedup?  In particular:

1. Does it make the difference between using MDTraj vs. a different package for your application?
2. Does your new code eliminate performance-based limits on the neat stuff you're trying to do with it, allowing you to do more, neater stuff? (related to 1, really)
3. Does this need to be part of MDTraj proper for it to be realistically useful to you? (i.e., can you use it in your own work as a kind of 'plug-in' type of thing?)

The answers to these questions may help define where the importance of this feature lies in the balancing act @rmcgibbo was referring to.

Relating to 3 -- perhaps another route you can go is to provide an option to register a histogramming function to use with the RDF, defaulting to numpy's implementation.  That way, you can play around with which one is best, and plug in your own if none of the available ones suffice, without having to dump anything like this into MDTraj proper.  Just a thought.

ctk3b:
@swails the speedup is not *critical* to my current work - it's more of a convenience at this point. It may convince some others in my group to use MDTraj (their bottleneck is currently the generation of a large amount of RDF data and they would like to perform most analysis in the python eco-system) but there are some other tools that would have to be ported over as well and I would need to do some proper benchmarking with them first to fully justify this. For now, it's easy enough to maintain a minor tweak on my copy of MDTraj to use this code if I need to.

I'll look into possible ways to speed up the pandas cython but I suspect this will be somewhat challenging to get into the numpy source.

kyleabeauchamp:
OK.  I think if this performance gain is important for your application, I'm willing to be flexible here.  I think we may just need to brainstorm about how we can optimize the performance versus maintainability ratio.

mpharrigan:
Why not just do it in cython

kyleabeauchamp:
Yeah, might be worth doing a separate PR to see if cython can give similar performance with more readability

ctk3b:
@mpharrigan that would be the other option. The current code does nothing fancy that prevents it from just being plain cython. But that doesn't really eliminate the cross-platform issue, right? It would be quite a bit simpler regardless.

kyleabeauchamp:
I'm not sure what the best approach is, will try to think about it as I find time

leeping:
Hi there,

Regarding performance, I wonder if anyone has gotten a rough timing from VMD's RDF calculation.  I've used it in all of my work, and I've done quite a few RDF calculations.  It could be a nice data point to have.

Thanks,

- Lee-Ping

gabrielelanaro:
I had same problem when I implemented rdf for chemlab. I resorted by using np.bincount, did you try that approach?

Ref: https://github.com/chemlab/chemlab/blob/master/chemlab/molsim/analysis.py

ctk3b:
@gabrielelanaro Hmm looks to be about 4x faster. Wouldn't the precision depend on the size of the bin width though? ` int_distances = np.rint(distances/binsize).astype(int)`. 

Something similar could probably be achieved by multiplying by a factor of 10 and dividing the values again after.

gabrielelanaro:
@ctk3b I didn't think about that issue, what do you mean by affecting the precision? Anyway, the multiplication approach would work too (and it shouldn't decrease performance)

ctk3b:
Sorry I didn't fully think that through. It shouldn't matter since your binning those values into that same size bin.

mdtraj/geometry/src/numericutils.c
```diff
@@ -0,0 +1,64 @@
...
+    float binwidth = (max_bin - min_bin) / n_bins;
+    for (i = 0; i < n_data; ++i)
+    {
+        value = data[i];
+        if (value < min_bin || value > max_bin)
+        {
+            continue;
+        }
+
+        bin = floor((value - min_bin) / binwidth);
```
jaimefrio: Take a look at [Agner Fog's optimization manual](http://www.agner.org/optimize/optimizing_cpp.pdf) and see what he has to say about "float to integer conversions" on section 7.11, page 42. Because you know that `value - min_bin` is a positive number, you are likely going to see a noticeable speed-up if you skip the call to `floor` and simply cast the result of the division to `int`. Last week at work we shaved 10% off a similar function call by doing just that. Basically, because `floor` returns a `double`, after the call to floor you still need to cast the result to int, so you are basically doing the exact same work twice! The only thing that `floor` does that integer casting doesn't is handle overflows properly. But since `bin` is after all an index you should probably make it `ssize_t` anyway, and if that overflows you are screwed anyway, because your histogram would not fit in memory.

rmcgibbo: I hadn't seen that optimization manual. That's a really nice resource!

jaimefrio: It really is pure gold for the premature optimizer that lives in all of us. ;-)

swails:
It sounds like this might effectively be done for us.  See the numpy discussion I just linked.  We'll see if/when it becomes a numpy PR, but even the pure-python/numpy 5x speedup will put it within a factor of 2 of the speedup seen here (and if they are convinced to go the Cython route, it'll be even faster :)).

swails:
Well that was fast!  @ctk3b -- do you still have the benchmarks lying around for your RDF calculations?  If so and you can find the time, try out numpy/numpy#6100 and see if you get the 10x speedup you got here with the new numpy code instead.

It would be great if we could just point users to "newer numpy" in order to get a faster RDF.

@gabrielelanaro -- this may apply to you with chemlab as well.  It looks like `numpy.histogram` will become fast enough (and I think they're strongly considering another 3x speedup by going to Cython) that this will become a moot point.

ctk3b:
Huzzah! The full RDF looks to be about ~3-4 times faster for this particular system (water box) which is consistent with @gabrielelanaro's solution in chemlab which also uses bincount.

However, it requires a small tweak to either the RDF code or alternatively the PR which may or may not be acceptable:

This [line](https://github.com/numpy/numpy/pull/6100/files#diff-9d44603693b9544b900e135a203b3ae2R188) in the PR checks to see if "the bins are equal width" by checking if `bins` is not iterable. Right now I set up the RDF code to just always provide an array of equal width bins to the histogram function. They're still equal width but the variable is an array and hence iterable. The easiest tweak is probably just to change how the RDF code works, but on the other hand, the numpy code is not technically checking for a case where the bins are equal width.

Timeit results:
```python
# 1.10.0.dev0+daf0497
In [2]: %timeit run() 
1 loops, best of 3: 10.1 s per loop

# #6100
In [3]: %timeit run()
1 loops, best of 3: 2.71 s per loop
```

And the performance breakdown with relevant parts shown:
New and improved:

```
85663 function calls (83563 primitive calls) in 2.769 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.046    0.046    2.769    2.769 test_rdf.py:4(run)
        1    0.000    0.000    2.596    2.596 //anaconda/envs/hist2/lib/python3.4/site-packages/mdtraj/geometry/rdf.py:33(compute_rdf)
        1    0.001    0.001    1.631    1.631 //anaconda/envs/hist2/lib/python3.4/site-packages/mdtraj/geometry/distance.py:42(compute_distances)
        1    1.628    1.628    1.628    1.628 {_geometry._dist_mic}
        1    0.731    0.731    0.960    0.960 //anaconda/envs/hist2/lib/python3.4/site-packages/numpy/lib/function_base.py:78(histogram)
     1831    0.097    0.000    0.097    0.000 {built-in method bincount}
        1    0.000    0.000    0.071    0.071 //anaconda/envs/hist2/lib/python3.4/site-packages/mdtraj/core/trajectory.py:294(load)
```

Old and totes unimproved:

```
130412 function calls (128312 primitive calls) in 10.081 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.048    0.048   10.081   10.081 test_rdf.py:4(run)
        1    0.000    0.000    9.907    9.907 //anaconda/envs/hist/lib/python3.4/site-packages/mdtraj/geometry/rdf.py:33(compute_rdf)
        1    0.056    0.056    8.263    8.263 //anaconda/envs/hist/lib/python3.4/site-packages/numpy/lib/function_base.py:78(histogram)
     1831    0.008    0.000    7.897    0.004 //anaconda/envs/hist/lib/python3.4/site-packages/numpy/core/fromnumeric.py:712(sort)
     1833    7.790    0.004    7.790    0.004 {method 'sort' of 'numpy.ndarray' objects}
        1    0.002    0.002    1.637    1.637 //anaconda/envs/hist/lib/python3.4/site-packages/mdtraj/geometry/distance.py:42(compute_distances)
        1    1.633    1.633    1.633    1.633 {_geometry._dist_mic}
     3662    0.207    0.000    0.207    0.000 {method 'searchsorted' of 'numpy.ndarray' objects}
```




Testing script:
```python
import mdtraj as md
from mdtraj.testing import get_fn

def run():
    traj = md.load(get_fn('tip3p_300K_1ATM.xtc'), top=get_fn('tip3p_300K_1ATM.pdb'))
    pairs = traj.top.select_pairs('all', 'all')
    r, g_r = md.compute_rdf(traj, pairs)

if __name__ == '__main__':
    import cProfile, pstats, io
    pr = cProfile.Profile()
    pr.enable()

    run()

    pr.disable()
    s = io.StringIO()
    sortby = 'cumulative'
    ps = pstats.Stats(pr, stream=s).sort_stats(sortby)
    ps.print_stats()
    print(s.getvalue())
```

swails:
Great!  My personal vote is for tweaking the RDF code in MDTraj to use an integer number of bins rather than providing the bin edges or midpoints.  Checking width equality will be tricky with floating-point bin.  I think giving the bin number is safer if it's easy to change here.

ctk3b:
Yeah, that's definitely the easiest and safest fix. I asked on the numpy PR just in case they do think it's a correctly/efficiently solvable corner case but I'll submit a PR to update the RDF code after seeing what they say.

rmcgibbo:
This is closed by #952, right?

ctk3b:
Yeah, it should be. If they get around to implementing the full histogram in cython, we'll get an even bigger boost but that should be discussed on the numpy repo.

Issue #620, opened 2014-10-05
Move WATER_TYPES to residue_names.py


kyleabeauchamp:
So `is_water` currently has hardcoded residue names (https://github.com/SimTk/mdtraj/blob/master/MDTraj/core/topology.py#L1076), which I suppose we should consolidate in `residue_names.py.

It might be useful to document how to get / extend the list of solvent types.  For now, I think it might be sufficient to include in the docstring how to programmatically list what, say, the PROTEIN_TYPES actually are.  

Eventually, we might want the ability to extend these types as well.

kyleabeauchamp:
Done

PR #740, opened 2015-02-25
Move water residue names to residue_names.py

kyleabeauchamp:
See #620

### DIFF
--- a/mdtraj/core/residue_names.py
+++ b/mdtraj/core/residue_names.py
@@ -50,5 +50,5 @@ _PROTEIN_RESIDUES = frozenset([
 
 
 _WATER_RESIDUES = frozenset([
-'H2O', 'HHO', 'OHH', 'HOH', 'OH2', 'SOL',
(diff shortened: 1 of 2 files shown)


mdtraj/core/residue_names.py
```diff
@@ -49,3 +49,6 @@
     'PHE', 'PRO', 'SER', 'THR', 'TRP', 'TYR', 'UNK', 'VAL'])
 
 
+_WATER_RESIDUES = frozenset([
+'H2O', 'HHO', 'OHH', 'HOH', 'OH2', 'SOL',
+                             'WAT', 'TIP', 'TIP2', 'TIP3', 'TIP4'])
```
rmcgibbo: weird tabs or something?

kyleabeauchamp:
fixed

mpharrigan:
This is strongly related to #637 

Specifically, If we normalize our inputs properly, there should only be *one* water residue name

cxhernandez:
Should 'HOH' still be in `_SOLVENT_TYPES ` then?

PR #744, opened 2015-02-27
[ENH] Improved efficiency of generating RDF pairs

ctk3b:
Generating `pairs` scales pretty terribly as the number of atoms increase. This change removes the intermediate list/set creation and uses `fromiter` which is both faster and more memory friendly.

### DIFF
--- a/mdtraj/geometry/rdf.py
+++ b/mdtraj/geometry/rdf.py
@@ -22,7 +22,7 @@
 
 from __future__ import print_function, division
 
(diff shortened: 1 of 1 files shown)


rmcgibbo:
+1. will merge after the tests run.

ctk3b:
So I don't forget and if anyone wants to test the memory/timings: https://gist.github.com/ctk3b/0673dd3bd482cd6271b3

PR #746, opened 2015-03-02
Overwrite not being passed to HDF5 and .lammpstrj and some small stuff (doc strings, pep8)

ctk3b:
Quick question: there are several docstrings with single quotes lying around with single quotes, e.g.:
https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/topology.py#L947

Is that intentional or just go ahead and fix it if I find em?

### DIFF
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -641,7 +641,7 @@ class Topology(object):
 
     @property
     def n_residues(self):
-        "Get the number of residues in the Topology"
+        """Get the number of residues in the Topology. """
         return len(self._residues)
 
     def atom(self, index):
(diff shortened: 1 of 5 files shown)


mdtraj/core/topology.py
```diff
@@ -641,7 +641,7 @@ def residues(self):
 
     @property
     def n_residues(self):
-        """Get the number of residues in the Topology"""
```
mpharrigan: Why are you *removing* the triple quotes in this case?

ctk3b: Woops sorry. Fixed.

mpharrigan: Haha cool

rmcgibbo:
Thanks.

PR #747, opened 2015-03-02
[ENH] Reader for .xyz files

ctk3b:
Super low priority but one step closer to "Read and write from every MD format imaginable".


mdtraj/core/trajectory.py
```diff
@@ -1236,6 +1238,19 @@ def save_lammpstrj(self, filename, force_overwrite=True):
                     cell_angles=self.unitcell_angles,
                     cell_lengths=self.unitcell_lengths)
 
+    def save_xyz(self, filename, force_overwrite=True):
+        """Save trajectory to LAMMPS custom dump format
```
rmcgibbo: LAMMPS -> xyz?

mdtraj/core/trajectory.py
```diff
@@ -1236,6 +1238,19 @@ def save_lammpstrj(self, filename, force_overwrite=True):
...
+
+        Parameters
+        ----------
+        filename : str
+            filesystem path in which to save the trajectory
+        force_overwrite : bool, default=True
+            Overwrite anything that exists at filename, if its already there
+        """
+        with XYZTrajectoryFile(filename, 'w', force_overwrite=True) as f:
```
rmcgibbo: pass through the `force_overwrite` option, instead of fixing to true?

mdtraj/formats/xyzfile.py
```diff
@@ -0,0 +1,350 @@
...
+            types = ['X' for _ in xrange(xyz.shape[1])]
+
+        in_units_of(xyz, 'nanometers', self.distance_unit, inplace=True)
+
+        for i in range(xyz.shape[0]):
+            self._fh.write('{0}\n'.format(xyz.shape[1]))
+            self._fh.write("Created with MDTraj {0}, {1}".format(version, str(date.today())))
+
+            for j, coord in enumerate(xyz[i]):
+                self._fh.write('{0} {1:8.3f} {2:8.3f} {3:8.3f}\n'.format(
```
rmcgibbo: is there are particular standard documented somewhere that this format adheres to that we can link to that specifies things like the level of precision?

ctk3b: From my understanding there is no universal standard but this is the one that VMD uses. I've updated the doc string to reference this fact.

rmcgibbo:
I know they're a little bit repetitive, but can you also add the XYZTrajectoryFile to the appropriate tests in https://github.com/mdtraj/mdtraj/blob/master/mdtraj/tests/test_trajectory.py, like `test_seek_read_mode` and maybe the list of files [here](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/tests/test_trajectory.py#L168-L169)

rmcgibbo:
This looks awesome!

PR #748, opened 2015-03-02
Add compilation fixes for extension modules

pfrstg:
* dcdplugin.h and xdrfile_xtc.h: include missing function declarations
* dridkernels.h: include #ifdef'd controlled extern "C" so a C++ compilier can
  read the header and successfully link.


pfrstg:
I'm having trouble running geometry/tests/test_dihedral.py, but it appears to be something about how pymol is invoked so I think it's unrelated to this change.

rmcgibbo:
LGTM.

PR #749, opened 2015-03-03
Compile on vsp-compute needs g++44 too

mpharrigan:
Not just CC=gcc44 but also CXX=g++44

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -22,9 +22,9 @@ Frequently Asked Questions
 
    .. note::
 
-   For Pande lab members on ``vsp-compute``, set the environment variable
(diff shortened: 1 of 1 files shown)


rmcgibbo:
@mpharrigan can we just get rid of what whole section though? with the latest release, sse4.1 support is not longer needed, just sse3.

mpharrigan:
mdtraj doesn't work on vsp-compute with default gcc 4.1

rmcgibbo:
But the whole section is titled " Why am I seeing an error message about my CPU and SSE 4.1?".

Can you maybe re-organize the question/answer so that it's up to date with the current code and issues that you encounter on centos 5?

PR #750, opened 2015-03-03
Update CentOS 5 instructions

mpharrigan:
cc #749

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -3,24 +3,17 @@
 Frequently Asked Questions
 ==========================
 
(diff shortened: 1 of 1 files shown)


mpharrigan:
I don't know what the exact error message was. I don't want to re-compile a broken installation to find out

rmcgibbo:
Why doesn't conda help here?

kyleabeauchamp:
I think my VM is centos 6, not centos 5.  It's hard to get backwards compatability that far back AFAIK

rmcgibbo:
Oh, the centos 5 thing is a glibc issue, right.

PR #757, opened 2015-03-03
test_rmsd_memmap: use tempfile to make directory for tmp files rather th...

pfrstg:
test_rmsd_memmap: use tempfile to make directory for tmp files rather than assuming current working directory is writable.


kyleabeauchamp:
lgtm

PR #759, opened 2015-03-04
[ENH] Compute nematic order parameter

ctk3b:
This compute is modeled after http://cmt.dur.ac.uk/sjc/thesis_dlc/node65.html (same as Allen & Tildesley, Ch. 11.5 and using the MOI to compute the molecular axes). This might not be directly relevant to most of the MDTraj developers' work but I use it to calculate ordering in monolayers and it's obviously also useful for bilayer and liquid crystal simulations.

Main points that still need addressing:
* Should `compute_inertia_tensor` be exposed via `__all__`? I think it might be more generally useful and could potentially be better off in a different file
*  `compute_inertia_tensor` currently assumes an unwrapped trajectory and should give you garbage otherwise
* Unit tests for inertia and Q-tensors
* There are some minor TODO's scattered throughout for which I would love any and all advice


ctk3b:
Was there a change to the linalg.eig broadcasting capabilities in recent numpy versions?

ctk3b:
I think this is the relevant update from the 1.8 release: http://docs.scipy.org/doc/numpy/release.html#support-for-linear-algebra-on-stacked-arrays

rmcgibbo:
I wasn't familiar with it, but this seems like a useful order parameter for materials, bilayers, and potentially other types of systems.

- The performance looks like it could be improved a lot. Many of the calculations iterate over the trajectory frame by frame. I think most of these could be vectorized.
- It would also be possible to do the 3x3 linear algebra in closed form, which might be faster or vectorizable, but I'm not sure about this.
- +1 to exposing `compute_inertia_tensor`.

ctk3b:
Just FYI, there might be a bug buried in here. Let me do some digging!

ctk3b:
Also in regards to iterating over all frames, for some calculations, the broadcasting over all frames doesn't work in numpy < 1.8. But there are definitely some other areas for improvement.

ctk3b:
OK this is producing correct results as best as I can tell and is consistent with code currently being used in our group. I'll find some time this weekend to flesh out unit tests and vectorize as much as possible.

leeping:
Cool. :)  I think this would be a nice contribution, and it could also be easy to copy and modify for implementing high-performance order parameters of different kinds (I typically default to looping over the trajectory.)  I think it would also be nice to add an example calculation / structure somewhere, so people know how to use it.

ctk3b:
I've vectorized the inertia tensor which was chewing up most of the time:
```python
In [24]: %timeit compute_inertia_tensor_slow(traj)
1 loops, best of 3: 3.32 s per loop

In [25]: %timeit compute_inertia_tensor(traj)
10 loops, best of 3: 47.2 ms per loop
```
Also added an example of a monolayer that I was testing with although it's not super exciting since it's very ordered.

rmcgibbo:
This is still labeled as WIP. Is that current?

mdtraj/geometry/order.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+def compute_nematic_order(traj, indices='chains'):
+    """Compute the nematic ordering of a group in every frame.
+
+    Parameters
+    ----------
+    traj : Trajectory
+        Trajectory to compute ordering in.
+    indices : str or list of lists, optional, default='chains'
```
rmcgibbo: {'chains', 'residues', list of lists}, optional, default='chains'

mdtraj/geometry/order.py
```diff
@@ -0,0 +1,263 @@
...
+from mdtraj.utils import ensure_type
+from mdtraj.utils.six import string_types
+
+
+__all__ = ['compute_nematic_order', 'compute_inertia_tensor']
+
+
+def compute_nematic_order(traj, indices='chains'):
+    """Compute the nematic ordering of a group in every frame.
+
```
rmcgibbo: Can we add a one sentence description of the meaning of the nematic order parameter, for people (like me) that are/were not familiar with it? (e.g. 'the nematic order parameter measures the tendency of molecular groups to be oriented along a particular spatial direction', if that is an accurate characterization, or something similar).

mdtraj/geometry/order.py
```diff
@@ -0,0 +1,263 @@
...
+    >>> import mdtraj as md
+    >>> from mdtraj.testing import get_fn
+    >>> traj = md.load(get_fn('monolayer.xtc'), top=get_fn('monolayer.pdb'))
+    >>> # Each of the 100 chains contains 36 atoms.
+    >>> chain_indices = [[n+x for x in range(36)] for n in range(0, 3600, 36)]
+    >>> S2 = md.compute_nematic_order(traj, indices=chain_indices)
+
+    The chains were attached to a flat, crystalline substrate and are thus
+    highly ordered with a mean S2 of ~0.996.
```
rmcgibbo: Is the value always between 0 and 1. If so, maybe that can go in the short description https://github.com/mdtraj/mdtraj/pull/759/files#r26452883

rmcgibbo:
This looks ready to merge to me. I assume that `_compute_Q_tensor` can be vectorized later if necessary.

ctk3b:
I think it's pretty much ready. My last issue is about numpy backwards compatibility and how you want to handle that. The remaining Travis cases that fail are for numpy < 1.8 because `linalg.norm` didn't use to accept an axis argument. Should I just keep making special cases that use slower versions for anyone using < 1.8 like I did [here](https://github.com/mdtraj/mdtraj/pull/759/files#diff-0c0232addf63cfda2d31dd9984cbf0e3R122) and [here](https://github.com/mdtraj/mdtraj/pull/759/files#diff-0c0232addf63cfda2d31dd9984cbf0e3R239)?

mpharrigan:
I wouldn't mind saying "this feature requires numpy >= 1.8" 

@rmcgibbo may have other thoughts

rmcgibbo:
Yeah, if we can easily work around older versions of numpy with a couple special cases, that seems easiest.

If dealing with compatibility for older numpys is or becomes too onerous accross the project, then we can raise the minimum required version. I'm not sure that this is necessary here.

ctk3b:
Alright I've added two more numpy version checks for now. Barring any further issues on Travis, I think it's good to go. I can always take the checks back out when and if the numpy requirements change.

rmcgibbo:
Thanks for all this hard work, @ctk3b.

ctk3b:
Thanks for incorporating the code :)

mpharrigan:
Wanna add it to whatsnew.rst?

ctk3b:
Sure. Any thoughts on where to group this in analysis.rst? Maybe an "Order Parameter" section? @leeping mentioned other order parameters that he uses and I'm toying with some others as well which I may end up adding.

mpharrigan:
New subsection "Order parameters"?

rmcgibbo:
Yeah, adding a new section to analysis.rst would be good -- it doesn't look like it fits well in any of the existing sections.

PR #760, opened 2015-03-04
Option to pre-compute pairs for RDF

ctk3b:
It should be possible to pre-compute your own pairs for the RDF calculation.

### DIFF
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -48,19 +48,22 @@
 ##############################################################################
 
 from __future__ import print_function, division
-import os
-import numpy as np
+
 import itertools
+import numpy as np
+import os
+import xml.etree.ElementTree as etree
+
 from mdtraj.core import element as elem
 from mdtraj.core.residue_names import _PROTEIN_RESIDUES, _WATER_RESIDUES
 from mdtraj.core.selection import parse_selection
-import xml.etree.ElementTree as etree
-
-from mdtraj.utils import ilen, import_
+from mdtraj.utils import ilen, import_, ensure_type
+from mdtraj.utils.six import string_types
 
 ##############################################################################
 # Utilities
 ##############################################################################
+
 
 def _topology_from_subset(topology, atom_indices):
     """Create a new topology that only contains the supplied indices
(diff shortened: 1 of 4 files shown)


rmcgibbo:
Not sure what the best API is here. Maybe just a single argument, and if it's a pair of strings then it does the original behavior, if it's a 2D array of integer indices, does the other?

mpharrigan:
Might I suggest:
```python
def rdf(a_indices=None, b_indices=None, pairs=None):
    _gave_indices = a_indices is not None and b_indices is not None
    _gave_pairs = pairs is not None
    assert _gave_indices xor _gave_pairs

    if _gave_indices:
         compute the pairs
    
    [do computation here]

**Comment by mpharrigan** [comment]
I.e. make the user do the selection before calling the function

**Comment by ctk3b** [comment]
@mpharrigan that sounds more robust, @rmcgibbo that sounds more elegant :)

**Comment by ctk3b** [comment]
Although @mpharrigan 's suggestion might actually be nicer for cases where generating aindices and bindices is not as simple

**Comment by ctk3b** [comment]
Could also be a combination of all of the above too: a) None: all-all, b) two strings are passed, c) a and b_indices are passed and d) pairs are passed.

**Comment by rmcgibbo** [comment]
I'm not sure what's the best thing, but in general I'd try to err on the side of simplicity, and not create too many different ways to accomplish the same thing.

**Comment by ctk3b** [comment]
Agreed. What are your thoughts on making the rdf only accept `pairs` but then providing a `generate_unique_pairs` function that users can optionally use to create those pairs?

**Comment by rmcgibbo** [comment]
That sounds fine.

**Comment by ctk3b** [comment]
How does this look? Also do you think `generate_unique_pairs` should be exposed or people will just have to know about it?

**Comment by rmcgibbo** [comment]
Maybe `generate_unique_pairs` should be a method on `Topology`, to parallel with `Topology.select`?

**Comment by rmcgibbo** [comment]
e.g.

```
class Topology(object):
    [...]

    def select_pairs(self, s1, s2):
        """Generate unique pairs of indices of atoms from two selection expressions      
        """
        return cross_product_without_duplicates(self.select(s1), self.select(s2))
```

ctk3b:
I've added a `select_pairs(selection1=None, selection2=None)` method to `Topology`. The selections can be either strings (resolved via `top.select`) or arrays of indices. 

There was also a bug that I introduced when I tried to get rid of the intermediate set creation: it (quite obviously >.<) created duplicates for cases where selection1 and selection2 are different. There's an appropriate unit test included now. I can't figure out a way to get rid of that set creation so it's back in for that case.

PR #772, opened 2015-03-18
[iterload] added skip keyword to skip first n frames.

marscher:
added a skip keyword to generate time lagged iterators. Currently the NetCDF loader complains about this new keyword, but I dont know how to fix this properly.

### DIFF
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -431,6 +431,8 @@ def iterload(filename, chunk=100, **kwargs):
         If not none, then read only a subset of the atoms coordinates from the
         file. This may be slightly slower than the standard read because it
         requires an extra copy, but will save memory.
+    skip : int, default=0
+        Skip first n frames.
 
     See Also
     --------
@@ -450,6 +452,8 @@ def iterload(filename, chunk=100, **kwargs):
     stride = kwargs.pop('stride', 1)
     atom_indices = cast_indices(kwargs.pop('atom_indices', None))
     top = kwargs.pop('top', None)
+    skip = kwargs.pop('skip', 0)
+
     extension = _get_extension(filename)
     if extension not in _TOPOLOGY_EXTS:
         topology = _parse_topology(top)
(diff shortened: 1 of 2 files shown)


mdtraj/tests/test_trajectory.py
```diff
@@ -478,6 +478,21 @@ def test_iterload():
...
+             'frame0.dcd', 'frame0.binpos', 'legacy_msmbuilder_trj0.lh5',
+             'frame0.xyz', 'frame0.lammpstrj']
+
+    for skip in [1, 2, 10, 99]:
+        for file in files:
+            t_ref = md.load(get_fn(file), top=get_fn('native.pdb'))
+            t = functools.reduce(lambda a, b: a.join(b), md.iterload(get_fn(file), skip=skip, top=get_fn('native.pdb'), chunk=100))
+            eq(t_ref.xyz[skip:], t.xyz)
+            print t_ref.time
+            print t.time
```
mpharrigan: python 3 please :)

rmcgibbo:
What is the use case for this?

mdtraj/tests/test_trajectory.py
marscher: sure, sorry :)

marscher:
* rebased on 1.4.2
* also fixed a bug in topology requiring formats with chunksize=0, where the topology is not being passed, since it is has been popped out of kwargs before the load call.

mdtraj/core/trajectory.py
```diff
@@ -460,29 +464,40 @@ def iterload(filename, chunk=100, **kwargs):
...
         t = load(filename, stride=stride, atom_indices=atom_indices)
         for i in range(0, len(t), chunk):
-            yield  t[i:i+chunk]
+            yield t[i:i+chunk]
 
     else:
         with (lambda x: open(x, n_atoms=topology.n_atoms)
               if extension in ('.crd', '.mdcrd')
               else open(filename))(filename) as f:
+            skipped = 0
```
rmcgibbo: it would be more efficient here to just use the `seek` method on the open file handle.

marscher: you're so right!

PR #776, opened 2015-03-23
Fix compute_rdf r_range argument

ctk3b:
Current code complains when a numpy array is passed instead of a list because the truth value is ambiguous.


rmcgibbo:
lgtm.

PR #781, opened 2015-03-31
Optional `n_bins` parameter for compute_rdf 

ctk3b:
Numpy's [arange](http://docs.scipy.org/doc/numpy/reference/generated/numpy.arange.html) can cause some weird issues with non-integer step sizes and of course we ran into that issue today:

```python
In [10]: np.arange(0, 3.35, (0.1 / 6)).shape
Out[10]: (201,)

In [11]: np.arange(0, 3.35 / 10, (0.1 / 6) / 10).shape
Out[11]: (202,)
```

This PR adds an option to specify `n_bins` which overrides the `bin_width` parameter and uses `linspace` instead. Also slightly modified the previous behavior to be right side inclusive when using arange so that using `bin_width` or `n_bins` both include `r_range[1]`.


mdtraj/geometry/rdf.py
```diff
@@ -67,7 +70,10 @@ def compute_rdf(traj, pairs=None, r_range=None, bin_width=0.005,
         r_range = np.array([0.0, 1.0])
     r_range = ensure_type(r_range, dtype=np.float64, ndim=1, name='r_range',
                           shape=(2,), warn_on_cast=False)
-    bins = np.arange(r_range[0], r_range[1], bin_width)
+    if n_bins:
```
mpharrigan: I think `is not None` is preferred

mdtraj/geometry/rdf.py
```diff
@@ -67,7 +70,10 @@ def compute_rdf(traj, pairs=None, r_range=None, bin_width=0.005,
         r_range = np.array([0.0, 1.0])
     r_range = ensure_type(r_range, dtype=np.float64, ndim=1, name='r_range',
                           shape=(2,), warn_on_cast=False)
-    bins = np.arange(r_range[0], r_range[1], bin_width)
+    if n_bins is not None:
```
swails: Actually, I'd go back to `if n_bins`.  `if n_bins is not None` is preferred to `if n_bins != None`, but you actually changed the meaning of your code with this change (if that's intentional, apologies).

I would intuitively think that any value for `n_bins` that evaluated to `False` (e.g., `0`, `False`, etc.) would fall back to `bin_width`.  And the error message will probably be a little strange if someone feeds an illegal number of bins (like <0 or 0)?

ctk3b: Could also just got the extra mile and assert that `n_bins` is a positive integer?   Doesn't seem very pythonic I suppose.

swails: I wouldn't do an `assert`.  Maybe something like:

```Python
if n_bins is not None:
    n_bins = int(n_bins)
    if n_bins <= 0:
        raise ValueError('n_bins must be a positive integer')
```

`assert`s get removed when you compile Python code with `-OO` (docstrings are removed when compiled with `-O`).  So you should only ever use `assert` if, in the absence of any internal bug, program behavior will never change if that line is removed (because you never know when it will behind the scenes).  That's still Pythonic, I think.

Again, apologies if I took what you said too literally and I was rehashing what you already knew.

ctk3b: No worries hahah. I think what you proposed is simple enough. We do need to trust the user to know at least a little about what they're doing :)

swails: True, but even people that know what they're doing may make a stupid mistake -- especially if it's the kind of thing where the bin number is a computed value in their script... perhaps they made the silly mistake of doing something like

```
n_bins = lower_bound - upper_bound
```

so it comes out to the negative of what they meant.  Without a check on illegal bounds, the traceback will look like this:

```Python
>>> np.histogram(np.random.rand(10), bins=[])
  File ".../python2.7/site-packages/numpy/lib/function_base.py", line 211, in histogram
    sa.searchsorted(bins[-1], 'right')]
IndexError: index -1 is out of bounds for axis 0 with size 0
```
The simple bounds check could save even an experienced user time by showing them exactly what their mistake was.  The status quo initiates a (hopefully short) bug hunt.

ctk3b: Very true. I think what you suggested fits the bill here.

rmcgibbo:
This is ready to merge, right?

ctk3b:
I believe so, yes.

Issue #206, opened 2013-11-01
Issue with DCDReporter


kyleabeauchamp:
1.  Docstrings are copy pasted from HDFTrajectory; need updating
2.  It seems to be mangling the input filenames:

```
simulation.reporters.append(mdtraj.reporters.DCDReporter("production.dcd", 200))

simulation.context.setVelocitiesToTemperature(temperature)
simulation.step(1000)
```

```
kyleb@kb-intel:~/dat/noe_data/ala-ser$ ls -lt
total 140
-rw-rw-r-- 1 kyleb kyleb 39196 Nov  1 14:38 __ii_changed
-rw-rw-r-- 1 kyleb kyleb  2069 Nov  1 14:38 nowater.pdb
-rw-rw-r-- 1 kyleb kyleb 52201 Nov  1 14:09 equil.pdb
-rw-rw-r-- 1 kyleb kyleb  2022 Nov  1 13:49 ala-ser.pdb
```

mpharrigan:
I'm getting this `fillvalue` thing now... very confusing

mpharrigan:
(run from IPython notebook)
```python
with DCDTrajectoryFile("in.dcd")) as f:
        n_runs = len(f) // chunk_size
        for _ in range(n_runs):
            run_p = "RUN{}".format(run_i)
            os.mkdir(run_p)
            
            # Read and write frames of dcd
            xyz, cell_lengths, cell_angles = f.read(chunk_size)
            with DCDTrajectoryFile("{run_p}/seed.dcd".format(run_p=run_p), 'w') as wf:
                wf.write(xyz, cell_lengths, cell_angles)
```

kyleabeauchamp:
Hmm, I just re-ran my original code above with no success at reproducing.  This will be a tricky one...

mpharrigan:
Calling `.encode('ascii')` on the filename fixes it

PR #784, opened 2015-03-31
BUG: Fix garbage in DCD out filename

mpharrigan:
cc #206


kyleabeauchamp:
Can you explain a bit more about what's going on here?  Is it because the string gets garbage collected or something?

mpharrigan:
I think so. The way I could reliably reproduce was passing the result of a `str.format()`, whereas normal strings seemed to work.

It looks like the str.format was getting garbage collected and it was picking a random string from memory, usually "fillvalue" but sometimes other things

kyleabeauchamp:
+1

PR #785, opened 2015-04-02
Support gzip for reading/writing xyz files.

pfrstg:


swails:
Should we add some kind of generalized file opener that automatically opens .bz2 and .gz files with the correct file handler?  It seems kinda crazy to do this for each ASCII-based trajectory type...

swails:
Also, in Python3 `gzip.open` opens a file in binary mode, which means that everything read from it are bytes that must be decoded (so the Py3 tests will likely fail with encoding-related exceptions like `AttributeError` or `ValueError`).  Whereas opening as ASCII returns `str` when read...

kyleabeauchamp:
Long run I agree, in the short run there are lots of different edge cases that I remember struggling with:

HTTP PDB requests
py2 vs py3

Might be easiest to get each subproblem working, have tests, and then refactor as a separate PR.  just my two cents

swails:
@kyleabeauchamp -- agreed.  I didn't actually mean in this PR (although it would be nice to add support for .bz2 files here, since it's an additional 2 or 3 lines).

pfrstg:
I agree that this should be a more general thing but I'm happy to defer that refactoring :-)

The API docs for gzip suggest TextIOWrapper which suggests we'd want somethign like
self._file = io.TextIOWrapper(gzip.open(filename, 'w'))

It looks like that should work in python 2.7 as well. Thoughts? Want that added to this PR?

pfrstg:
Actually, I have to hack the file interface on another format as well so I think I may do the refactoring to handle .gz. and .bz2 for all file types. I probably won't try to make the same utility function handle URLs.

So I'm going to close these two PRs. I'm on vacation next week -- I'll get to this when I am back.

rmcgibbo:
sounds good. have a good vacation!

PR #786, opened 2015-04-02
Support writing .pdb.gz files

pfrstg:
Also change how temp file names are handled in test_trajectory to not use arbitrary indices.

### DIFF
--- a/mdtraj/tests/test_trajectory.py
+++ b/mdtraj/tests/test_trajectory.py
@@ -48,7 +48,7 @@ for suffix, (fd, temp) in {
       'nc' : tempfile.mkstemp(suffix='.nc'),
       'lh5' : tempfile.mkstemp(suffix='.lh5'),
       'lammpstrj' : tempfile.mkstemp(suffix='.lammpstrj'),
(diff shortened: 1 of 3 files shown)


mdtraj/tests/test_trajectory.py
```diff
@@ -36,25 +36,27 @@
...
+      'xyz' : tempfile.mkstemp(suffix='.xyz')}.iteritems():
+    os.close(fd)
+    tmpfns[suffix] = temp
 
 def teardown_module(module):
     """remove the temporary file created by tests in this file
     this gets automatically called by nose"""
 
-    for e in [temp1, temp2, temp3, temp4, temp5, temp6, temp7, temp8]:
+    for e in tmpfns.itervalues():
```
rmcgibbo: no `itervalues` on py3. `.values` is fine, since this is a trivial amount of memory.

pfrstg:
Done (fixed my iteritems also)

pfrstg:
I'm going to leave this open because the refactoring in test_trajectory is a good idea and the omission of writing .pdb.gz is just broken now.

The better more general gz support will be a future PR.

PR #788, opened 2015-04-03
load_hoomdxml and tests for load_hoomdxml

tcmoore3:
Add support for using a HOOMD-Blue XML file to load topology information. Since HOOMD wasn't developed with the biophysics community in mind, there is explicit no chain/residue information in the xml file. To get around this, the bond list is used to find the chains, assuming no interchain bonds.


rmcgibbo:
Cool! I haven't used HOOMD, but it looks like a nice tool. A couple small things
-  Guard the import of networkx using the same pattern as https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/topology.py#L454, since it's an optional dependency but not a required dependency.
- Add the function to the docs, e.g. here https://github.com/mdtraj/mdtraj/blob/master/docs/load_functions.rst

tcmoore3:
It appears the networkx package never gets installed on Travis, causing the tests to fail. Do you have a preference between adding it to the conda build (like e.g., the optional dependencies scipy, pandas, and pytables) or coming up with a workaround for for Travis?

mpharrigan:
You can add it to the "test" section of the conda recipe

https://github.com/mdtraj/mdtraj/blob/master/devtools/conda-recipe/meta.yaml#L28

PR #789, opened 2015-04-06
fix IPython 3.0 compatibility issues

golobor:
Hi, 
the new version of IPython 3.0 introduced [some changes] (https://github.com/ipython/ipython/wiki/IPython-3.0-comm-and-widget-migration-document) into the widget interface, so that TrajectoryView notebook widget stopped working. This commit seems to fix all of the issues and restores TrajectoryView functionality in the notebook.

Anton.

### DIFF
--- a/mdtraj/html/install.py
+++ b/mdtraj/html/install.py
@@ -50,9 +50,10 @@ def enable_notebook():
     """
     libs = ['iview.js','surface.min.js','objexporter.js','filesaver.js','context.js']
     fns = [resource_filename('mdtraj', os.path.join('html', 'static', f)) for f in libs]
-    install_nbextension(fns, verbose=0)
+    for fn in fns:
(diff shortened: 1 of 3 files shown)


rmcgibbo:
Awesome! Thanks so much for fixing this. Just tested on my machine and it works.

golobor:
Sure! Actually, thank _you_ for making such an awesome tool! :)

rmcgibbo:
I tried running this with IPython v2, and as expected, it doesn't work. I should create a page in the docs about this feature, and it can say that IPython v3 is the minimum supported version there.

golobor:
ouch. That makes sense. The good thing is that IPython 3.0 is a great release with a lot of new features and no downsides (besides the issues with custom widgets), so it's a reasonable thing to ask.

PR #798, opened 2015-04-17
NMR shift wrapper fix for >= Pandas 0.14.0

### LINKED ISSUE #797: NMR shift wrappers are broken with Pandas 0.16
Pretty minor issue pops up in Pandas 0.16 due to deprecating some vintage argument names.
```
  File "/Users/cing/anaconda/lib/python2.7/site-packages/mdtraj-1.4.0.dev0-py2.7-macosx-10.5-x86_64.egg/mdtraj/nmr/shift_wrappers.py", line 162, in chemical_shifts_shiftx2
    results = results.pivot_table(rows=["resSeq", "name"], cols="frame", values="SHIFT")
TypeError: pivot_table() got an unexpected keyword argument 'rows'
```
One line fix, rows to "index" and cols to "columns" fixes it. I'll submit a PR for due diligence, but now you can't claim to support "Pandas Version >= 0.9.0", the earliest you can do it 0.14.0!

cing:
Nothin' fancy, fixes #797

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -46,8 +46,8 @@ Optional packages:
         We use scipy for loading and saving AMBER netcdf formatted
         trajectories.
 
-    `Pandas <http://pandas.pydata.org>`_ >= 0.9.0
-        Some functionality, including mol2 parsing,  requires pandas.
+    `Pandas <http://pandas.pydata.org>`_ >= 0.14.0
+        Some functionality, including mol2 parsing and computing NMR observables requires pandas.
 
     `PyTables <http://www.pytables.org/>`_ >= 2.4.0
         Working with HDF5 formatted trajectories requires the PyTables
--- a/mdtraj/nmr/shift_wrappers.py
+++ b/mdtraj/nmr/shift_wrappers.py
@@ -159,7 +159,7 @@ def chemical_shifts_shiftx2(trj, pH=5.0, temperature=298.00):
             results.append(d)
 
     results = pd.concat(results)
-    results = results.pivot_table(rows=["resSeq", "name"], cols="frame", values="SHIFT")
+    results = results.pivot_table(index=["resSeq", "name"], columns="frame", values="SHIFT")
     return results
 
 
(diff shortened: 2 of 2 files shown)


rmcgibbo:
1. The test failure here comes from the fact that most of the travis matrix is using numpy 1.6 and numpy 1.7, for which the continuum channels don't have a recent enough pandas.
2. For omnia, we produce binaries for numpy 1.8 and numpy 1.9. These are the versions that continuum is building new packages against its conda binaries (do `conda search pandas`). Some new continuum conda binaries are only built against numpy 1.9
3. I think we mind as well up the travis spec to test against numpy 1.8 and numpy 1.9, which would fix the test failure here. We can write on the dependency page that numpy >= 1.8 is recommended, but older versions may work (but are untested).
4. what do you think @kyleabeauchamp?
5. sorry @cing that this kind of complicates this otherwise simple PR.

rmcgibbo:
An alternative is simply to special-case the behavior in this PR for pandas < 0.14 and pandas >= 0.14, but I think that's probably the wrong approach. With only a limited testing grid to cover multiple versions of python, numpy, pandas, and all of our other dependencies, there's no way we can get full test coverage of different combinations of old versions. It's also messy and hard to maintain special cases like that.

swails:
I'm late to the party here, but I think it would be easy enough to keep support for both -- just switch to using `groupby` rather than `pivot_table`.  I'll issue a PR to @cing's branch doing this.

swails:
Actually, I don't think `groupby` does exactly the same thing that `pivot_table` does.  There are only 2 places where `pivot_table` is used, so it wouldn't be *too* much of a maintenance burden to work around it.  pandas 0.14 was only released a year ago (not quite), so dropping support for pandas <0.14 would break mol2 parsing and chemical shift wrappers from any Pandas installation more than a year old (not a problem for `conda

rmcgibbo:
Closing in favor of #811, which also fixes this bug.

Issue #599, opened 2014-09-17
Output formats in registry


rmcgibbo:
We have a registration mechanism for dispatching to the right input file loader based on the filename extension, but we don't have anything similar for output files. There's just as dict inside `Trajectory.save` that maps extensions to the appropriate saver method. https://github.com/SimTk/mdtraj/blob/master/MDTraj/core/trajectory.py#L1123

This makes it hard, for example, to programmatically list all of the supported filename extensions that MDTraj can save output trajectories in.

We should unify this with the format registry code used in the input side.

PR #799, opened 2015-04-17
ENH: Add save_gro

mpharrigan:
cc #599

### DIFF
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -46,6 +46,7 @@ from mdtraj.formats import ArcTrajectoryFile
 from mdtraj.formats import DTRTrajectoryFile
 from mdtraj.formats import LAMMPSTrajectoryFile
 from mdtraj.formats import XYZTrajectoryFile
+from mdtraj.formats import GroTrajectoryFile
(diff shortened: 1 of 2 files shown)


mpharrigan:
wt2

VMD refuses to load the saved file unless I delete the trailing 6 box vector specifications, and then it looks crazy

mpharrigan:
For > 100k atoms, gromacs seems to drop the first digit of atom id, and vmd is ok with that. mdtraj expands the column to 6 characters which messes everything up

mpharrigan:
Added truncation

rmcgibbo:
lgtm.

rmcgibbo:
> VMD refuses to load the saved file unless I delete the trailing 6 box vector specifications, and then it looks crazy

Does this mean something is wrong?

mpharrigan:
With the indexing fix that should now read: "VMD refuses to load the saved file unless I delete the trailing 6 box vector specifications, and then it looks fine"

The gro format spec I found said that those six values should be legit. I don't know why VMD has a problem with them

rmcgibbo:
Okay, merging then.

Issue #800, opened 2015-04-18
What is `atom.serial`


mpharrigan:
And how is it different than `atom.index`?

This doesn't seem to be documented anywhere

rmcgibbo:
For the PDB format, it's supposed to be the number associated in the pdb file with the atom, called "serial" [here](http://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM). These aren't always contiguous or starting with zero.

mpharrigan:
Maybe we can add at least a little comment in the code

rmcgibbo:
+1

PR #802, opened 2015-04-19
DOC: Comment about atom.serial

mpharrigan:
cc #800


rmcgibbo:
What about in the docs too?

mpharrigan:
Added. p.s. github web editor is pretty snazzy

rmcgibbo:
yeah, it makes it pretty easy.

PR #804, opened 2015-04-21
Make vmd-readable .gro

mpharrigan:
VMD Doesn't like two spaces before the number of atoms.
It fails with `gromacsplugin) Warning, error reading box, unexpected
end-of-file reached`.


mpharrigan:
I'm going to merge this

PR #808, opened 2015-04-24
Add .ncdf as a recognized suffix for the NetCDF loader

### LINKED ISSUE #807: NetCDF loader not registered for ".ncdf"
the save routine for netcdf creates files with that suffix, but the loader is not registered for it. So I guess the writer is registered, but the loader is not.

t = Trajectory(...)
t.save('foo.ncdf')
mdtraj.load('foo.ncdf')

yields IOError: Sorry, no loader for filename=foo.ncdf

swails:
It's supported in Trajectory.save, and it's not like another format has any
kind of claim on that suffix.

Closes #807


rmcgibbo:
Lgtm

PR #810, opened 2015-04-24
Support .gz and .bz2

pfrstg:
Factored out code into io to support this.

I've only made .xyz and .pdb support these. What others formats should I also change?


pfrstg:
I'm also not 100% sure about the python3 support (we use python2 internally) -- do you have tests that will catch problems here?

kyleabeauchamp:
The current Travis grid should catch most of the problems, although I recall previously wrestling with py2/3 issues on the pdb.gz.

rmcgibbo:
The tests look complete enough that they'll catch this on py3. There might be some string/bytes issues -- this is a tricky bit.

rmcgibbo:
@pfrstg: I'll do the py3 fixes, and then submit a PR into your branch.

rmcgibbo:
Struggling to get git to work to file a PR into your branch. Here are the modifications that make this PR work with PY3.

https://github.com/rmcgibbo/mdtraj/commit/3ea1c949006310d95f68f4223be884e44502ebe2

rmcgibbo:
```
diff --git a/mdtraj/io.py b/mdtraj/io.py
index dfdcfd5..f3289fb 100644
--- a/mdtraj/io.py
+++ b/mdtraj/io.py
@@ -74,14 +74,15 @@ True
 Functions
 ---------
 """
-from __future__ import print_function, division
+from __future__ import print_function, division, absolute_import
+import io
 import bz2
 import gzip
 import os
 import warnings
 import numpy as np
 from mdtraj.utils import import_
-from mdtraj.utils.six import PY3, iteritems, StringIO
+from mdtraj.utils.six import PY2, PY3, iteritems, StringIO
 if PY3:
     basestring = str
 tables = import_('tables')
@@ -291,6 +292,25 @@ def loadh(file, name=Ellipsis, deferred=True):
 
 
 def open_maybe_zipped(filename, mode, force_overwrite=True):
+    """Open a file in text (not binary) mode, transparently handling
+    .gz or .bz2 compresssion, with utf-8 encoding.
+
+    Parameters
+    ----------
+    filename : str
+        Path to file. Compression will be automatically detected if
+        the filename ends in .gz or .bz2.
+    mode : {'r', 'w'}
+        Mode in which to open file
+    force_overwrite: bool, default=True
+        If 'w', should we overwrite the file if something with `filename`
+        already exists?
+
+    Returns
+    -------
+    handle : file
+        Open file handle.
+    """
     _, extension = os.path.splitext(filename.lower())
     if mode == 'r':
         if extension == '.gz':
@@ -305,11 +325,18 @@ def open_maybe_zipped(filename, mode, force_overwrite=True):
         if os.path.exists(filename) and not force_overwrite:
             raise IOError('"%s" already exists' % filename)
         if extension == '.gz':
-            return gzip.GzipFile(filename, 'w')
+            if PY2:
+                return gzip.GzipFile(filename, 'w')
+            binary_fh = gzip.GzipFile(filename, 'wb')
+            return io.TextIOWrapper(binary_fh, encoding='utf-8')
         elif extension == '.bz2':
-            return bz2.BZ2File(filename, 'w')
+            if PY2:
+                return bz2.BZ2File(filename, 'w')
+            binary_fh = bz2.BZ2File(filename, 'wb')
+            return io.TextIOWrapper(binary_fh, encoding='utf-8')
         else:
             return open(filename, 'w')
+
     else:
         raise ValueError('Invalid mode "%s"' % mode)
 
diff --git a/mdtraj/tests/test_io.py b/mdtraj/tests/test_io.py
index 25ae2f9..4f0bcc1 100644
--- a/mdtraj/tests/test_io.py
+++ b/mdtraj/tests/test_io.py
@@ -24,6 +24,8 @@ import bz2
 import glob
 import gzip
 import os
+import gzip
+import bz2
 import os.path
 import tables
 import tempfile
@@ -87,28 +89,28 @@ class test_open_maybe_zipped(unittest.TestCase):
     def test_read_gz(self):
         fn = os.path.join(self.tmpdir, 'read.gz')
         with gzip.GzipFile(fn, 'w') as f:
-          f.write(u'COOKIE')
+          f.write('COOKIE'.encode('utf-8'))
         eq(io.open_maybe_zipped(fn, 'r').read(), u'COOKIE')
 
     def test_write_gz(self):
         fn = os.path.join(self.tmpdir, 'write.gz')
-        f = io.open_maybe_zipped(fn, 'w')
-        f.write(u'COOKIE')
-        f.close()
+        with io.open_maybe_zipped(fn, 'w') as f:
+            f.write('COOKIE')
+
         with gzip.GzipFile(fn, 'r') as f:
           eq(f.read().decode('utf-8'), u'COOKIE')
 
     def test_read_bz2(self):
         fn = os.path.join(self.tmpdir, 'read.bz2')
         with bz2.BZ2File(fn, 'w') as f:
-          f.write(u'COOKIE')
+          f.write('COOKIE'.encode('utf-8'))
         eq(io.open_maybe_zipped(fn, 'r').read(), u'COOKIE')
 
     def test_write_bz2(self):
         fn = os.path.join(self.tmpdir, 'write.bz2')
-        f = io.open_maybe_zipped(fn, 'w')
-        f.write(u'COOKIE')
-        f.close()
+        with io.open_maybe_zipped(fn, 'w') as f:
+            f.write('COOKIE')
+
         with bz2.BZ2File(fn, 'r') as f:
           eq(f.read().decode('utf-8'), u'COOKIE')
```

swails:
It seems like the context manager protocol in `gzip.GzipFile` is only supported in Py27 and above.

pfrstg:
And you aim to support python < 2.7? I can remove that style of usage....

swails:
I think the targets are Python 2.6+.  The last Py26 release (2.6.9) was in October 2013 (although there will be no more).

swails:
You could also monkey-patch `gzip.GzipFile` and `bz2.BZ2File` and add `__enter__` and `__exit__` methods that do the right thing, but as far as I can tell you only used that style in the tests, so avoiding it is probably easier.

rmcgibbo:
I'm fine removing py26 support. http://www.curiousefficiency.org/posts/2015/04/stop-supporting-python26.html

swails:
First step is to remove it from the build matrix, then :).

kyleabeauchamp:
Well that was an anticlimactic response, I was hoping for a showdown on legacy python support...

swails:
Haha.  I'm with @rmcgibbo on this one.  Most OSes have moved up, or have an option to install Python 2.7.  For those that don't, Anaconda or a simple "./configure && make install" does the trick.

None of the devs use Python 2.6, which is probably the best reason to drop support.

kyleabeauchamp:
+1

pfrstg:
PR updated with rmcgibbo patch.

Left in the with clauses that will break <2.7

rmcgibbo:
Is a file missing? (from the travis-ci test log)

```
======================================================================
ERROR: mdtraj.tests.test_xyz.test_read_gz
----------------------------------------------------------------------
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/mdtraj/tests/test_xyz.py", line 55, in test_read_gz
    traj = md.load(get_fn('frame0.xyz.gz'), top=get_fn('native.pdb'))
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/mdtraj/testing/testing.py", line 98, in get_fn
    'added it, you\'ll have to re install' % fn)
ValueError: Sorry! /home/travis/miniconda/envs/_test/lib/python2.7/site-packages/mdtraj/testing/reference/frame0.xyz.gz does not exists. If you just added it, you'll have to re install
```

mdtraj/io.py
```diff
@@ -72,11 +75,13 @@
 ---------
 """
 from __future__ import print_function, division
```
rmcgibbo: this also needs `import io`, and probably `from __future__ import absolute_import`.

pfrstg:
Added missing file and missing imports. 2.6 will remain broken, but I'll check travis when it's done.

pfrstg:
py3 still broken -- will fix it....

pfrstg:
Python 2.7 and 3.x are all happy with this PR now

PR #832, opened 2015-05-21
Fix Gromacs box vectors

leeping:
This PR re-orders the unit cell vectors that are parsed from `gro.py`, because the previous version was assigning the values incorrectly.


rmcgibbo:
Can we add a test?

rmcgibbo:
Closing in favor of #836.

PR #833, opened 2015-05-21
Calculate some more NMR J-couplings

leeping:
I believe these formulas were in the same paper that was cited in the original function `compute_J3_HN_HA()`, but it's been so long that I don't really remember.


rmcgibbo:
+1, needs test tho.

kyleabeauchamp:
I just checked and these formulas are from the Bax2007 paper table 1.

rmcgibbo:
Closing for #837, which contains the same commit with tests.

PR #835, opened 2015-05-26
Make virtual site a new element

swails:
This aims to make it so no element will raise a `NameError` when a virtual site is present (since that is fairly rare so far).  The goal is to make it just as easy to identify them (`atom.element is elem.virtual` instead of `atom.element is None`) while removing the need to special-case virtually every element inspection for virtual sites before trying to access an attribute of the element.

### DIFF
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -283,7 +283,11 @@ class Topology(object):
             for residue in chain.residues:
                 r = out.addResidue(residue.name, c)
                 for atom in residue.atoms:
-                    a = out.addAtom(atom.name, app.Element.getBySymbol(atom.element.symbol), r)
+                    if atom.element is elem.virtual:
+                        element = None
+                    else:
+                        element = app.Element.getBySymbol(atom.element.symbol)
+                    a = out.addAtom(atom.name, element, r)
                     atom_mapping[atom] = a
 
         for a1, a2 in self.bonds:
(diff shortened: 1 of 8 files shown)


mpharrigan:
+1 for the idea (haven't looked at code yet)

swails:
ping @rmcgibbo -- I think you were in Korea when I submitted this PR?

What do you think about the basic idea?

rmcgibbo:
thanks for the ping. back now.

rmcgibbo:
+1 i'm good to merge. hopefully this will be easier to maintain than the `None`.

rmcgibbo:
Any idea what will happen with topologies that have been saved into MDTraj h5 files containing vsites that are reloaded using the new code after this merge?

swails:
> Any idea what will happen with topologies that have been saved into MDTraj h5 files containing vsites that are reloaded using the new code after this merge?

I think it should just work.  Passing `None` into the element when adding an atom gets translated into `virtual`.  Unless that code doesn't use `add_atom` to build a `Topology`?

rmcgibbo:
Okay sounds good.

mdtraj/core/topology.py
```diff
@@ -572,6 +568,7 @@ def add_atom(self, name, element, residue, serial=None):
         atom : mdtraj.topology.Atom
             the newly created Atom
         """
+        if element is None: element = elem.virtual
```
swails: Here -- if the element is passed in as None, it gets turned into virtual.

swails:
Is this worth bringing up in the OpenMM repo?  The same thing is done there, too, except the use of `None` is probably a bit more ingrained than it is here.  And it needs to be special-cased more frequently in modeling tasks than in trajectory analysis tasks, so perhaps it's a benefit in OMM while it was just a hassle here...

rmcgibbo:
I don't really have strong feelings one way or the other.

PR #843, opened 2015-06-04
[core.trajectory] add hash implementation

marscher:
This might be useful for caching purposes of Trajectory objects. One can also efficiently compare two trajectories for equality. If you do not consider this useful, just close the PR.

### DIFF
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -846,11 +846,9 @@ class Trajectory(object):
 
     def __hash__(self):
         def _hash_numpy_array(x):
-            x.flags.writeable = False
             hash_value = hash(x.shape)
             hash_value ^= hash(x.strides)
-            hash_value ^= hash(x.data)
-            x.flags.writeable = True
+            hash_value ^= hash(x.data.tobytes())
             return hash_value
 
         hash_value = hash(self.top)
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -28,7 +28,6 @@
 from __future__ import print_function, division
 import os
 import warnings
-import functools
 from copy import deepcopy
 from collections import Iterable
 import numpy as np
@@ -42,7 +41,6 @@ from mdtraj.formats import NetCDFTrajectoryFile
 from mdtraj.formats import LH5TrajectoryFile
 from mdtraj.formats import PDBTrajectoryFile
 from mdtraj.formats import MDCRDTrajectoryFile
-from mdtraj.formats import ArcTrajectoryFile
 from mdtraj.formats import DTRTrajectoryFile
 from mdtraj.formats import LAMMPSTrajectoryFile
 from mdtraj.formats import XYZTrajectoryFile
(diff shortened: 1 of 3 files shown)


rmcgibbo:
This seems like a good feature. It needs a test, and there's a python3 failure.
```
======================================================================
ERROR: mdtraj.tests.test_trajectory.test_slice2
----------------------------------------------------------------------
  File "/home/travis/miniconda/envs/_test/lib/python3.3/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/_test/lib/python3.3/site-packages/mdtraj/tests/test_trajectory.py", line 124, in <lambda>
    yield lambda: t[0] == t[[0,1]][0]
  File "/home/travis/miniconda/envs/_test/lib/python3.3/site-packages/mdtraj/core/trajectory.py", line 888, in __eq__
    return self.__hash__() == other.__hash__()
  File "/home/travis/miniconda/envs/_test/lib/python3.3/site-packages/mdtraj/core/trajectory.py", line 879, in __hash__
    hash_value = hash(self.top)
TypeError: unhashable type: 'Topology'
```

marscher:
The error message states, that we would also need to hash Topology,
which might be slow, since it is completely object orientated, which
would involve hashing every chain, residue and atom. The hash for the
topology instance has to be cached though.

Have to think about a good test.

rmcgibbo:
Under most circumstances, I think about the trajectory as conceptually immutable. Obviously t's not actually immutable. You can change any of the numpy arrays however you like, and some methods do modify it (`center_coordinates`, `atom_slice` or `remove_solvent` with `inplace=True`, which is not the default).

But for most caching purposes, wouldn't just using the `id` of the object be sufficient?

marscher:
Am 05.06.2015 um 09:51 schrieb Robert T. McGibbon:
> Under most circumstances, I think about the trajectory as conceptually
> immutable. Obviously t's not actually immutable. You can change any of
> the numpy arrays however you like, and some methods do modify it
> (|center_coordinates|, |atom_slice| or |remove_solvent| with
> |inplace=True|, which is not the default).
So it is better to recalculate it.
> 
> But for most caching purposes, wouldn't just using the |id| of the
> object be sufficient?
By doing so, you will say two content identical objects are not equal. I
think using the id in the hash calculation is the standard behavior of
object, which might cause a lot of collisions.

rmcgibbo:
> So it is better to recalculate it.

Yeah, that's probably less risky.

And I don't think there's really any harm in going through the whole topology tree to hash the topology. The `__eq__` method already needs to do this. It's not that slow. No reason to prematurely optimize this.

mdtraj/core/trajectory.py
```diff
@@ -112,6 +110,21 @@ def _assert_files_or_dirs_exist(names):
...
+        hash_value ^= hash(x.strides)
+        hash_value ^= hash(x.data.tobytes())
+        return hash_value
+else:
+    def _hash_numpy_array(x):
+        x.flags.writeable = False
+        hash_value = hash(x.shape)
+        hash_value ^= hash(x.strides)
+        hash_value ^= hash(x.data)
+        x.flags.writeable = True
```
rmcgibbo: can we restore the writeable flag back to its original setting instead of unconditionally making it true?

rmcgibbo:
I fixed the merge conflicts in #898, and i'll merge it shortly.

Issue #842, opened 2015-06-03
Loading .gro in chunks


OndrejMarsalek:
I wanted to read a large .gro file in chunks using `iterload`. Getting the first chunk takes a very long time and it seems like it is trying to parse the whole file. Using a shorter file (1 frame) is fine. I would expect either the same behavior as with other formats or an error message.

As a smaller issue, `iterload` did not accept the `top` keyword argument when called with a .gro file. It would be nice to be able to use the same code regardless of the format. I guess that would mean either ignoring the provided topology or using it to override the one parsed from the .gro file.

rmcgibbo:
> Getting the first chunk takes a very long time and it seems like it is trying to parse the whole file. Using a shorter file (1 frame) is fine. I would expect either the same behavior as with other formats or an error message.

This issue covers the same ground as #809. `iterload` currently only has an optimized code path for dcd, xtc, h5, and lh5. For the other formats, it currently just loads the whole thing into memory and then yields the chunks.

The easy way to fix this is to add another `elif` statement before https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/trajectory.py#L506-L509. The more complete way to fix this is to make that section of code more generic so that each format doesn't need to be special-cased. It anyone wanted to take a stab at implementing either of these, that would be great.

> As a smaller issue, `iterload` did not accept the top keyword argument when called with a .gro file.

Right. It looks like there's a bug here: https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/trajectory.py#L367. That line should be more inclusive, and cover the same formats as https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/trajectory.py#L126-L127.

rmcgibbo:
This is fixed now.

Issue #809, opened 2015-04-24
Does iterload load complete files into memory before chunking them?


marscher:
I wonder if this the case for formats not explicitly handled in this function (like xtc, dcd, lh5 and h5).

in core.Trajectory.py:

            t = load(filename, **kwargs)
            for i in range(len(t), chunk):
                yield t[i:i+chunk]

rmcgibbo:
Yes, that should probably be extended to cover more formats.

marscher:
I guess the infrastructure to do so is already there for most formats?

rmcgibbo:
yeah, there's really nothing special about the (dcd, xtc, lh5, h5) loaders -- all the infrastructure is there for the other formats too.

rmcgibbo:
This is now fixed for all formats.

Issue #794, opened 2015-04-13
FR: Extend `load_frame` to accept a list of indices


rmcgibbo:
Extend `load_frame` to support loading more than one frame from a trajectory

```
frames = md.load_frame('trajectory.xtc', [1,10,100])
```

This would be identical to loading the entire trajectory, and then slicing out these particular frames, e.g.

```
frames == md.load('trajectory.xtc')[[1,10,100]]
```

but more memory efficient.

This feature would further https://github.com/msmbuilder/msmbuilder/issues/543

PR #844, opened 2015-06-05
WIP: Refactor iterload, add read_as_traj

rmcgibbo:
Refactoring `iterload` to address the root cause of #842, #809. Also will probably go a long way towards #794.

The root cause of the issue is that the code for converting from the returned data from `XYZTrajectoryFile.read()`, which is a bunch of numpy arrays that depend on the trajectory format, to a `Trajectory` object, is split up over many functions and generally copy-paste duplicated.

This adds a new method, `read_as_traj`, to each of the `XYZTrajectoryFile` objects which looks like read, but returns a trajectory object instead of a bunch of numpy arrays. All of the different loaders can then go through this method to avoid duplication, including `iterload`.


rmcgibbo:
This PR also fixes a bug with multiline gro files. A newline was missing.

rmcgibbo:
This PR makes one change to the API of `LAMMPSTrajectoryFile.write` and `XYZTrajectoryFile.write` which brings them into line with all of the other formats. For each formats, the quantities returned by `.read()` and accepted by `.write()` should be in identical units, which is the unit system of _that specific trajectory file_, not necessarily the MDTraj unit system. This means that for trajectory formats that use angstroms (like LAMMPS and XYZ, which use angstroms), the `write` method needs to accept coordinates in units of angstroms, not nanometers.

This behavior is consistent with the idea that these `ABCTrajectoryFile` classes are supposed to be low-level. If you want the unit handling done for you, that's done by the higher-level APIs that use the module level functions, `md.load()`, `md.iterload()`, `md.load_xyz`, `md.load_lammpstrj`, and the save methods on the trajectory object (`Trajectory.save_xyz`, `Trajectory.save_lammpstrj`, etc).

It's kind of a judgement call whether this is a good API or not, but it's currently how 9 of the 11 low-level `ABCTrajectoryFile` classes that use angstroms work. This fixes the remaining two formats to bring them in line.

Issue #845, opened 2015-06-05
Units bugs in LAMMPS trajectory loader


rmcgibbo:
There are some units bugs in the LAMMPS code. On the load side, it converts distances from angstroms to nanometers, but doesn't convert the unitcell lengths. On the save side, I don't think it does any conversion.

https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/lammpstrj.py#L114-L115

ctk3b:
This may not be structured particularly intelligently but I think it is happening correctly:

On write: https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/lammpstrj.py#L453
On read: https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/lammpstrj.py#L295

Is there a particular test that this is breaking? 

Semi-related, there IS a bug that I was literally just looking at. The reader assumes the ordering of columns [here](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/lammpstrj.py#L344) to be the default for the format. Users can, however, totally customize this so it should be replaced with a lookup. I'm going to submit a PR later tonight for that but if there are any other issues, I can try to address those too.

rmcgibbo:
Oh, I guess you're right. I'm refactoring some stuff for #844 -- I'm going to modify it to follow the general pattern that all of the other formats use (conversion called in `Trajectory.save_xyz` on write, and in currently in `load_xyz` or read, but I'm moving that to a new method called `read_as_traj`).

PR #847, opened 2015-06-05
ENH: Enable .lammpstrj to parse custom column orders

ctk3b:
Users can technically [customize lammps trajectories](http://lammps.sandia.gov/doc/dump.html) to contain any information in any order. This PR removes any assumptions about column ordering.


rmcgibbo:
Do you have any small files laying around using a custom / nonstandard header that we can add as a test?

ctk3b:
I added one. However, there's a unit issue. I think you may have been correct about a unit bug in #845  afterall. I'm about to get on a plane but hopefully #844 fixes that. Sorry about that. Looking back on how I organized that unit conversion code is quite painful...

rmcgibbo:
I restarted the build on master after merging #844.

ctk3b:
Looks fixed. Thank you!

PR #860, opened 2015-06-09
Fix #858

### LINKED ISSUE #858: Weird test failure?
@ctk3b do you have any idea what this test failure is? looking at the build history, it looks stochastic which is really weird 
https://travis-ci.org/mdtraj/mdtraj/jobs/66089453

swails:


ctk3b:
So it's quite clear why what I did was wrong but why did it not raise a RuntimeError? Typically when you try to modify any collection during iteration, it does raise this.

E.g.: 
```python
foo = {1:'a', 2:'b', 3:'c', 4:'d'}
for bar in foo:
    foo.pop(bar) 
RuntimeError                              Traceback (most recent call last)
<ipython-input-4-308c3003fd29> in <module>()
----> 1 for bar in foo:
      2     foo.pop(bar)
      3 

RuntimeError: dictionary changed size during iteration

**Comment by swails** [comment]
You probably added a key at the same time you deleted one.  So maybe the size didn't change, and that's all that Python checks?

**Comment by ctk3b** [comment]
That would explain it. Thank you.

**Comment by swails** [comment]
> That would explain it. Thank you.

No problem.  It's a hand-wavy explanation though, you'd have to look in the Python source code to verify.  But it at least appears to work that way in my tests:

```Python
>>> l = {'foo' : 0, 'bar' : 1, 'baz' : 2}
>>> for key in l:
...     if key == 'foo':
...             l[key[0]] = l.pop(key)
... 
>>> l
{'baz': 2, 'bar': 1, 'f': 0}
```

This dict is too small to reproduce the error you were seeing, though.  Very subtle :).

ctk3b:
Ah! I actually just watched this the other day which explains that behavior: https://www.youtube.com/watch?v=C4Kc8xzcA68

Should've have tipped me off but alas...

rmcgibbo:
Good work team!

swails:
I just checked the Python source, and as the error message states, it just checks the dict size during iteration.  Probably too expensive to *also* check that all the keys haven't changed.

PR #861, opened 2015-06-09
Lammpstrj column ordering fixes

ctk3b:
This should hopefully fix two issues:
* ensure that coordinates of different styles don't get mixed if multiple styles are present
* ensure the predictability of the style of coordinate that gets read in from a file with multiple styles

@swails does this look like a decent solution to what we talked about in #858?

### DIFF
--- a/mdtraj/formats/lammpstrj.py
+++ b/mdtraj/formats/lammpstrj.py
@@ -199,8 +199,8 @@ class LAMMPSTrajectoryFile(object):
         --------
         read : Returns the raw data from the file
 
-        Note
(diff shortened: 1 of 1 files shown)


swails:
If I understand the intent correctly, this looks much better.

ctk3b:
The point you brought up about predictability was something that definitely needed to be fixed. This solution happens to also provide a check for the fringe case where someone is not outputting all 3 coordinates of the same style.

swails:
Then this is definitely the right approach IMO.

Issue #863, opened 2015-06-11
[ENH] SliderTrajectoryView


cxhernandez:
As suggested by @mpharrigan in https://github.com/rmcgibbo/mdtraj-paper/issues/14#issuecomment-110950235

cxhernandez:
Although I think it should be called TrajectorySliderView...

mpharrigan:
+1 your name

PR #865, opened 2015-06-11
Added TrajectorySliderView

cxhernandez:
As discussed in https://github.com/mdtraj/mdtraj/issues/863

### DIFF
--- a/mdtraj/html/__init__.py
+++ b/mdtraj/html/__init__.py
@@ -1,6 +1,6 @@
 try:
     from .install import enable_notebook
-    from .trajectory_widget import TrajectoryView
+    from .trajectory_widget import TrajectoryView, TrajectorySliderView
     from .linked_heatmap import TrajectoryHeatmap
 except ImportError:
     pass
--- a/mdtraj/html/trajectory_widget.py
+++ b/mdtraj/html/trajectory_widget.py
@@ -6,7 +6,6 @@ import mdtraj as md
 
 from IPython.display import display, Javascript
 from IPython.html.widgets import DOMWidget, IntSlider, Box
-from IPython.display import display
 from IPython.html.widgets import interactive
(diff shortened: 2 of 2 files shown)


rmcgibbo:
are you supposed to call `display` inside the function?

cxhernandez:
@rmcgibbo: I made `TrajectorySliderView` a function (which returns a TrajectoryView class), but I'm not sure if you had imagined each one of these as a separate class.

cxhernandez:
It works in my notebook.... Is there something wrong with calling display?

rmcgibbo:
`display` essentially called automatically on the return value of each cell in the notebook.

cxhernandez:
> display essentially called automatically on the return value of each cell in the notebook

Not sure I follow. Are you saying it's unnecessary, since I'm returning the `TrajectoryView ` object?

rmcgibbo:
I don't entirely remember how this IPython API works. But basically, I think having the library call `display` is not the correct usage, even if it does work. I think we should be returning some kind of container.

cxhernandez:
Ah, I found an example [here](https://books.google.com/books?id=yZafBAAAQBAJ&pg=PT207&lpg=PT207&dq=ipython+widget+container&source=bl&ots=vVk32QMXP_&sig=z-eASvffRP-1UFF-RZi6S2W3cR4&hl=en&sa=X&ved=0CE0Q6AEwB2oVChMI4L3DxLyGxgIVUi6ICh3ssABQ#v=onepage&q=ipython%20widget%20container&f=false)

cxhernandez:
Though `ContainerWidget` is apparently deprecated as well...

mpharrigan:
Nice commit emoji

cxhernandez:
silly windows. Is this okay to merge?

rmcgibbo:
Don't be delayed on my behalf. Whoever else can hit the merge button if
it's ready.

Issue #864, opened 2015-06-11
Reading gro files with more than 2 characters before decimal point in velocity


mattwthompson:
I recently ran into an error in which mdtraj fails to load a *.gro file with 3 characters before the decimal point of a velocity element. Specifically, the issues is a velocity element like -10.0, which leaves no space between it and the previous velocity. I have highlighted the problem and made a dummy file with the problem line:

    1FOO    foo 9328   3.064   2.394   6.775-10.3772 -0.2281  2.8391

located at https://gist.github.com/mattwthompson/8a5ffaf15002c41df535. The file that this came from was produced by a gromacs simulation, like a confout.gro. It can also be read by gromacs to a grompp command.

~~Christoph and I have tracked it to line 344 in~~

~~https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/gro.py#L344~~

~~which fails because ```_isint(line[15:20])``` returns ```None``` in line 482~~

~~https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/gro.py#L482~~

~~We suggest changing it to ```_isint(line[15:20].strip())```~~

mpharrigan:
I'm confused: isn't `line[15:20]` 3.064?

rmcgibbo:
```
In [33]: line = '1FOO    foo 9328   3.064   2.394   6.775-10.3772 -0.2281  2.8391'

In [34]: line[15:20]
Out[34]: '8   3'
```

mpharrigan:
My counting by eye has failed me yet again. What field should that be picking up?

mattwthompson:
Hm, that doesn't appear to make any sense. Ignore the suggestion, it seems unlreated to the issue later in the line.

Not sure if the spaces are showing up in that line, but it should be 
>>> line = '    1FOO    foo 9328   3.064   2.394   6.775-10.3772 -0.2281  2.8391'
>>> line[15:20]
' 9328'

rmcgibbo:
I think the function `_is_gro_coord` is wrong. It shouldn't use `str.split`. It should use explicit offsets, since you can't depend on there being whitespace.

swails:
@leeping implemented a rigorous GRO parser in ParmEd that supports arbitrary precision GRO files.

See https://github.com/ParmEd/ParmEd/blob/master/parmed/gromacs/gromacsgro.py

[...]

PR #866, opened 2015-06-11
ENH: Fix GRO parser for files with higher precision.

swails:
Also adds the ability to write with arbitrary precision.

Should address #864


rmcgibbo:
Thanks, Jason. I'll try to come up with some good test cases later today
(I'm on the train now).

swails:
The failure is a timeout on one of the branches.

> Thanks, Jason. I'll try to come up with some good test cases later today
> (I'm on the train now).

One test I'd like to add is a round-trip where I write a high-precision GRO file and try to read it back in.  If I find time, I'll add it, but I have a Telluride conference coming up on Sunday and I have to get ready for that (and come up with something to present on).

rmcgibbo:
Yeah, don't worry about it, if you have other stuff to do. I can add the
test.

swails:
Awesome, thanks!

swails:
I took a break from my presentation and went ahead and added tests (which revealed a mistake I made in arbitrary precision writing support).  Should be fixed now, and the original v_error.gro file was added as a test.  This should be good to go.

Issue #870, opened 2015-06-13
[FR] Add Residue Code Attribute


cxhernandez:
It might be nice to be able to get residue 1 letter codes from the `Topology` object like:

``` python
>> res = topology.residue(1)
>> print(res.name)
'ALA'
>> print(res.code)
'A'
```

kyleabeauchamp:
If you do this, it might be cool to have a to_fasta decorator as well

cxhernandez:
:+1:

rmcgibbo:
Good idea. What is the desired `code` for `Residue`s that aren't protein?

To implement, probably all that's required is to add the mapping as a dict to https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/residue_names.py, and then one method on `Residue`, although we could also add something that uses this to the atom selection syntax.

cxhernandez:
Maybe just leave it at `None` if it's not protein or is non-canonical?

rmcgibbo:
+1

PR #871, opened 2015-06-14
One letter codes

cxhernandez:
As discussed in https://github.com/mdtraj/mdtraj/issues/870


cxhernandez:
I decided to include all the amino acids recognized by the PDB Chemical Component Dictionary, but this can easily be changed if you all decide that's overkill. 

I also added a `to_fasta` method to the `Topology` class.

cxhernandez:
Also, need to add a build test still

rmcgibbo:
Where exactly do these codes come from? I haven't looked at the PBD component dictionary before.

cxhernandez:
here: ftp://ftp.wwpdb.org/pub/pdb/data/monomers/components.cif

swails:
That's the CIF file.  Here's the page for it: http://www.wwpdb.org/data/ccd

swails:
There is a field `one_letter_code` that has this information.

rmcgibbo:
Alright. I think I'm the least knowledgeable person in the room here about most of these issues, I'm probably not the best person to do code review.

I'll let you guys decide what makes sense for this feature (full list of all amino acids in the dictionary, or some truncated set), do code review, and then merge.

cxhernandez:
Any thoughts @swails, @kyleabeauchamp, or anybody else?

mdtraj/core/topology.py
```diff
@@ -196,7 +199,9 @@ def __repr__(self):
         return "<%s at 0x%02x>" % (self._string_summary_basic(), id(self))
 
     def _string_summary_basic(self):
-        return "mdtraj.Topology with %d chains, %d residues, %d atoms, %d bonds" % (self.n_chains, self.n_residues, self.n_atoms, len(self._bonds))
+        return "mdtraj.Topology with %d chains, %d residues, \
+                %d atoms, %d bonds" % (self.n_chains, self.n_residues,
+                                       self.n_atoms, len(self._bonds))
```
swails: This should be changed.  All that gets cancelled is the `\n` -- the indentation spaces still get added to the returned string.  Change to this:

```Python
    def _string_summary_basic(self):
        return ("mdtraj.Topology with %d chains, %d residues, "
                "%d atoms, %d bonds" % (self.n_chains, self.n_residues,
                                        self.n_atoms, len(self._bonds)))

File: mdtraj/core/topology.py [hunk 2]
```diff
@@ -297,7 +327,8 @@ def to_openmm(self, traj=None):
             angles = traj.unitcell_angles[0]
 
             if np.linalg.norm(angles - 90.0) > 1E-4:
-                raise(ValueError("Unitcell angles must be 90.0 to use in OpenMM topology."))
+                raise(ValueError("Unitcell angles must be 90.0 to use \
+                                 in OpenMM topology."))
```
swails: Same here.  When breaking strings across lines, use the technique I described above.  This adds unnecessary spaces.  Also, this is no longer true (OpenMM supports triclinics), but I'll do this in a different PR.

mdtraj/core/residue_names.py
```diff
@@ -20,35 +20,849 @@
...
+                     'XX1': 'K',
+                     'YCM': 'C',
+                     'YOF': 'Y',
+                     'YTH': 'T',
+                     'Z01': 'A',
+                     'ZAL': 'A',
+                     'ZCL': 'F',
+                     'ZFB': 'X',
+                     'ZU0': 'T',
+                     'ZZJ': 'A'}
```
swails: Are all of these codes derived amino acids?  I don't think they are.  Single-letter codes should be separated from amino acid identification, IMO

swails: (Unless I'm wrong and these are all amino acids or modified versions)

swails:
I added my reviews.  I'm +1 on this.

mdtraj/core/topology.py
cxhernandez: Ah, good point.

mdtraj/core/residue_names.py
cxhernandez: In curating this `dict`, I grep'd for components that had a [`_chem_comp.type`](http://mmcif.wwpdb.org/dictionaries/mmcif_pdbx.dic/Items/_chem_comp.type.html) listed as 'peptide-like' or 'peptide-linking' in the `CIF`.

I checked a few of the weird ones, and they turned out to be modified amino acids. I'm assuming the bulk of these are modified residues but could definitely be wrong.

swails: I'm satisfied. That's exactly what I would have done.

mpharrigan:
Add to selection language and docs for selection lang?

cxhernandez:
How painful would that be? I've never touched that part of the code.

mpharrigan:
Should be a one line change addition? Let me know if you want help

cxhernandez:
Oh, okay cool! 

Also, any idea why this build keeps failing? I thought I started from master....

kyleabeauchamp:
No complaints on my end

kyleabeauchamp:
No idea about the travis failures, looks like a real test failure, but I'm not sure where

cxhernandez:
`ACE` and `NME` weren't included in the PDB CCD for some reason. I think that screwed up some of the atom selection tests.

cxhernandez:
okay just gotta add some docs + tests now

cxhernandez:
Do I need to update anymore of the docs?

mdtraj/core/residue_names.py
```diff
@@ -20,35 +20,851 @@
...
+    'SB', 'SCN', 'SE4', 'SEK', 'SM', 'SMO', 'SO3', 'SO4', 'SOD', 'SR',
+    'Sm', 'Sn', 'T1A', 'TB', 'TBA', 'TCN', 'TEA', 'THE', 'TL', 'TMA',
+    'TRA', 'UNX', 'V', 'V2+', 'VN3', 'VO4', 'W', 'WO5', 'Y1', 'YB', 'YB2',
+    'YH', 'YT3', 'ZN', 'ZN2', 'ZN3', 'ZNA', 'ZNO', 'ZO3'])
 
 _WATER_RESIDUES = frozenset([
     'H2O', 'HHO', 'OHH', 'HOH', 'OH2', 'SOL',
     'WAT', 'TIP', 'TIP2', 'TIP3', 'TIP4'])
+
+_AMINO_ACID_CODES = {'ACE': None,
```
mpharrigan: Can we condense this to fewer lines?

cxhernandez: fineeeee

PR #876, opened 2015-06-22
WIP: Store PDB chain identifier

rmcgibbo:
For #873. There is still an issue to work out with in the test though.

[Review on Reviewable](https://reviewable.io/reviews/mdtraj/mdtraj/876)

### DIFF
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -386,10 +386,12 @@ class Topology(object):
         pd = import_('pandas')
         data = [(atom.serial, atom.name, atom.element.symbol,
                  atom.residue.resSeq, atom.residue.name,
-                 atom.residue.chain.id) for atom in self.atoms]
+                 atom.residue.chain.index, atom.residue.chain.id)
+                for atom in self.atoms]
 
         atoms = pd.DataFrame(data, columns=["serial", "name", "element",
-                                            "resSeq", "resName", "chainID"])
+                                            "resSeq", "resName", "chainIndex",
+                                            "chainID"])
 
         bonds = np.array([(a.index, b.index) for (a, b) in self.bonds])
         return atoms, bonds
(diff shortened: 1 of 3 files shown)


mpharrigan:
Is this worth finishing?

evanfeinberg:
@mpharrigan I think it's worth finishing. Is it safe for me to merge this WIP to use in the mean time while the checks are fixed? Can I help get this working somehow? I think this feature would be particularly useful for simulations that entail protein protein interactions.

cxhernandez:
Fork [this branch](https://github.com/rmcgibbo/mdtraj/tree/wip-chain-ids) and have a crack at it!

``` bash
$ git clone git@github.com:rmcgibbo/mdtraj.git && cd mdtraj
$ git checkout wip-chain-ids
```

evanfeinberg:
Awesome, thanks @cxhernandez

evanfeinberg:
Update on this. Thanks again to @cxhernandez , @rmcgibbo , and @mpharrigan  for working on this. So this WIP PR works like a charm when one reads in a .dcd traj / .pdb top combination into an mdtraj trajectory object. The chains are correctly read in topology.chain.id. However, there is a bug, which I suspect is relatively easy to fix. When saving that mdtraj trajectory object in .h5 format, and then subsequently loading that .h5 file as a new trajectory object, the chain names get messed up. For some reason, they get renamed alphabetically from A to Z. So they topology.chain.id attribute is retained but its  value is reset. I assume this is an issue with either saving or loading or both. 

Can someone suggest how @amirbarati and I might fix this? Any hints would be appreciated.

rmcgibbo:
I suspect you need to modify this method: https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/hdf5.py#L305

evanfeinberg:
Thanks @rmcgibbo , we will have a crack at it this week. Shouldn't be too bad.

rmcgibbo:
@evanfeinberg any progress?

evanfeinberg:
Thanks for  the tips, I think it's fixed. n00b question: how do I commit/push to this particular PR?

rmcgibbo:
Actually, you can't. You want to submit a new PR, but it can incorporate the commits from this PR.

rmcgibbo:
If you're not sure how to do this I can also show you on gchat.

mpharrigan:
Any updates on this @evanfeinberg ?

evanfeinberg:
It's working now but failing in cases where one is loading a trajectory
that lacks chain information to start with. Once that fix is in it should
be done.

evanfeinberg:
Ok I think it's fixed. Will make PR ASAP

rmcgibbo:
This is outdated. Feel free to re-open if desired.

Issue #877, opened 2015-06-22
`force_overwrite` ignored for .xtc


OndrejMarsalek:
The `force_overwrite` flag seems to be silently ignored for .xtc, while for .pdb, it raises an `IOError`.  The .xtc file is overwritten, judging from the time stamp.

Example:

```
In [1]: import mdtraj as mdt
In [2]: trj = mdt.load('test.pdb')
In [3]: trj.save('test.xtc', force_overwrite=False)
In [4]: trj.save('test.xtc', force_overwrite=False)
In [5]: trj.save('test.pdb', force_overwrite=False)
---------------------------------------------------------------------------
IOError                                   Traceback (most recent call last)
[...]
IOError: "test.pdb" already exists
```

rmcgibbo:
Good spot. I added a test for this to check all of the formats for this behavior. Will fix soon.

rmcgibbo:
Fixed

PR #878, opened 2015-06-22
Force overwrite check 

rmcgibbo:
- [x] Add test for #877
 - [x] fix bug


mpharrigan:
Wanna finish this before we release 1.5?

PR #879, opened 2015-06-23
Fix #851

### LINKED ISSUE #851: Trajectory widget failing on virtual site lookup?
I assume this is related to the recent change for virtual sites in #835?
```python
mdtraj/mdtraj/html/trajectory_widget.py in _computeSecondaryStructure(self)
    138                 for a in top.residue(r).atoms:
    139                     result[a.index] = {
--> 140                         'ss': SS_MAP[ss],
    141                         'ssbegin': (r==rindxs[0] and ss in set(['H', 'E'])),
    142                         'ssend': (r==rindxs[-1] and ss in set(['H', 'E']))}

KeyError: 'NA'
```

Ignoring `'NA'` in the secondary structure computation fixes the issue for me but I'm not sure if this is the best way to solve this.

ctk3b:
I forgot to follow up on this but the following fixes the issue for my use cases. I'm not sure if this is the best way to solve it.

### DIFF
--- a/mdtraj/html/trajectory_widget.py
+++ b/mdtraj/html/trajectory_widget.py
@@ -124,10 +124,10 @@ class TrajectoryView(DOMWidget):
     secondaryStructure = Enum(['ribbon', 'strand', 'cylinder & plate',
                                'C alpha trace', 'nothing'], 'cylinder & plate',
                                sync=True)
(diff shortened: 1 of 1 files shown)


rmcgibbo:
Fixed by #896

Issue #883, opened 2015-07-03
principal component analysis


vvoelz:
Sorry to nitpick, but the spelling should be "principal" not "principle".  I noticed you have several scripts and documentation pages with the incorrect spelling

rmcgibbo:
good catch. thanks!

mpharrigan:
Wow, good to know... I always thought it was principle. Makes sense in hindsight

mpharrigan:
Fixed with referenced PRs

PR #884, opened 2015-07-03
Spelling: principal component analysis

mpharrigan:
cc #883


mpharrigan:
This is all I found with `grep`. Did I miss anything?

PR #886, opened 2015-07-08
changed _topology_from_subset to create topology with correct number of residues

hockyg:
When using trajectory.atom_slice, new trajectory file has the correct number of residues, but trajectory.top still had the original number of resides. This is because previously, there was an iteration over residues in the original topology, and a newResidue created for each one. 

In this modified version, only newResidues are created if there is an atom from that residue in the subset of atoms specified.

There is a point on the lines below where empty residues are deleted, but this still resulted in residues which internally contained an index which did not go from [0,new_n_residues]. 

That line may no longer be necessary with this change, but I did not remove it.
Additionally, I would expect a similar problem for chains, but it did not arise for my test PDB file, and so I did not make that additional modification.

### DIFF
--- a/mdtraj/core/selection.py
+++ b/mdtraj/core/selection.py
@@ -111,6 +111,7 @@ class SelectionKeyword(object):
         # (('nucleic', 'is_nucleic'), _chain('residue', 'is_nucleic')),
         (('water', 'waters', 'is_water'), _chain('residue', 'is_water')),
         (('name',), _chain('name')),
+        (('segment_id','segname',), _chain('segment_id')),
         (('index',), _chain('index')),
         (('n_bonds',), _chain('n_bonds')),
         (('residue', 'resSeq'), _chain('residue', 'resSeq')),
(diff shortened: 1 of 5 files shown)


rmcgibbo:
This PR seems to contain two separate features -- the segment_id thing is something different, no?

hockyg:
Hi Robert,

I the segment_id thing is a new commit. I was going to send a separate pull
request about that later, after I found out if you wanted to merge this
change. Whichever hash came with my original pull request shouldn't have
the new segment id changes.

It should be this one:

https://github.com/hockyg/mdtraj/commit/23f986d4b1cce939a96f418f464a9b1657ffc2ee

By the way, I did get to a point where the issue with the number of chains
that is similar to this issue with the number of residues arose. My patch
may not be the best if you'd want to change both at once (i.e. only
instantiate a new chain if it's going to have atoms in it).

rmcgibbo:
Fixed by #897.

PR #910, opened 2015-08-17
[core/topology] made all topology related classes hashable

marscher:
This was necessary since Python3 requires __eq__ and __hash__ impl.
Added test to assert equal hash values for "equal" objects.

### DIFF
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -234,14 +234,6 @@ class Topology(object):
     def __deepcopy__(self, *args):
         return self.copy()
 
-    def __hash__(self):
-        hash_value = hash(tuple(self._chains))
-        hash_value ^= hash(tuple(self._atoms))
-        hash_value ^= hash(tuple(self._bonds))
-        hash_value ^= hash(tuple(self._residues))
-
-        return hash_value
-
     def join(self, other):
         """Join two topologies together
 
@@ -505,6 +497,14 @@ class Topology(object):
         g.add_nodes_from(self.atoms)
         g.add_edges_from(self.bonds)
         return g
+
+    def __hash__(self):
+        hash_value = hash(tuple(self._chains))
+        hash_value ^= hash(tuple(self._residues))
+        hash_value ^= hash(tuple(self._atoms))
+        hash_value ^= hash(tuple(self._bonds))
+
+        return hash_value
 
     def __eq__(self, other):
         """Are two topologies equal?
@@ -1220,6 +1220,13 @@ class Chain(object):
         """Get the number of atoms in this Chain"""
         return sum(r.n_atoms for r in self._residues)
 
+    def __eq__(self, other):
+        return (self.index == other.index and
+                self._residues == other.residues)
+
+    def __hash__(self, *args, **kwargs):
+        hash_value = hash(tuple(self._residues)) ^ hash(self.index)
+        return hash_value
 
 class Residue(object):
     """A Residue object represents a residue within a Topology.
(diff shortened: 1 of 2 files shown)


marscher:
@rmcgibbo I need help regarding the error in compute_contacts. It seems there is no comparison with \__eq__ methods, so this seems unrelated to my changes.

rmcgibbo:
The error isn't related to compute contacts per se, it's just that the topology object is broken somehow and it thst manifests in many tests.

rmcgibbo:
Various places make checks like `chain != None`. The problem, I think, is that the `__eq__` methods need to check the type.

e.g. https://github.com/marscher/mdtraj/blob/fix_topology_unhashable_py3/mdtraj/core/topology.py#L517-L520

marscher:
Am 17.08.2015 um 18:36 schrieb Robert T. McGibbon:
> Various places make checks like |chain != None|. The problem, I think,
> is that the |__eq__| methods need to check the type.
> 
> e.g.
> https://github.com/marscher/mdtraj/blob/fix_topology_unhashable_py3/mdtraj/core/topology.py#L517-L520
> 
> —
> <https://github.com/mdtraj/mdtraj/pull/910#issuecomment-131883641>.
> 
Thank you for the hint! I will check it out.

swails:
> Various places make checks like chain != None. The problem, I think, is that the __eq__ methods need to check the type.

This is what `grep` tells me:

```
[swails@Robin ~/src/mdtraj/mdtraj (master) ]$ grep "== None" `find . -name "*.py"`
./formats/pdb/pdbstructure.py:        if self.default_model == None:
./formats/pdb/pdbstructure.py:        if self._current_model == None:
./formats/pdb/pdbstructure.py:        if alt_loc == None:
./formats/pdb/pdbstructure.py:        if alt_loc == None:
./formats/pdb/pdbstructure.py:        if alt_loc == None:
./formats/pdb/pdbstructure.py:        if alt_loc == None:
./formats/pdb/pdbstructure.py:                if locs == None: # means all locations
./formats/pdb/pdbstructure.py:        if (id == None):
./formats/pdb/pdbstructure.py:        if (id == None):
./formats/pdb/pdbstructure.py:        if serial_number == None:
./formats/pdb/pdbstructure.py:        if alternate_location_indicator == None:
./formats/pdb/pdbstructure.py:        if alt_loc == None:
./utils/unit/doctests.py:        >>> meter == None
./utils/unit/mymatrix.py:    if n == None:
./utils/unit/quantity.py:        if unit == None: # one argument version, copied from UList
./utils/unit/quantity.py:        if value == None:
./utils/validation.py:        Is ``val == None`` acceptable?
[swails@Robin ~/src/mdtraj/mdtraj (master) ]$ grep -e "!= None" `find . -name "*.py"`
./formats/pdb/pdbstructure.py:        if (self.formal_charge != None): formal_charge = "%+2d" % self.formal_charge
./utils/unit/doctests.py:        >>> meter != None
./utils/unit/quantity.py:            if guide_unit != None:
```

These should probably be changed into `is` and `is not`, which are the preferred ways of comparing to `None` (and bypasses any calls to `__eq__`) except in rare cases...  @rmcgibbo -- do you want me to file a PR for this?

rmcgibbo:
There are other places that still need this check, like a loop in the pdb writer that does `chain != prevChain` while iterating over the chain. For the first iteration, `prevChain` is initialized to `None`.

rmcgibbo:
I don't really think a PR for those things is necessary.

rmcgibbo:
it looks like there's a new issue on 32 bit win py27, which used to be passing.

marscher:
Yes, indeed. But the output is not informative to me. It seems like the 
test phase of conda-build already passed and something is going to be 
installed, which makes no sense.

rmcgibbo:
yeah, it's not clear to me either.

Issue #912, opened 2015-08-17
Appveyor python 3 failures


rmcgibbo:
Debugging these appveyor python 3 failures, https://ci.appveyor.com/project/rmcgibbo/mdtraj/build/1.0.524, they're coming from a segfault in pytables. grr.

mpharrigan:
Fixed by #913

PR #913, opened 2015-08-17
Fix windows python3 appveyor error

rmcgibbo:
Seems to be a segfault anytimes pytables interacts with a BLOSC-compressed file on windows py3, so I regenerated two very old reference files, and skipped the lh5 tests on win py3. cc #912


mpharrigan:
Is this still relevant? Probably could use a rebase or two

rmcgibbo:
Okay I think this is fixed, by rooting out all of the BLOSC-compressed files from the tests basically.

rmcgibbo:
@mpharrigan can you apply https://github.com/msmbuilder/msmbuilder/pull/650 to this repo too?

mpharrigan:
Sure. I'll do it tomorrow

rmcgibbo:
IIRC, there's one issue here that I don't understand where on certain versions of Python, the DSSP tests hang or segfault or something, and I think it's somewhat random.

rmcgibbo:
But otherwise, this should probably be merged.

mpharrigan:
+1

PR #915, opened 2015-08-18
Superpose+rmsd with different reference, query indices

mpharrigan:


rmcgibbo:
Seems like a good idea.

rmcgibbo:
+1

cxhernandez:
+1000

mpharrigan:
Shall we merge this? cc #923

Issue #916, opened 2015-08-20
[FR] Principal axes of rotation


mpharrigan:
Analysis functions to (1) get moment of inertia tensor and (2) sorted principal axes of rotation

This would be a nice thing to have.

cxhernandez:
+1

ctk3b:
MOI tensor calculation is here: https://github.com/mdtraj/mdtraj/blob/master/mdtraj/geometry/order.py#L135

mpharrigan:
Haha whoops. Sorry @ctk3b 

New request! Add that function to the docs, and maybe a convenience function for sorted principal axes

ctk3b:
+1 :)

PR #917, opened 2015-08-20
Update analysis.rst

ctk3b:
Not quite sure which category this would otherwise go under but I figured this would be the best fit.

See #916


rmcgibbo:
lgtm

PR #930, opened 2015-09-22
Fix MIC code for general triclinic unit cells.

swails:
Not fully correct yet.  I'll post some files to compare against what cpptraj computes.  It disagrees on ~20% of the frames in my test case compared to what cpptraj computes (but it's a lot better than the 95% it used to disagree with).

This may not be the best approach, but it retains backwards compatibility (i.e., `unitcell_vectors` are still row vectors instead of the column vectors specified by Tuckerman's book) while getting closer to the correct MIC distance calculation.


rmcgibbo:
This looks good. I think tests are critical though.

swails:
> This looks good. I think tests are critical though.

Definitely.  Did you look at the link I provided?  I'm not even convinced that this code is working yet -- once I figure out why the test case I put on Dropbox isn't working as well as I think it should be, I will turn that into a test case here.

swails:
OK, so this is not going to work for the general case.  As you can see in my example, there are instances where this approach *does* give the minimum distance (which happens much more often than not -- about 75% of the time), and there are cases where it does not (the remaining 25% of the frames).  In particular, it seems that this approach is appropriate for MD, where cutoffs *must* be small enough so that they do not intersect the same plane of multiple periodic images.  However, there are corners of the unit cell where *no* such cutoff can ever reach, and when two particles are separated within the same unit cell by more than the maximum allowable cutoff, this algorithm will not always give the minimum distance.  This is discussed to some extent here: http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.57.1696&rep=rep1&type=pdf

For cubic cells, this approach works, but for triclinic cells we need either what GROMACS does or what cpptraj does (which are probably the same thing).

swails:
The non-optimized (pure Python) distance routines now work correctly in all cases that I can tell.  What I had to do was do a brute force search over all 27 adjacent unit cells to see if the distance decreased at all.  This is quite a bit slower than *not* having to do this, obviously, but is rather critical to getting the correct distances.

The next step is to add this into the optimized version.  I am not, however, at all experienced with using SSE intrinsics, so I won't be using them for what I have to add.  After all this is done and working, we can think about optimization (like special-casing orthorhombic cells which don't need the 27-nearest-box-search).

swails:
I've already spent more time than I have on the C code, so I'm giving up.  I don't know the SSE intrinsics, and it's going to be too tedious to do this the right way.  The problem right now is in `minimum_image`, since that is the function used in all `*_mic` functions (rather than re-using the displacement MIC code).  While I would've been able to tack on the extra loop over all unit cell neighbors in non-SSE code at the end of `_dist_mic`, this would not fix the currently incorrect behavior of `_angle_mic` and `_dihedral_mic`.

So someone else will have to handle fixing the C code.  I'll add tests to this PR by early next week (hopefully later today, but I have to focus on other things for now).

swails:
OK, so I thought of an approach that will at least be a stopgap before someone implements an actual vectorized version, but it is at least *partly* vectorized for nonorthorhombic cells (it keeps the original vectorized parts, and then loops over all 27 nearest unit cells without vectorization).

As for the angle and dihedral calculations that were broken because of this, the optimized versions of those were disabled with a warning about the fallback to the slower implementation.  This should at least get everything *correct*, with distances being at least comparable in performance to other packages.

rmcgibbo:
@swails thanks for dealing with this. I can work on the speed later, when I have more time, but it would be good to get this fix in as soon as possible, regardless of the performance.

swails:
Where the speed is needed the most right now is in the angle and dihedral routines that currently need to fall back to the Python implementation.  Those aren't of interest to me (I needed the distance stuff working, since I plan on adding wrapping/imaging capabilities shortly).

I'm finishing up this PR with a few more tests (the ones I used to implement this).

swails:
The Travis failure has to do with the fact that `np.linalg.norm` didn't get an `axis` keyword until 1.8.  I can work around this for now (but the axis keyword is a nice addition...).  For future reference, what's the target numpy version we aim to support, and what's the policy of bumping that requirement up?

swails:
OK, the last failure was due to an incompatibility with numpy 1.6, but that should finally be fixed.  As that was the final error, this last commit should pass the Travis tests.

Should be good to merge.

swails:
Fixes #908

swails:
OK -- should be good to go.

swails:
This is ready to merge on my end.  If someone could review the code here, we can close out #908.

At some point we should figure out the appveyor failures, but that was failing before this PR.

swails:
Oops, one small fix.

swails:
Sorry -- after Travis passes, it should be good to go.

PR #934, opened 2015-09-30
Automized setup (pip) to drag-in build time dependencies (numpy, cython)

### LINKED ISSUE #933: Automatic dependency resolvation with pip
I wonder if you would accept a PR which drags in dependencies like NumPy and Cython during installation with pip. If one installs a mdtraj dependent project in a clean environment the order of installation might get resolved wrong, if these are not declared in a setuptools specific manner.

This would not interfere with conda, since automatic downloading has been blocked in recent versions of conda-build anyway.

If one opts in cython during setup time, one should considerate to request the users to have at least setuptools >= 18, because since this version the hassle of cython builds has been fixed.

What do you think?

marscher:
* NumPy includes and Cython extensions are handled transparently.
* Requires setuptools >= 18

closes #933

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -131,8 +131,7 @@ def rmsd_extensions():
                          'mdtraj/rmsd/src/rotation.c',
                          'mdtraj/rmsd/src/center.c',
                          'mdtraj/rmsd/_rmsd.pyx'],
-                     include_dirs=['mdtraj/rmsd/include',
-                                  ],
+                     include_dirs=['mdtraj/rmsd/include'],
                      extra_compile_args=compiler_args,
                      libraries=compiler_libraries)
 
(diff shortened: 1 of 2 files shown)


rmcgibbo:
Unfortunately, inclusion the numpy include directories (including `np.get_include()` in the `Extension` configuration) is not optional -- note the failures on Travis. The cython extensions use typedefs from `npy_common.h` and such, and need to be compiled against numpy.

marscher:
Added numpy inc dir, but now Appveyor Py3 fails. I can not see an useful error message.

swails:
> Added numpy inc dir, but now Appveyor Py3 fails. I can not see an useful error message.

I believe it's *been* failing.  This is current master: https://ci.appveyor.com/project/rmcgibbo/mdtraj/branch/master

rmcgibbo:
What is the thing that was fixed / added in setuptools 18 that makes this work?

marscher:
The transparent Cython handling is new in this version. I guess they're
searching the sources list for pyx files and cythonize them on the fly.
This way setuptools is able to install cython during setup time
(setup_requires dependency).

https://bitbucket.org/pypa/setuptools/issue/288

setup.py
```diff
@@ -172,13 +147,13 @@ def rmsd_extensions():
                            'mdtraj/rmsd/_lprmsd.pyx'],
                        language='c++',
                        include_dirs=[
-                           'mdtraj/rmsd/include', numpy.get_include()],
+                           'mdtraj/rmsd/include',
+                                    ],
```
mpharrigan: this formatting looks weird

rmcgibbo:
Sweet.

PR #939, opened 2015-10-09
Follow the Amber NetCDF Trajectory Spec (for compatibility with cpptraj)

mpharrigan:
`cpptraj` was refusing to read mdtraj-generated `.nc` files because we weren't following the spec to a T (or an "xyz")

This fixes it and adds a test. The test requires cpptraj. Should we skip on travis and how?

### DIFF
--- a/mdtraj/formats/netcdf.py
+++ b/mdtraj/formats/netcdf.py
@@ -450,17 +450,17 @@ class NetCDFTrajectoryFile(object):
             self._handle.createDimension('label', 5)
 
             # Define variables to store unit cell data
-            self._handle.createVariable('cell_spatial', 'c', ('cell_spatial',))
-            cell_angles = self._handle.createVariable('cell_angular', 'c', ('cell_spatial', 'label'))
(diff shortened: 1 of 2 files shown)


rmcgibbo:
you can skip on travis by adding a `@skipif` decorator, and testing on the travis environment variable. i'll figure out the best way to get cpptraj on travis.

rmcgibbo:
See e.g. the tests against DSSP https://github.com/mdtraj/mdtraj/blob/master/mdtraj/geometry/tests/test_dssp.py#L55

mpharrigan:
Skipped

rmcgibbo:
something bad is happening on travis related to numpy and compilation.

rmcgibbo:
@mpharrigan might be possible to add cpptraj to the travis config using the amber conda channel: https://github.com/omnia-md/conda-recipes/issues/340#issuecomment-146746320

mpharrigan:
Conda is installing numpy 1.10 regardless of env variable (?)

Could this be conflicting somehow

rmcgibbo:
The  issue is from https://github.com/conda/conda-build/pull/573, and it's because the version of numpy used during compilation isn't being included in the build string anymore.

PR #941, opened 2015-10-09
Update conda packaging

mpharrigan:
Also get rid of py2.6 stuff, as we haven't been building 2.6 conda packages for a while

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -11,15 +11,13 @@ requirements:
   build:
     - python
     - cython
-    - numpy
+    - numpy x.x
     - setuptools
 
   run:
     - python
     - setuptools
-    - argparse # [py26]
-    - ordereddict # [py26]
-    - numpy
+    - numpy x.x
     - pandas
     - scipy
     - pytables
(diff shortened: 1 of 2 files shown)


mpharrigan:
The current working hypothesis is that this will fix travis tests*, but we will still run into problems when trying to upload packages for different versions of numpy because numpy version is not reflected in the build string

*travis tests appear to be broken because it downloads numpy 1.10 during build (ignoring `CONDA_NPY` env variable) but then tests against 1.8 or 1.9 (respecting the env variable)

mpharrigan:
The opposite is happening now. It's being built with 1.9 and tested against 1.10 (which seems to work because abi is backwards compatible (?))

mpharrigan:
Adding `x.x` to both build and run seems to fix everything (and it puts it in the build string)

mpharrigan:
We can't test py35 until a package for (at least) scripttest is uploaded for py35. Probably shiftx2 as well. Everything else is not-omnia

mpharrigan:
Ok, travis is back to passing.
I cancelled the other travis (which is triggered by pushes to the branch which is in this repo not my fork)
appveyor is stuck in a big queue of commits that I can't cancel

devtools/conda-recipe/meta.yaml
```diff
@@ -11,15 +11,13 @@ requirements:
   build:
     - python
     - cython
-    - numpy
+    - numpy x.x
```
marscher: May I ask what this pattern does exactly?

marscher: Ah I see. Just ignore my jitter :)

marscher:
Is support for NumPy 1.10 on the agenda for the whole omnia stuff? Do we need this "numpy x.x" pattern then in all numpy dependent packages?

mpharrigan:
@marscher : @rmcgibbo updated the conda recipies in omnia where "numpy x.x" was required (i.e. wherever you link against the numpy c api)

numpy 1.10 is on the agenda https://github.com/omnia-md/conda-recipes/issues/342

PR #952, opened 2015-10-20
Use improved numpy histogram in RDF.

ctk3b:
This changes how the numpy histogram is invoked so that it uses the new optimizations from https://github.com/numpy/numpy/pull/6100 and as discussed in #734.


rmcgibbo:
This doesn't actually depend on having the new version of numpy, right? It's just that if you do have the new version, this will cause a faster codepath to be used?

ctk3b:
Yup, numpy>=1.10 users get  a pretty significant speed boost.

rmcgibbo:
Awesome.

Issue #957, opened 2015-10-23
[FR] Add trajectory smoothing


cxhernandez:
Helpful for making pretty movies

cxhernandez:
@rmcgibbo: I tried using the Butterworth Filter you implemented in [nbeterpolator](https://github.com/rmcgibbo/nebterpolator/blob/master/nebterpolator/smoothing.py#L112) and am liking the results compared to using a plain moving average. What do you think?

Issue #959, opened 2015-10-27
Release 1.5


mpharrigan:
Last release was June. There've been some major fixes and enhancement, my favourites include*

 - writing netcdf trajectories "correctly"
 - that pbc distances bug
 - rmsd with different reference / query indices

(* I don't have that fancy script that finds all the PRs merged since a release)

swails:
+1

And I think you can go to the release page on github to see the list of commits since the last release.

mpharrigan:
Yeah, but it includes all the commits which have helpful titles such as https://github.com/mdtraj/mdtraj/commit/4ca59d11464a0b1ac3210f523304ecc0b3c78aed :)

rmcgibbo:
```
$ gh-util list-prs mdtraj/mdtraj
Listing merged PRs since: MDTraj 1.4.2

PR #955: Force clang on darwin
PR #954: [setup] require at least setuptools>=18 before doing anything.
PR #952: Use improved numpy histogram in RDF.
PR #951: Update BibTeX reference
PR #950: Pin version of msmb_theme
PR #945: Addresses mol2 reading bug #944
PR #943: [setup] do not call parent function via super, because Extension is an old-style class.
PR #942: Fix setup.py develop with latest version of setuptools
PR #935: Get the IPython notebook viewer working with IPython/Jupyter 4
PR #941: Update conda packaging
PR #939: Follow the Amber NetCDF Trajectory Spec (for compatibility with cpptraj)
PR #940: Update test_dssp.py
PR #938: Update dcd.pyx
PR #934: Automized setup (pip) to drag-in build time dependencies (numpy, cython)
PR #930: Fix MIC code for general triclinic unit cells.
PR #926: Fix dev upload
PR #925: Fix openmm-dev package uploads to binstar
PR #915: Superpose+rmsd with different reference, query indices
PR #922: Update trajectory_files.rst
PR #921: Update load_functions.rst
PR #917: Update analysis.rst
PR #911: Update README.md
PR #905: bugfix/dcdreader: fix sharing mode under Windows
PR #902: Update installation.rst
PR #901: Add  load_hoomdxml to top level __init__.py to supress sphinx docs warning.
PR #900: Hoomd fix
PR #899: MAINT: Use travis-ci container infastructure
PR #898: Add hash implementation
PR #897: Fix n_residues set by _topology_from_subset
PR #896: Fix #851, part 2
PR #885: Fix bug and sorting in hoomdxml reader
[...]

PR #962, opened 2015-10-28
smoothing

cxhernandez:
ref https://github.com/mdtraj/mdtraj/issues/957


mdtraj/core/trajectory.py
```diff
@@ -1669,6 +1669,74 @@ def remove_solvent(self, exclude=None, inplace=False):
...
+        inplace : bool, default=False
+            The return value is either ``self``, or the new trajectory,
+            depending on the value of ``inplace``.
+        Returns
+        -------
+        traj : md.Trajectory
+            The return value is either ``self``, or the new smoothed trajectory,
+            depending on the value of ``inplace``.
+        """
+        from scipy.signal import lfilter, lfilter_zi, filtfilt, butter
```
cxhernandez: Wasn't sure if I should import these at the top...

mpharrigan:
smooth like butter...worth

mdtraj/core/trajectory.py
```diff
@@ -1669,6 +1669,74 @@ def remove_solvent(self, exclude=None, inplace=False):
 
         return self.atom_slice(atom_indices, inplace = inplace)
 
+    def smooth(self, width=11, order=3, atom_indices=None, inplace=False):
+        """
+        Smoothen a trajectory using a zero-delay Buttersworth filter.
+        This code is copied from the scipy cookbook, with sytlistic improvements.
+        http://www.scipy.org/Cookbook/FiltFilt
+        Parameters
```
mpharrigan: linebreak before parameters

mdtraj/core/trajectory.py
```diff
@@ -1669,6 +1669,74 @@ def remove_solvent(self, exclude=None, inplace=False):
...
+        order : int, optional
+            The order of the filter. A small odd number is recommended. Higher
+            order filters cutoff more quickly, but have worse numerical
+            properties.
+        atom_indices : array-like, dtype=int, shape=(n_atoms)
+            List of indices of atoms to retain in the new trajectory.
+        inplace : bool, default=False
+            The return value is either ``self``, or the new trajectory,
+            depending on the value of ``inplace``.
+        Returns
```
mpharrigan: likewise

cxhernandez:
@mpharrigan you should give this a try on some of your trajectories. I've only tried it on WW, but the results looked pretty good.

mpharrigan:
Should we include this in v1.5 cc #959 ?

cxhernandez:
I think it'd be nice to have in v1.5

cxhernandez:
Any idea what the appveyor errors are for py3?

mpharrigan:
well the build matrices are different, which leads me to believe that the "push" ci uses older code from the rest of the codebase whereas the "pr" ci does a merge with HEAD first

mpharrigan:
You should probably add a test for this function

mdtraj/core/trajectory.py
```diff
@@ -1669,6 +1669,76 @@ def remove_solvent(self, exclude=None, inplace=False):
...
 
+    def smooth(self, width=11, order=3, atom_indices=None, inplace=False):
+        """
+        Smoothen a trajectory using a zero-delay Buttersworth filter.
+        This code is copied from the scipy cookbook, with sytlistic improvements.
+        http://www.scipy.org/Cookbook/FiltFilt
+        
+        Parameters
+        ----------
+        width : float
```
mpharrigan: you have a check below that makes sure this is int

mpharrigan: I don't think `width` should have a default

mdtraj/core/trajectory.py
```diff
@@ -1669,6 +1669,76 @@ def remove_solvent(self, exclude=None, inplace=False):
...
+            This acts very similar to the window_len in the window smoother. In
+            the implementation, the frequency of the low-pass filter is taken to
+            be two over this width, so it's like "half the period" of the sinusiod
+            where the filter starts to kick in.
+        order : int, optional
+            The order of the filter. A small odd number is recommended. Higher
+            order filters cutoff more quickly, but have worse numerical
+            properties.
+        atom_indices : array-like, dtype=int, shape=(n_atoms)
+            List of indices of atoms to retain in the new trajectory.
```
mpharrigan: Add something about the default is None and that keeps all the atoms

mdtraj/core/trajectory.py
```diff
@@ -1669,6 +1669,76 @@ def remove_solvent(self, exclude=None, inplace=False):
...
+
+        xyz = self.xyz.copy()
+
+        for i in atom_indices:
+            for j in range(3):
+
+                signal = xyz[:, i, j]
+
+                # find nearest odd integer
+                pad = int(np.ceil((width + 1)/2)*2 - 1)
```
mpharrigan: Move outside the loops?

mpharrigan: Actually almost all of this should be moved outside the loop, no?

cxhernandez: good catch!

rmcgibbo: do they scipy filter functions allow broadcasting, or is it really necessary to loop over the atom indices and dimensions.

mdtraj/core/trajectory.py
cxhernandez: Yeah that was left over from @rmcgibbo's code. He has this line that converts it into an odd integer:

`pad = int(np.ceil((width + 1)/2)*2 - 1)`

Philosophically, I think the user should know that `width` has to be an odd whole number so they have at least some idea of what's going on under the hood, but maybe that's asking for too much?

mpharrigan: Whatever you think is best re: int or float, just make sure the docstring is correct

I don't think it should have a default value though

cxhernandez:
What should the test look like? Making sure it spits out a trajectory of the same length?

hainm:
> What should the test look like? Making sure it spits out a trajectory of the same length?

What are your expected traj? assert to your expected one. You can visualize it, if you think the smoothing seems ok, I think you can assert to that (or to output from other program).

mpharrigan:
This needs a good alignment within the trajectory right?

mdtraj/core/trajectory.py
```diff
@@ -1669,6 +1669,80 @@ def remove_solvent(self, exclude=None, inplace=False):
 
         return self.atom_slice(atom_indices, inplace = inplace)
 
+    def smooth(self, width, order=3, atom_indices=None, inplace=False):
+        """
```
rmcgibbo: No need for the extra linebreak here, and put the reference to the scipy cookbook as a citation instead (e.g. https://github.com/mdtraj/mdtraj/blob/master/mdtraj/geometry/dssp.py#L75-L78).

rmcgibbo:
Yeah, I think it would might make sense to either do a progressive alignment of the frames before smoothing, or at least to mention something about alignment in the docs.

cxhernandez:
Yep, alignment is necessary.

cxhernandez:
Alright. I added added frame0-smooth.dcd which is used as a reference for the `test_smooth`.

cxhernandez:
One of the appveyor tests failed on the trajectory comparison :confused:

cxhernandez:
Any ideas why the trajectories are different for Win/Lin?

cxhernandez:
Any ideas why the trajectories are different for Win/Lin?

rmcgibbo:
Strange.

rmcgibbo:
Can you try adding a simpler test case that applies the same filter function to a known 1D signal, like `np.sin(np.arange(100))` or something?

rmcgibbo:
e.g.

```
pad = 5
signal = np.sin(np.arange(100))
padded = np.r_[signal[pad - 1: 0: -1], signal, signal[-1: -pad: -1]]
z, _ = lfilter(b, a, padded, zi=zi*padded[0])
[... etc]

assert result == [hardcoded result that you compute on your computer and just copy-paste in]
```
And then we can see if it matches in windows.

rmcgibbo:
I can also hop on a Windows VM, or boot into Windows in lab.

cxhernandez:
I just tried it on a VM, and the results are not the same :disappointed:

rmcgibbo:
What exactly isn't consistent? Can you get it down to a single method?

cxhernandez:
Actually, even generating the signal using `np.sin(np.arange(100))` results in small differences.

rmcgibbo:
Are you testing for actual equality? https://possiblywrong.wordpress.com/2013/11/15/floating-point-equality-its-worse-than-you-think/

cxhernandez:
Okay– after using `mdtraj.testing.eq`, everything seems to check out.

rmcgibbo:
Sweet!

rmcgibbo:
FWIW, I would recommend reading http://www.johndcook.com/blog/2009/04/06/numbers-are-a-leaky-abstraction/

rmcgibbo:
Is this still a work in progress? It looks ready to merge to me.

rmcgibbo:
p.s. Thanks for implementing this!

cxhernandez:
np!

Issue #963, opened 2015-10-28
NameError: name 'os' is not defined


cxhernandez:
I think https://github.com/mdtraj/mdtraj/commit/edecc302f33e477821de01f66da1acf3969505d7 is missing `import os`.

```
[cxh@local ~/Developer/mdtraj]$ python setup.py install
C compiler:
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/usr/include/c++/4.2.1
Apple LLVM version 7.0.0 (clang-700.0.72)
Target: x86_64-apple-darwin15.0.0
Thread model: posix
Attempting to autodetect OpenMP support... Did not detect OpenMP support
Attempting to autodetect SSE3   support... Compiler supports SSE3
Attempting to autodetect SSE4.1 support... Compiler supports SSE4.1

Traceback (most recent call last):
  File "setup.py", line 61, in <module>
    os.environ['CXX'] = 'clang++'
NameError: name 'os' is not defined
```

mpharrigan:
We need to convince @rmcgibbo to use an ide

![code1](https://cloud.githubusercontent.com/assets/4967059/10794903/6b56be30-7d54-11e5-869d-9ad86da60639.png)

cxhernandez:
:+1:

PR #964, opened 2015-10-28
import os

cxhernandez:
ref https://github.com/mdtraj/mdtraj/issues/963


mpharrigan:
https://www.python.org/dev/peps/pep-0008/#imports

also looks like you can remove `pkg_resources` as an import. Yay pycharm

cxhernandez:
pep8 nazi!

PR #966, opened 2015-10-30
set comprehension

cxhernandez:


mpharrigan:
Note that this doesn't work in py2.6

rmcgibbo:
Good riddance.

mpharrigan:
:)

hainm:
not sure if it's worth supporting py2.6.

rmcgibbo:
http://www.curiousefficiency.org/posts/2015/04/stop-supporting-python26.html

PR #973, opened 2015-11-04
Miniconda changed default install path

mpharrigan:
Might as well switch to python 3 while we're at it


cxhernandez:
Is this why travis has been acting up?

rmcgibbo:
+1

Issue #970, opened 2015-11-04
All PDBs from RCSB contain CRYST1 record, but some of them are nonsensical fillers


swails:
The question is, what heuristic test can we come up with that will serve as a useful test for when we can issue warnings?

I have 2 ideas:

1. Calculate a simple density using the volume of the unit cell and the total mass of the system.  Emit a warning if it is well outside the range of acceptable values.  For instance, I think the unit cell dimensions for NMR structures are tiny (1x1x1 A).  This is from 2KOC:

```
CRYST1    1.000    1.000    1.000  90.00  90.00  90.00 P 1           1
```

2. Pull out the `EXPDTA` record and fail if it exists by is *not* X-Ray diffraction.

I personally like the first option better, as it is a more useful warning IMO.  I can work on a PR if people agree.

cc pandegroup/openmm#1176

mpharrigan:
Fixed by #974

PR #974, opened 2015-11-04
Add a heuristic check for valid unit cells on PDB file reads.

swails:
Discard unit cell vectors on PDB files if resulting particle density is absurdly
large. RCSB PDB files *all* have CRYST1 records, and dummy values are inserted
(typically 1x1x1 A) whenever the experiment doesn't actually define a unit cell
(like for NMR experiments).


swails:
My idea for fixing #970

rmcgibbo:
This seems like a pretty good idea. @rafwiewiora: thoughts?. @swails: can you move the test into `mdtraj/formats/pdb/tests/`

swails:
Sure. I think the heuristic needs to be double-checked. It's causing some unexpected test failures so I think I may need to recheck typical particle densities in condensed systems. 

I didn't want to keep going on it until we decided it was the thing to do. If people are generally favorable I'll clean it up and fix the failures.

rafwiewiora:
Looks good, thanks @swails ! I'm happy to run this on a PDB-wide search and see if any problems.

rmcgibbo:
Maybe add a kwarg to turn this behavior off though. I can see for certain situations (like the test suite, maybe others) where you really just want to load the information in the file as faithfully as posible, and apply any needed heuristics downstream or afterwards.

jchodera:
> Maybe add a kwarg to turn this behavior off though. 

+1

swails:
OK.  I finally got back to this and fixed it up.  I merged master and rebased to clean up the merge a bit.  I double-checked the heuristic, and 100 atoms/nm^3 is the correct value given standard water concentration of 55.5M.  The tests that had failed were because the box in some of the PDB file were made up and *were* actually absurd.

I added a keyword `no_boxchk` to turn off the box check (should I change it to `skip_boxchk`? better name suggestions?).  But I also had to add some code to pass kwargs through some of the dispatchers, like `load_topology` and some of the other private helpers in that module.  This was already done for `md.load()`, but not the others.  So I think this improves overall consistency there.

I'm satisfied with the implementation here for merging, but I'd appreciate an extra set of eyes glancing over it.  I'd like to solicit bikeshed opinions/advice on some issues, but I'll do that in-line.

mdtraj/core/trajectory.py
```diff
@@ -162,20 +162,20 @@ def _parse_topology(top):
...
-        topology = load_gro(top).topology
+        topology = load_gro(top, **kwargs).topology
     elif isinstance(top, string_types) and (ext in ['.arc']):
-        topology = load_arc(top).topology
+        topology = load_arc(top, **kwargs).topology
     elif isinstance(top, string_types) and (ext in ['.hoomdxml']):
-        topology = load_hoomdxml(top).topology
+        topology = load_hoomdxml(top, **kwargs).topology
     elif isinstance(top, Trajectory):
         topology = top.topology
```
swails: In this branch, `kwargs` is ignored.  In every other routine, if `kwargs` contains an unrecognized/unsupported argument, an exception is thrown.  Should I add something here like

```Python
if kwargs:
    raise TypeError('Unrecognized arguments %s' % ', '.join(kwargs.keys()))
```

for consistency?

swails:
Finally passes.  Pending discussion this is ready to merge.

swails:
What do people think?

jchodera:
This scheme sounds fine:
```
+        # Only one CRYST1 record is allowed, so only do this check for the first
+        # frame. Some RCSB PDB files do not *really* have a unit cell, but still
+        # have a CRYST1 record with a dummy definition. These boxes are usually
+        # tiny (e.g., 1 A^3), so check that the particle density in the unit
+        # cell is not absurdly high. Standard water density is ~55 M, which
+        # yields a particle density ~100 atoms per cubic nm. It should be safe
+        # to say that no particle density should exceed 10x that.
```

Issue #982, opened 2015-11-06
Slim package size


mpharrigan:
The testing/reference folder is 60 MB

```
[snip]
100K    ncinpcrd.rst7
104K    4OH9.pdb
112K    ala_dipeptide_trj
132K    frame0.binpos
152K    inpcrd
156K    frame0.binpos.h5
156K    frame0.nc
172K    1ncw.pdb.gz
172K    frame0.dcd
180K    alanine-dipeptide-explicit.pdb
192K    ala_dipeptide.pdb
196K    1am7_protein.pdb
212K    4K6Q.pdb
240K    GG-tip4pew.pdb
268K    mdcrd.nc
272K    g_sas_ref.dat
276K    frame0.lammpstrj
280K    monolayer.pdb
352K    alanine-dipeptide-explicit.prmtop
356K    freesolve_v0.3.tar.bz2
408K    3nch.pdb.gz
448K    frame0.trr
480K    1am7_corrected.xtc
480K    1am7_uncorrected.xtc
488K    well-mixed.hoomdxml
524K    frame0.mdcrd
552K    frame0.mdcrdbox
556K    frame0.gro
592K    frame0.xyz
628K    nitrogen.arc
692K    2EQQ.pdb
748K    frame0.pdb
820K    nitrogen.pdb
976K    tz2.truncoct.parm7
1.1M    tip3p_300K_1ATM.xtc
1.4M    monolayer.xtc
1.4M    overflow.crd
1.6M    test_hoomdxml.dcd
1.6M    water_hex.pdb.gz
1.8M    1vii_sustiva_water.pdb
1.9M    test_bad.nc
1.9M    test_good.nc
2.0M    frame0.dtr
2.6M    test.parm7
28M     3pqr_memb.psf
58M     total
```

@swails git blame is gitblaming you for 3pqr_memb.psf. Is there a smaller file that could be used for testing?

swails:
I can make it bigger :)

mpharrigan:
:100: MB

swails:
Actually, if I use ParmEd, I think I can reach the target:

```Python
import parmed as pmd
(pmd.load_file('3pqr_memb.psf') * 4).save('mdtraj_target.psf')
```

`mdtraj_target.psf` should be just over the 100 MB mark :)

swails:
Give me a second to find my serious hat and figure out why I added such a ginormous file in the first place.

mpharrigan:
[...]

PR #983, opened 2015-11-06
Skip large psf file if it does not exist

mpharrigan:
We can delete this file before creating the pypi
distribution.

cc #982


rmcgibbo:
lol

rmcgibbo:
If we decide that the size of the sdist is a problem, we should fix it systematically for all of the large files.

PR #986, opened 2015-11-07
Update devtools/README.md

rmcgibbo:
Instructions for building wheels.


mpharrigan:
what is twine?

rmcgibbo:
https://pypi.python.org/pypi/twine it let's you first build the dist, and
then upload it in a second step.

Issue #989, opened 2015-11-08
Win32 DSSP Crash


rmcgibbo:
On 32-bit Python, the Appveyor tests seem to stochastically fail/exit after printing `mdtraj.geometry.tests.test_dssp.test_6 ...` in a way that suggest that it's a segfault. The  crash doesn't happen every time, but appears to happen independently of py2 vs py3, and appears not to happen on 64-bit python.

Someone should debug this and fix the issue. I assume that it's a segfault in the dssp code related to the size of different data types.

PR #993, opened 2015-11-11
DSSP

rmcgibbo:
Logged onto windows. I can't seem to reproduce #989


mpharrigan:
Did this fix it?

rmcgibbo:
I think it "fixed" the test -- but note switched one PDB for another. For whatever reason, the `alanine-dipeptide-explicit.pdb` still causes the thing to crash on Appveyor, I think. But I couldn't get it to crash myself, on windows.

PR #995, opened 2015-11-16
[contact] convert list to array greatly improves performance

marscher:
In compute_contacts right after the call into compute_contacts, there is some inefficiency due to the usage of a python list for n_atoms_per_residue_pair.

See markovmodel/pyemma#448

Converting the list to an array greatly improved the runtime from 60s to 20s for the same data set.

### DIFF
--- a/mdtraj/geometry/contact.py
+++ b/mdtraj/geometry/contact.py
@@ -182,6 +182,8 @@ def compute_contacts(traj, contacts='all', scheme='closest-heavy', ignore_nonpro
         # now squash the results based on residue membership
         n_residue_pairs = len(residue_pairs)
         distances = np.zeros((len(traj), n_residue_pairs), dtype=np.float32)
+        n_atom_pairs_per_residue_pair = np.asarray(n_atom_pairs_per_residue_pair)
+
         for i in xrange(n_residue_pairs):
             index = int(np.sum(n_atom_pairs_per_residue_pair[:i]))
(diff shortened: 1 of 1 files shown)


marscher:
proof

rmcgibbo:
Thanks, @marscher!

marscher:
The speed of this function could be also improved if not all distances are being calculated and evaluated afterwards. One could do so in the distance c function to stop evaluate distances for a residue, if a distance   below the threshold already has been found.

@rmcgibbo
Would you accept a PR patching this functionality into the low level compute distance function?

cc @franknoe

rmcgibbo:
What threshold? The semantics are still "find the two closest atoms from the two residues", not just any pair within a certain distance, so I dont see how short-circuiting helps too much.

The real speed improvement comes from using a voxel-based data structure I assume.

franknoe:
@marscher , I think what you are referring to would be a different feature function. We often need to compute binary contacts, i.e. 1 if any pair of atoms between two groups is smaller than some threshold. In such a function, one could stop evaluating once the first pair below this threshold is found.

However, for large systems that might not save too much overall because most residues do not have contacts. For those, one could use triangle equalities - first compute maximum distance from any reference atom in the residue (e.g. Calpha), then compute distances between Calphas and exclude all pairs that are not in contact based on the triangle inequalities.

Those are two cheap optimizations for a binary contact function. The more sophisticated version would be to use an efficient neighbor structure such as a grid or a KD tree.

Issue #992, opened 2015-11-09
Support of __len__ in xyz files? 


ppxasjsm:
It would be useful if `__len__` in xyz files was implemented, so that there is a support for reading .xyz files with pyemma. 
See https://github.com/markovmodel/PyEMMA/issues/621#

rmcgibbo:
I'm not sure that this can be done efficiently, but that probably doesn't matter.

marscher:
Thank you @ctk3b and @rmcgibbo :+1:

PR #998, opened 2015-11-19
Support __len__ in xyz files

ctk3b:
Resolves: #992

### DIFF
--- a/mdtraj/formats/xyzfile.py
+++ b/mdtraj/formats/xyzfile.py
@@ -142,6 +142,7 @@ class XYZTrajectoryFile(object):
         self._filename = filename
         self._mode = mode
         self._frame_index = 0
+        self._n_frames = None
         # track which line we're on. this is not essential, but its useful
         # when reporting errors to the user to say what line it occured on.
         self._line_counter = 0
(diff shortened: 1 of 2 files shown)


swails:
Won't this exhaust the file iterator and prevent actually *parsing* the file?

marscher:
yes, it will. Probably it is better to create a helper method creating the iter and invoke that again in the length method.

ctk3b:
True. So would the best workaround be to open a new filehandle, iterate through that and report that?

marscher:
exactly

swails:
> True. So would the best workaround be to open a new filehandle, iterate through that and report that?

That's one option.  Another is to only support `__len__` on file handles that support `seek` and `tell` (then you can `tell()`, iterate through, then `seek()` back to where you told at the beginning).  This also only works on files that have not started iterating through already (i.e., `__len__` would change based on doing something like `iterload`).

Using a new file handle is probably the easiest way to do this, albeit possibly less efficient than some alternatives (that don't need to actually *parse* the file in Python, but can rather just look for newlines or byte counts).  But you know what they say about premature optimization...

marscher:
You should also cache this value, once you have computed it.

swails:
> You should also cache this value, once you have computed it.

+1. Easy optimization that reduces potential filesystem thrashing

marscher:
:+1:

rmcgibbo:
Can you add a quick test?

mdtraj/formats/xyzfile.py
```diff
@@ -383,4 +384,12 @@ def tell(self):
 
     def __len__(self):
         """Number of frames in the file. """
-        raise NotImplementedError()
+        if str(self._mode) != 'r':
+            raise NotImplementedError('len() only available in mode="r" currently')
+        if not self._is_open:
```
marscher: If you're opening a new handle, why do require the parsers file handle to be opened?

rmcgibbo: the fact that it opens a new handle should be considered an implementation detail.

rmcgibbo:
Thanks, guys!

PR #1001, opened 2015-12-03
assign segment_id to residues and add it to selection language

hockyg:
Both psf and pdb topology readers get a segment_id from files, but this is not stored anywhere. This is an extremely useful way to do selections for systems with complex topology (vs chains, which are numbered).

These changes have segment_id propagated back to the topology residue object, which has a default segment_id of an empty string. It requires adding several lines to preserve this segment_id when the topology is duplicated. 

Alternatively, add_residue could be modified to always take a segment_id.


rmcgibbo:
Thanks for writing this up, Glen! Looks like a great feature to add. Can you add a test or two to [test_topology.py](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/tests/test_topology.py) and potentially [test_pdb.py](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/pdb/tests/test_pdb.py) and [test_psf.py](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/tests/test_psf.py) to confirm that the feature works as expected? Also, do we need to modify any of the pdb writing code to make sure this data gets saved back out to disk?

I'll make a couple other comments inline.

mdtraj/core/topology.py
```diff
@@ -88,12 +88,17 @@ def _topology_from_subset(topology, atom_indices):
...
         for residue in chain._residues:
             resSeq = getattr(residue, 'resSeq', None) or residue.index
-            newResidue = newTopology.add_residue(residue.name, newChain,
-                                                 resSeq)
             for atom in residue._atoms:
                 if atom.index in atom_indices:
+                    if not residue == previous_residue:
+                        newResidue = newTopology.add_residue(residue.name, newChain,
+                                                             resSeq)
+                        newResidue.segment_id = residue.segment_id
```
rmcgibbo: instead of setting the attribute on the new object, `segment_id` should be added as a new argument to the `Topology.add_residue` factory function, to parallel the `resSeq` record, I think.

Also, what is the reason for moving this logic inside the for loop in this PR?

mdtraj/core/topology.py
```diff
@@ -1169,6 +1177,8 @@ class Residue(object):
         The chain within which this residue belongs
     resSeq : int
         The residue sequence number
+    segment_id : str
+        An optional label for the segment to which this residue belongs
     """
 
     def __init__(self, name, index, chain, resSeq):
```
rmcgibbo: It will also need to be added to the constructor here, probably with a default value of the empty string.

hockyg: Thanks Robert

Okay, let's change the constructor. I can do that and add tests. The writer
doesn't support this feature get. An additional point is I have an
improvement to pdstructure which fixes some edge cases related to this and
makes the ATOM parser match the letter of the law of the newer pdb spec at
the 70-80 character range. I didn't know if I should add this in the same
commit?

rmcgibbo: Cool! Another commit would probably be easier, but it's not a big deal either way.

mdtraj/core/topology.py
hockyg: I believe the reason for moving the new residue command is related to what happens when there are no selected atoms from a residue. The add_residue command increments the number of residues and the residue's serial number, and then decides whether it's empty or not. It seems like most of the time you want your new topology to have the number of residues equal to the number of non-empty residues. The resSeq takes care of the fact that moving this inside the loop renumbers the residues, since if you select by resSeq it should still be the old numbering scheme. Does that seem right?

hockyg:
Close this pull request in favor of #1002

PR #1002, opened 2015-12-04
add segment_id as residue attribute

hockyg:
Both psf and pdb topology readers get a segment_id from files, but this is not stored anywhere. This is an extremely useful way to do selections for systems with complex topology (vs chains, which are numbered).

Segment_id is now elevated to a main, although not required, property of the residue class. 
Segment_id is read from psf and pdb file readers and stored in this attribute. It has also been added to the data frame object list of columns (which could be a problem if anyone is storing topologies as data frames for future use). 
Segment_id or segname (vmd style) can be used in atom selections.

save_pdb has been adjusted to write out segment_id in the proper style. 
I believe I have also successfully added segment_id to the topology information in hdf5 writing without breaking backwards compatibility to previously saved hdf5 files (a warning is raised that there is no segment_id information). 

Tests have been added to make sure these features are working based on the tri-alanene pdb and psf files, which is not ideal since there is only a single segment name for the entire structure.

### DIFF
--- a/mdtraj/core/selection.py
+++ b/mdtraj/core/selection.py
@@ -116,6 +116,7 @@ class SelectionKeyword(object):
         (('residue', 'resSeq'), _chain('residue', 'resSeq')),
         (('resname', 'resn'), _chain('residue', 'name')),
         (('resid', 'resi'), _chain('residue', 'index')),
+        (('segment_id','segname',), _chain('segment_id')),
 
         # Atom.residue.chain.<attribute>
         (('chainid',), _chain('residue', 'chain', 'index')),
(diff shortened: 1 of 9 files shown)


rmcgibbo:
Is this an alternative to https://github.com/mdtraj/mdtraj/pull/1001?

hockyg:
This is a full replacement for #1001 according to suggestions of @rmcgibbo except for the unrelated change in that pull request related to topology_from_subset

rmcgibbo:
It looks like there are a couple test failures. You can fix them by adding more commits onto this branch -- you shouldn't need to recreate a new PR for it.

hockyg:
Oh whoops, I don't think I ran the full set of tests, and some skip because
I'm missing libraries on my test machine. I can do that and get back to you.

On Thu, Dec 3, 2015 at 8:23 PM, Robert T. McGibbon <notifications@github.com

rmcgibbo:
np. this is why CI is so useful  :)

rmcgibbo:
Thanks, @hockyg.

PR #1005, opened 2015-12-08
Bugfix: trajectory.slice() crashed when copy=False

schilli:
trajectory.slice() crashed when copy=False, because the variable `topology` remained unassigned.


schilli:
It's an
`HTTPError: HTTP Error 500: Internal Server Error`
raised by `urllib2.py`

It is not caused by my code changes and I also get that error when I run the tests on the master branch of mdtraj.

rmcgibbo:
Looks great. Thanks!

PR #1012, opened 2016-01-01
Clean up some cython code to use typed memoryviews

rmcgibbo:
This fixes some more of what used to be in #550.


mpharrigan:
cool :)

PR #1023, opened 2016-01-21
[xtc|trr] added xdrlib2 (seek support).

marscher:
This method will read an integer per frame to determine the byte offsets between
frames. It will only be called once the user wants to know the lengths of the
trajectory or he/she demands random access (seeking). The offsets are accessible
via a property of the XTC|TRRTrajectoryFile and could in principle cached.

It should be discussed/measured whether the approximations about the total number of frames made are obsolete by this method (of course one has to do a quick scan over the whole file, but it will probably save lots of reallocations/memcopies).

For those concerned about licensing issues, the original source code (xdrlib) is BSD, so a modification to this library should remain under this license. So there should be no issue with MDAnalysis is GPL and mdtraj is LGPL. But I'm no lawyer...

### DIFF
--- a/mdtraj/formats/xtc/include/ms_stdint.h
+++ b/mdtraj/formats/xtc/include/ms_stdint.h
@@ -1,0 +1,259 @@
+// ISO C9x  compliant stdint.h for Microsoft Visual Studio
+// Based on ISO/IEC 9899:TC2 Committee draft (May 6, 2005) WG14/N1124
+//
+//  Copyright (c) 2006-2013 Alexander Chemeris
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//   1. Redistributions of source code must retain the above copyright notice,
+//      this list of conditions and the following disclaimer.
+//
+//   2. Redistributions in binary form must reproduce the above copyright
+//      notice, this list of conditions and the following disclaimer in the
+//      documentation and/or other materials provided with the distribution.
+//
+//   3. Neither the name of the product nor the names of its contributors may
+//      be used to endorse or promote products derived from this software
+//      without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR IMPLIED
+// WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO
+// EVENT SHALL THE AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+// OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+// WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+// ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+///////////////////////////////////////////////////////////////////////////////
+
+#ifndef _MSC_VER // [
+#error "Use this header only with Microsoft Visual C++ compilers!"
+#endif // _MSC_VER ]
+
+#ifndef _MSC_STDINT_H_ // [
+#define _MSC_STDINT_H_
+
+#if _MSC_VER > 1000
+#pragma once
+#endif
+
+#if _MSC_VER >= 1600 // [
+#include <stdint.h>
+#else // ] _MSC_VER >= 1600 [
+
+#include <limits.h>
+
+// For Visual Studio 6 in C++ mode and for many Visual Studio versions when
+// compiling for ARM we should wrap <wchar.h> include with 'extern "C++" {}'
+// or compiler give many errors like this:
+//   error C2733: second C linkage of overloaded function 'wmemchr' not allowed
+#ifdef __cplusplus
+extern "C" {
+#endif
+#  include <wchar.h>
+#ifdef __cplusplus
+}
+#endif
+
+// Define _W64 macros to mark types changing their size, like intptr_t.
+#ifndef _W64
+#  if !defined(__midl) && (defined(_X86_) || defined(_M_IX86)) && _MSC_VER >= 1300
+#     define _W64 __w64
+#  else
+#     define _W64
+#  endif
+#endif
+
+
+// 7.18.1 Integer types
+
+// 7.18.1.1 Exact-width integer types
+
+// Visual Studio 6 and Embedded Visual C++ 4 doesn't
+// realize that, e.g. char has the same size as __int8
+// so we give up on __intX for them.
+#if (_MSC_VER < 1300)
+   typedef signed char       int8_t;
+   typedef signed short      int16_t;
+   typedef signed int        int32_t;
+   typedef unsigned char     uint8_t;
+   typedef unsigned short    uint16_t;
+   typedef unsigned int      uint32_t;
+#else
+   typedef signed __int8     int8_t;
+   typedef signed __int16    int16_t;
+   typedef signed __int32    int32_t;
+   typedef unsigned __int8   uint8_t;
+   typedef unsigned __int16  uint16_t;
+   typedef unsigned __int32  uint32_t;
+#endif
+typedef signed __int64       int64_t;
+typedef unsigned __int64     uint64_t;
+
+
+// 7.18.1.2 Minimum-width integer types
+typedef int8_t    int_least8_t;
+typedef int16_t   int_least16_t;
+typedef int32_t   int_least32_t;
+typedef int64_t   int_least64_t;
+typedef uint8_t   uint_least8_t;
+typedef uint16_t  uint_least16_t;
+typedef uint32_t  uint_least32_t;
+typedef uint64_t  uint_least64_t;
+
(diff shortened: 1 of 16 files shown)


mpharrigan:
It doesn't matter that xdrlib is BSD, xdrlib2 is GPL which is incompatible with mdtraj's LGPL

rmcgibbo:
- Do you have any measurements showing that this method of determining the byte offsets actually helps? It still has to iterate over the whole file. Regardless of how many integers need to be read from the file, all the memory still needs to be paged in, which I assume is more expensive than the in-cache comutation afterwards. The easiest way to read the byte offsets would just be to call tell() a bunch of times as you iterate though.
- Also, this code leaks the offset buffer, I think.

rmcgibbo:
> It doesn't matter that xdrlib is BSD, xdrlib2 is GPL which is incompatible with mdtraj's LGPL

Oh, sorry. Maybe I didn't read that clearly enough the first time. If this contributed code is only available only under the terms of the GPL then it's not compatible.

marscher:
Well it is questionable that xdrlib2 has its own license, since that would have required a re-licensing from BSD to GPL.
The offsets buffer is not leaked, since it is being wrapped in a numpy array, which explicitly takes the ownership of that memory region.
Concerning performance: I will add some measurements - stand by.

rmcgibbo:
> Well it is questionable that xdrlib2 has its own license, since that would have required a re-licensing from BSD to GPL.

I don't think there's any inconsistency here. But perhaps you could email the true author of the patch and ask.

mpharrigan:
For the record, it looks like mdanalysis doesn't use this code anymore https://github.com/MDAnalysis/mdanalysis/pull/441

marscher:
Regarding timings for seeking on a XTC files containing 120066 atoms, 38951 residues and 9600 frames, the following benchmark script leads to the results:
``` python 
from mdtraj.formats import XTCTrajectoryFile as xtcfile
length = 9600
frames = (100, 300, 1000, 5000, 9600-1)
def seeking():
    with xtcfile('output.xtc') as fh:
        for pos in frames:#xrange(9600):
            fh.seek(pos)
            print "pos:", fh.tell()

%timeit seeking()
```
mdtraj-1.5.1: 1 loops, best of 3: 2min 36s per loop
mdtraj-1.6.0.dev1.dev-ca58fb5: 10 loops, best of 3: 33.2 ms per loop

marscher:
I also fixed a bug in "xdrlib", where the offset array is totally wrong beyond the 2 GB area. So now we have the working full speed also for "large" files. I've tested it locally with a 20GB file with tiny frames. Index creation takes like ~5 seconds (SSD for the win) and seeking is blazingly fast. Since this feature is mission critical in PyEMMA, I'd love to see this in the next release.

marscher:
@gph82, @franknoe FYI: random access for XTC file format.

franknoe:
Sounds great. What about the licensing issue? I agree we can't use GPL without permission.

gph82:
this is awesome for my save_trajs and the like!

On 01/25/2016 07:40 PM, Frank Noe wrote:
> Sounds great. What about the licensing issue? I agree we can't use GPL without
> permission.
>
> —
> <https://github.com/mdtraj/mdtraj/pull/1023#issuecomment-174615930>.
>

-- 
Dr. Guillermo Pérez-Hernández
Freie Universität Berlin
Institute for Mathematics
Arnimallee 6
D-14195 Berlin
tel       0049 30 838 75776

rmcgibbo:
Thanks for adding all this stuff @marscher. I still am very wary about the copyright thought.

franknoe:
In case if the author doesn't agree - would splitting the xtc stuff off 
in a GPL package and creating a dependency to it solve the issue?
Other alternatives? What does MDAnalysis use now?

Am 25/01/16 um 20:02 schrieb Robert T. McGibbon:
>
> Thanks for adding all this stuff @marscher 
> <https://github.com/marscher>. I still am very wary about the 
> copyright thought.
>
> —
> <https://github.com/mdtraj/mdtraj/pull/1023#issuecomment-174622494>.
>


-- 

----------------------------------------------
Prof. Dr. Frank Noe
Head of Computational Molecular Biology group
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de
Mail:  Arnimallee 6, 14195 Berlin, Germany
----------------------------------------------

rmcgibbo:
I think you could probably [clean room](https://en.wikipedia.org/wiki/Clean_room_design) re-write the feature.

marscher:
On 25.01.2016 20:10, Robert T. McGibbon wrote:
> I think you could probably clean room
> <https://en.wikipedia.org/wiki/Clean_room_design> re-write the feature.

LOL, I'm not a Samsung engineer stealing latest Apple design and want to
avoid a multi billion dollar case. Seriously, copyright is superstition
in this case.

Maybe the authors of the lib will give their permission anyway, since I
also fixed a (serious) bug in it.

franknoe:
Erm, this is not about money, but we have to respect the licence. Please 
ask the author.

Am 25/01/16 um 21:17 schrieb Martin K. Scherer:
> On 25.01.2016 20:10, Robert T. McGibbon wrote:
> > I think you could probably clean room
> > <https://en.wikipedia.org/wiki/Clean_room_design> re-write the feature.
>
> LOL, I'm not a Samsung engineer stealing latest Apple design and want to
> avoid a multi billion dollar case. Seriously, copyright is superstition
> in this case.
>
> Maybe the authors of the lib will give their permission anyway, since I
> also fixed a (serious) bug in it.
>
> —
> <https://github.com/mdtraj/mdtraj/pull/1023#issuecomment-174648770>.
>


-- 

----------------------------------------------
Prof. Dr. Frank Noe
Head of Computational Molecular Biology group
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de
Mail:  Arnimallee 6, 14195 Berlin, Germany
----------------------------------------------

hainm:
>  but we have to respect the licence. Please ask the author.

absolutely agree with this.

kain88-de:
I'm the author of the recent changes in MDAnalysis

# First the Licenses

I have written all the cython-wrapper code and that specific code is licensed under the GPLv2. I hereby give you permission to use that code in mdtraj with your license. 

The changes to libxdrfile have been done by @mnmelo. You should ask him for those.

# Leaking offsets buffer

You are currently leaking the offset buffer.  This happens because you don't tell numpy about the actual size of the offsets buffer.  This is the reason I'm using a special [arraywrapper](https://github.com/MDAnalysis/mdanalysis/blob/develop/package/MDAnalysis/lib/formats/cython_util.pyx). Even without the size issue you should use this wrapper to convert c-style arrays to numpy arrays because afail `PyArray_SimpleNewFromData` doesn't take ownership of the pointer. Of course you can have the wrapper with the LGPL license as well.

rmcgibbo:
I think by setting the OWNDATA flag, the memory is handled properly: http://docs.scipy.org/doc/numpy-1.10.1/user/c-info.how-to-extend.html#c.PyArray_SimpleNewFromData

kain88-de:
Nope this is not enough in this case. He is still leaking memory. The `offsets` array is allocated upto `est_nframes` and not `nframes`. `est_nframes` is an overestimate of the actual number of frames. Any numpy array derived from `offsets` should be of size `est_nframes`. This is what [I'm doing](https://github.com/MDAnalysis/mdanalysis/blob/5330a7c7a7e0e780824f10a445567757c9355862/package/MDAnalysis/lib/formats/xdrlib.pyx#L337), that is the reason the `read_xtc_nframes` functions returns that value.

With the owndata flag. Does that work for different numpy versions? I found a [stackoverflow](http://stackoverflow.com/questions/8708758/can-i-force-a-numpy-ndarray-to-take-ownership-of-its-memory#18312855) post saying that it might be problematic for numpy <1.7. But I'm not sure about that.

rmcgibbo:
> Any numpy array derived from offsets should be of size est_nframes.

In the end it won't really matter though, because it's just calling `free` on the pointer to the beginning of the segment. @marscher you could allocate the array with the bigger size but then just return a smaller view of it.

I think that SO post just shows that you might need to use a slightly different call signature to compile against earlier versions of the Numpy C API. I believe we run some of our CI tests against Numpy 1.6, so I assume that any issues here will get fixed up before merge.

mnmelo:
Thanks for pointing the out the licensing question, @kain88-de.
As the copyright holder of the libxdrfile2 changes I decided [back then](https://groups.google.com/forum/#!searchin/mdnalysis-devel/Relicensing$20of$20libxdrfile$20to$20GPL/mdnalysis-devel/WCXKesf3Soo/sQVA8WH8brwJ) to publish them under GPLv2, and this was agreed on by the MDAnalysis dev team (notice the contrast to the previous LGPL licencing of the libxdrfile bundled with MDAnalysis).

This was done expressly so. If you include part of the libxdrfile2 code you must follow the GPLv2 when distributing. Of course, since the base Gromacs libxdrfile library is LGPL (or, apparently now, BSD), you can base yourselves on that without stricter obligations. GPL enforcement essentially boils down to the indexing additions I made to libxdrfile.

rmcgibbo:
Okay, thanks for clarifying, @mnmelo. Obviously we'd appreciate a license exemption but I understand you're position as the copyright holder. I think that basically settles it, so I'm going to close this PR.

I would encourage the interested parties to independently reimplement the feature (it's not so complex) without consulting any separately-copyrighted code.

franknoe:
I agree.

Am 26/01/16 um 00:03 schrieb Robert T. McGibbon:
>
> Okay, thanks for clarifying, @mnmelo <https://github.com/mnmelo>. 
> Obviously we'd appreciate a license exemption but I understand you're 
> position as the copyright holder. I think that basically settles it, 
> so I'm going to close this PR.
>
> I would encourage the interested parties to independently reimplement 
> the feature (it's not so complex) without consulting any 
> separately-copyrighted code.
>
> —
> <https://github.com/mdtraj/mdtraj/pull/1023#issuecomment-174718132>.
>


-- 

----------------------------------------------
Prof. Dr. Frank Noe
Head of Computational Molecular Biology group
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de
Mail:  Arnimallee 6, 14195 Berlin, Germany
----------------------------------------------

mnmelo:
Thanks for understanding and complying, everyone!
If I may suggest something: building an index on the python side could work, and would work around the licensing. The speed penalty might not be much, especially if you can get bindings to let you skip whole frames after reading the header.
Good luck!

PR #1030, opened 2016-02-03
Remove very annoying warnings filter

rmcgibbo:
Sorry for all the pain I caused you, @mpharrigan.

### DIFF
--- a/mdtraj/utils/__init__.py
+++ b/mdtraj/utils/__init__.py
@@ -18,10 +18,6 @@ __all__ = ["ensure_type", "import_", "in_units_of",
            "enter_temp_directory", "timing", "deprecated"]
 
 
-# Make sure that DeprecationWarning get printed
-warnings.simplefilter("always", DeprecationWarning)
-
-
 def ilen(iterable):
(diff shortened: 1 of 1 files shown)


mpharrigan:
Oh, I was going to submit a pr that uses the module filter so we can still get mdtraj warnings

mpharrigan:
maybe even that's a little overbearing

rmcgibbo:
I guess that would be fine. I think it seems better perhaps just to not mess with the filters and let the client decide.

mpharrigan:
yeah, let's not mess with it

marscher:
Ah there we got it... Thank you! It caused a lot of trouble with new IPython notebook version deprecation warnings spam. Hope this will be released soon.

Carreau:
Having filter turned on for a specific module is relatively easy. 
IPython does turn the DeprecationWarning only in `__main__` on by default. We are pushing to have that true in Python repl as well. you can help pushing if you like.

http://bugs.python.org/issue24294

I'm working on simpler way to filter warning in CI when using nose:

https://github.com/Carreau/nose_warnings_filters

If it can helps you...

PR #1038, opened 2016-02-09
[xtc|trr] added seek support to provide fast random access.

marscher:
This method will read an integer per frame to determine the byte offsets between
frames for XTC files. TRR files on the other hand store the sizes of a frame in
a private data structure in xdrlib (which has been exposed to handle this).

For small XTC files (<= 9 atoms), the coordinates are written uncompressed,
which is reflected in the offset calculation.

Note that the amount of bytes needed for frames can be figured out by reading
the original source code of xdrlib (GROMACS).

The byte offset array will only be constructed if requested by the user,if he/she
wants to know the lengths of the trajectory or he/she demands random access
(seeking). The offsets are accessible via a property of the
XTC|TRRTrajectoryFile and cached (stored to disk and restored via the setter).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,6 +4,7 @@ environment:
     # /E:ON and /V:ON options are not enabled in the batch script intepreter
     # See: http://stackoverflow.com/a/13751649/163740
     CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\devtools\\appveyor-ci\\run_with_env.cmd"
+    PYTHONUNBUFFERED: 1
 
   matrix:
     - PYTHON: "C:\\Miniconda"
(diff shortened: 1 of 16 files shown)


rmcgibbo:
Wow, this is excellent. Thanks so much for adding this, @marscher!

rmcgibbo:
One of the Appveyor issues is just Appveyor being flakey, but the other could be real (?)

marscher:
There was a problem in the uncompressed xtc offset calculation (mixing Python ints with int64 is a bad thing on 32 bit systems).

marscher:
rebasing in progress... :+1:

mpharrigan:
Thanks! This looks good

PR #1041, opened 2016-02-13
Force overwrite

### LINKED ISSUE #877: `force_overwrite` ignored for .xtc

mpharrigan:
supersedes #878 
fixes #877

### DIFF
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -1572,6 +1572,8 @@ class Trajectory(object):
         ----------
         filename : str
             filesystem path in which to save the trajectory
+        force_overwrite : bool, default=True
+            Overwrite anything that exists at filename, if it's already there
         """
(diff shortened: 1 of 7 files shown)


rmcgibbo:
@mpharrigan i merged your commits here into #878 and then fixed the tests, so #878 should pass on the CI now.

Issue #705, opened 2015-01-29
Refactor baker_hubbard and wernet_nilsson


rmcgibbo:
1. The two functions should have a consistent API, but they don't.
2. They should use `md.compute_neighbors` to be more memory efficient.
3. They should make sense for applications like "compute all the hydrogen bonds made by this specific subset of the atoms (e.g. one or more residues, small molecule), in every frame."

IMO, we're probably going to have to change the function signature for `baker_hubbard`, especially the `freq` argument.

cc: http://discourse.mdtraj.org/t/tracking-of-hbonds-between-two-groups

PR #1045, opened 2016-02-16
Add optional parameters to specify hbond donors and acceptors of interest

juliebehr:
Added optional inputs `proposed_donor_indices` and `proposed_acceptor_indices` to `geometry.hbond.wernet_nilsson` and `geometry.hbond.baker_hubbard`.  

Both arguments are `np.ndarray`s of desired atom indices, `shape=(n_donor_indices,)`, dtype=`int`, default=`None`.

I'm using these to find the hydrogen bonds formed by a single residue of interest.  I also added a check for defining `xh_donors` if either `oh_donors` or `nh_donors` is empty, because `np.array(nh_donors + oh_donors)` raised a `ValueError` if one is empty and the other is not.


mdtraj/geometry/hbond.py
```diff
@@ -123,10 +127,13 @@ def wernet_nilsson(traj, exclude_water=True, periodic=True):
         raise ValueError('wernet_nilsson requires that traj contain topology '
                          'information')
 
-    def get_donors(e0, e1):
+    def get_donors(e0, e1, proposed_donor_indices=None):
         elems = set((e0, e1))
         bonditer = traj.topology.bonds
-        atoms = [(b[0], b[1]) for b in bonditer if set((b[0].element.symbol, b[1].element.symbol)) == elems]
+        if proposed_donor_indices==None:
```
mpharrigan: It's better to use `if x is None` when looking for None values in python

http://stackoverflow.com/a/14247383

mdtraj/geometry/hbond.py
```diff
@@ -141,9 +148,14 @@ def get_donors(e0, e1):
 
         return indices
 
-    nh_donors = get_donors('N', 'H')
-    oh_donors = get_donors('O', 'H')
-    xh_donors = np.array(nh_donors + oh_donors)
+    nh_donors = get_donors('N', 'H', proposed_donor_indices=proposed_donor_indices)
+    oh_donors = get_donors('O', 'H', proposed_donor_indices=proposed_donor_indices)
+    if nh_donors != [] and oh_donors != []:
```
mpharrigan: Similarly, the more robust thing to do here is check `if len(x) > 0` (see line 160 for an example)

mpharrigan:
Thanks for the contribution! Could you:

 - try to wrap lines at 80 characters per pep8, especially in the long list comprehensions. I know many lines in this file don't follow this guideline, but we try to get closer to pep8 with each contribution. A good way to wrap the list comprehensions would be something like

```python
my_list = [some_func(x)
           for x in some_iterator
           if x == some_condition]
```
 - add a quick test to make sure the new options are working as you expect, and to make sure no one breaks the functionality in the future

 - I made a couple in-line comments to fix as well

juliebehr:
Sure! Making these changes now.

mpharrigan:
cc'ing #705 for the record

juliebehr:
Tests have been added and pep8 guidelines incorporated.  There are a few long list comprehensions which I think have technically been indented correctly but it makes them look a little goofy; I'm happy to align them differently if preferred.

mdtraj/geometry/hbond.py
```diff
@@ -151,9 +181,31 @@ def get_donors(e0, e1):
...
+                         if (a.element.symbol == 'O'
+                             and a.residue.name != 'HOH')
+                            or a.element.symbol == 'N']
+        else:
+            acceptors = [a.index
+                         for a in traj.topology.atoms
+                         if ((a.element.symbol == 'O'
+                              and a.residue.name != 'HOH')
+                             or a.element.symbol == 'N')
+                            and a.index in proposed_acceptor_indices]
```
juliebehr: For example, here

ctk3b: You could define a small function inside here:

```python
def is_acceptor(atom):
    return (a.element.symbol == 'O' and a.residue.name != 'HOH')
            or a.element.symbol == 'N'
```

juliebehr:
I've added a function ```is_acceptor(atom)``` per suggestion by @ctk3b.  Please let me know if I should make any other changes!

mdtraj/geometry/hbond.py
```diff
@@ -143,17 +166,29 @@ def get_donors(e0, e1):
...
+        if exclude_water and atom.residue.name == 'HOH':
+            return False
+        if (proposed_acceptor_indices is not None and
+            atom.index not in proposed_acceptor_indices):
+            return False
+        return (atom.element.symbol == 'O' or atom.element.symbol == 'N')
+
+    acceptors = [a.index
+                 for a in traj.topology.atoms
+                 if is_acceptor(a)]
```
juliebehr: New ```is_acceptor``` function, above, is exactly the same in ```wernet_nilsson``` and ```baker_hubbard```.

cxhernandez: Perhaps this should be added as an `Atom` property?

rmcgibbo:
This is outdated. Feel free to re-open if desired.

Issue #1047, opened 2016-02-18
`mdtraj.Topology.from_openmm` fails due to missing `residue.segment_id`


jhprinz:
I recently updated mdtraj, openmm and openmmtolls to
```
mdtraj                    1.6.1               np110py27_0
openmm                    6.3.1                    py27_1
openmmtools               0.7.5                    py27_1
```
and when trying to convert a Topology from a testsystem to mdtraj Topology I get the following error

```py
import openmmtools as omt
import mdtraj as md
testsystem = omt.testsystems.AlanineDipeptideVacuum()
print testsystem.topology
>>> <Topology; 1 chains, 3 residues, 22 atoms, 21 bonds>

md.Topology.from_openmm(testsystem.topology)
```

resulting in

```py
    370             c = out.add_chain()
    371             for residue in chain.residues():
--> 372                 r = out.add_residue(residue.name, c, residue.segment_id)
    373                 for atom in residue.atoms():
    374                     if atom.element is None:

AttributeError: 'Residue' object has no attribute 'segment_id'
```

I assume that openmm has changed its Topology object somehow. Will this be fixed with the upcoming openmm 7.0 or should I make PR to quick fix this. 

From what I can tell they renamed `segment_id` to `id`?

andrrizzi:
I can confirm this. I had the same problem and had to downgrade to ``mdtraj 1.5.1``. [This is the commit](https://github.com/mdtraj/mdtraj/commit/c4d43469705f395e21d2a6dc789223f4e0bbc6d3#diff-ce95bbb0db9dbb5dce66ba686ed67f2d) that changed it.

jchodera:
Has this been fixed yet?  Is it safe to upgrade to the latest `mdtraj`?

jchodera:
This is still present in the latest `mdtraj`:
https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/topology.py#L351-L372
```python
    @classmethod
    def from_openmm(cls, value):
        """Create a mdtraj topology from an OpenMM topology
        Parameters
        ----------
        value : simtk.openmm.app.Topology
            An OpenMM topology that you wish to convert to a
            mdtraj topology.
        """
        app = import_('simtk.openmm.app')

        if not isinstance(value, app.Topology):
            raise TypeError('value must be an OpenMM Topology. '
                            'You supplied a %s' % type(value))

        out = cls()
        atom_mapping = {}

        for chain in value.chains():
            c = out.add_chain()
[...]

PR #1048, opened 2016-02-18
Rename `segment_id` to `id`

jhprinz:
Quickfix to adapt to changes from openmm. They seem to have changed `residue.segment_id` to `residue.id` this causes `mdtraj.Topology.from_openmm` to fail.

Resolves issue #1047.

### DIFF
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -369,7 +369,7 @@ class Topology(object):
         for chain in value.chains():
             c = out.add_chain()
             for residue in chain.residues():
-                r = out.add_residue(residue.name, c, residue.segment_id)
+                r = out.add_residue(residue.name, c, residue.id)
                 for atom in residue.atoms():
                     if atom.element is None:
                         element = elem.virtual


mpharrigan:
When was this changed in openmm? Is there any way to make this work with old and new versions? Is there a test we can add to make sure this doesn't happen again?

rmcgibbo:
there's: c4d43469705f395e21d2a6dc789223f4e0bbc6d3. this was an mdtraj change. i'm not exactly sure if `id` `segment_id` are supposed to be same thing.

juliebehr:
I assume @rmcgibbo is correct that ```id``` and ```segment_id``` are 2 different things, but without @jhprinz's fix ```mdtraj.Topology.from_openmm``` attempts to find an attribute called ```segment_id``` in the openmm residue and fails.  

I've looked at topology.py [#L1235](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/topology.py#L1235) and I'm not positive what the meaning of ```segment_id``` is intended to be, but maybe taking ```residue.chain.id``` from the openmm topology would contain the intended information?

juliebehr:
Or, if the meaning of ```segment_id``` is something different entirely, then perhaps the best behavior here is the default where ```segment_id=''``` because the openmm topology will not have this information.

mpharrigan:
cc #1002 cc @hockyg

> if the meaning of segment_id is something different entirely, then perhaps the best behavior here is the default where segment_id='' because the openmm topology will not have this information.

This sounds good to me. I think segment_id *is* something different

mpharrigan:
Subsumed by #1088

hockyg:
Just to clarify, segment id is a custom label for a section of a protein that is more expressive than chain id (4 letters max in a pdb file). It is very useful for setting up systems more complex than 1 protein. I agree that having it as an empty string by default in all cases is appropriate

mpharrigan:
Thanks for the clarification @hockyg

Issue #1055, opened 2016-02-22
to_fasta() should return '?' if no AA code is found for a particular position


cxhernandez:
Rather than omit a `rescode` if one isn't found (e.g. non-standard amino acid), maybe `to_fasta` should return '?'.

example: MET-TYR-ACK-ARG-GLU-SER should return 'MY?RES' rather than 'MYRES'.

rmcgibbo:
That seems reasonable.

PR #1056, opened 2016-02-22
have to_fasta return '?' if no AA code is found 

cxhernandez:
re: #1055

### DIFF
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -290,8 +290,12 @@ class Topology(object):
         fasta : String or list of Strings
            A FASTA string for each chain specified.
         """
-        fasta = lambda c: "".join([res.code for res in c.residues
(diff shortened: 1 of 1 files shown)


cxhernandez:
One potential drawback that I'm realizing now is that [`_PROTEIN_RESIDUES`](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/residue_names.py#L177) is just populated by keys of `_AMINO_ACID_CODES`, so this really only works for `NME` and `ACE`....

A couple of options could be:
+ Return '?' for any kind of residue that isn't found (including waters and small molecules where this doesn't really make sense)
+ Make `is_protein` a settable property
+ Forget about this PR

rmcgibbo:
This seems outdated. Feel free to re-open if not.

PR #1057, opened 2016-02-23
Created compute_neighborlist() function

peastman:


peastman:
I think this is ready to merge.  Travis reports a test failure on Python 3.5, but it's clearly unrelated to this.  And I can't tell what's up with AppVeyor.  I suspect it's just timing out on unrelated things.

PR #1058, opened 2016-02-24
[WIP] Created function for imaging molecules

peastman:
This implementation should be considered preliminary.  It hasn't been at all thoroughly tested.  I tried it on various PDB files and it always seemed to do something reasonable, but I didn't have anything especially challenging on hand.  I'd really like to see how it handles a long simulation of a large complex, or something like that.  So if you have any good test cases, please try it out.

I also added a method to Topology for identifying molecules based on bonds.  I needed it for this feature, and since it seemed a generally useful thing to have, I made it pubic.

Is `image_molecules()` a good name?  Any ideas for a better one?

### DIFF
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -62,6 +62,7 @@ from mdtraj.utils import (ensure_type, in_units_of, lengths_and_angles_to_box_ve
 from mdtraj.utils.six.moves import xrange
 from mdtraj.utils.six import PY3, string_types
 from mdtraj import _rmsd
+from mdtraj import compute_distances
 from mdtraj import FormatRegistry
 from mdtraj.geometry import distance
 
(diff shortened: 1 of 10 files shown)


mpharrigan:
I'll check this on my ion channels

cxhernandez:
Is there anything wrong with just calling it `image` or `reimage`?

cxhernandez:
Also: thank you!!!

mpharrigan:
I got a memory error 
`<mdtraj.Trajectory with 1530 frames, 82545 atoms, 18680 residues, and unitcells at 0x7fdf829787b8>`

```
MemoryError                               Traceback (most recent call last)
<ipython-input-11-7cb9f17c783a> in <module>()
----> 1 timage = traj.image_molecules()

/home/harrigan/opt/miniconda3/lib/python3.4/site-packages/mdtraj-1.7.0.dev0-py3.4-linux-x86_64.egg/mdtraj/core/trajectory.py in image_molecules(self, inplace)
   1872             for mol2 in range(mol1):
   1873                 pairs = np.array([(atom1.index,atom2.index) for atom1 in anchor_molecules[mol1] for atom2 in anchor_molecules[mol2]], dtype=int)
-> 1874                 dists = compute_distances(self, pairs)
   1875                 for frame in range(dists.shape[0]):
   1876                     nearest_pair_index = np.argmin(dists[frame])

/home/harrigan/opt/miniconda3/lib/python3.4/site-packages/mdtraj-1.7.0.dev0-py3.4-linux-x86_64.egg/mdtraj/geometry/distance.py in compute_distances(traj, atom_pairs, periodic, opt)
     77         orthogonal = np.allclose(traj.unitcell_angles, 90)
     78         if opt:
---> 79             out = np.empty((xyz.shape[0], pairs.shape[0]), dtype=np.float32)
     80             _geometry._dist_mic(xyz, pairs, box.transpose(0, 2, 1).copy(), out, orthogonal)
     81             return out

MemoryError: 
```

mpharrigan:
It runs with fewer frames in the trajectory, but there seem to be imaging issues. Investigating

mpharrigan:
It looks like there's something wrong with the imaging itself:

btw, how about factoring out the heuristics for anchor molecules into functions so you can see what it's picking

screenshot from 2016-02-24 16 17 58

peastman:
@rmcgibbo would it be possible to add an option to `compute_distances()` to only compute them for a single frame?  I only need them for one frame at a time, and it looks like it's running out of memory trying to create a giant array that will contain all the required distances for all 1530 frames at once.

rmcgibbo:
Yeah, it would be possible. I'm worried it might be too slow though -- maybe all of this should be done in C++.

rmcgibbo:
But perhaps that's to see for later.

mdtraj/core/trajectory.py
```diff
@@ -1818,3 +1819,122 @@ def _check_valid_unitcell(self):
...
+        for mol in anchor_molecules:
+            anchor_atom_indices += [atom.index for atom in mol]
+
+        # Compute the distance between each pair of anchor molecules in each frame.
+
+        anchor_dist = np.zeros((self.n_frames, num_anchors, num_anchors))
+        anchor_nearest_atoms = np.zeros((self.n_frames, num_anchors, num_anchors, 2), dtype=int)
+        for mol1 in range(num_anchors):
+            for mol2 in range(mol1):
+                pairs = np.array([(atom1.index,atom2.index) for atom1 in anchor_molecules[mol1] for atom2 in anchor_molecules[mol2]], dtype=int)
```
marscher: consider numpy.fromiter instead of explicitly creating this list

peastman:
Using some C++ may be necessary.  All I really need is this: given two molecules and their positions in one frame, identify the pair of atoms that form the closest contact between the two molecules.  I could write that in Python, but it would be really slow.  `compute_distances()` is faster, but requires creating a huge output array with far more information than I really need.  Perhaps I should add a new function in distance.py, implemented in C++, to compute that?

peastman:
Ok, new version.  I created a find_closest_contact() function that does exactly the calculation I needed, eliminating the need for huge arrays of distances.  See if it works better now.

peastman:
Anyone?

rmcgibbo:
(cc @mpharrigan). I assume they're still at the biophysical society conference.

mpharrigan:
@peastman did you see above that even when I did a couple frames (to get rid of the memory problem) there were some imaging issues

Did you change any of the imaging code with your recent commits? I can test after we get back from bps conference

peastman:
Yes, there are changes that might affect that.  I don't have your file, though, so I don't know for sure what was causing the problem with it.

peastman:
@mpharrigan I still need your file to reproduce this.

mpharrigan:
sent you a dropbox link via email

peastman:
Ok, new version checked in.  Try it now.

mpharrigan:
Wow! This is a huge game changer. Works like a charm

mpharrigan:
works with 80k atom, homo dimer membrane protein and 170k atom homo tetramer membrane protein

cxhernandez:
appveyor is failing:

```
geometry2.cpp
mdtraj/geometry/src/geometry2.cpp(38) : error C2065: 'NULL' : undeclared identifier
mdtraj/geometry/src/geometry2.cpp(51) : error C2065: 'NULL' : undeclared identifier
error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio 9.0\\VC\\Bin\\amd64\\cl.exe' failed with exit status 2
 
C:\Miniconda-x64\conda-bld\work>if errorlevel 1 exit 1 
Command failed: C:\windows\system32\cmd.exe /c call bld.bat

```

peastman:
No idea why.  `NULL` is a keyword in C++, and setup.py specifies `language='C++'` for that extension.  Anyone know what's going on?

rmcgibbo:
http://stackoverflow.com/a/15549896/1079728?

On Tue, Mar 8, 2016 at 3:00 PM, peastman <notifications@github.com> wrote:

> No idea why. NULL is a keyword in C++, and setup.py specifies
> language='C++' for that extension. Anyone know what's going on?
>
> —
> <https://github.com/mdtraj/mdtraj/pull/1058#issuecomment-194007831>.
>



-- 
-Robert

peastman:
AppVeyor is happy now.  Anything else that should be changed before we merge this?  For example, is the function name ok, or do we want to change it?

cxhernandez:
>Is there anything wrong with just calling it image or reimage?

peastman:
Well, `image()` sounds like it's going to return an image.  `reimage()` would be more reasonable.  But the question is whether it's better than the current name.

cxhernandez:
I'd prefer `reimage` just on the basis of being fewer characters to type out but that's about it.

mpharrigan:
Something shorter would "fit in" better with the rest of the mdtraj api, but this should be @peastman 's call, since he's very good at this sort of thing and actually implemented it :)

peastman:
I don't really have a strong opinion about it!  Fitting in well with the rest of the API is very important.  Although Trajectory already has other methods of similar length (`remove_solvent`, `restrict_atoms`, `center_coordinates`, etc.).  So really, I'm happy to follow the consensus here.  If people prefer `reimage` to `image_molecules`, I can change it.

swails:
I have a slight preference for `image_molecules`.  It is more descriptive and less ambiguous.  It describes pretty precisely what it does.  For instance, imaging atoms (which is something I believe GROMACS does unless told otherwise) is something different (and usually less useful).  `image_molecules` is really not a long name, and IMO is actually more consistent with the rest of the API (e.g., `compute_distances` or `compute_displacements` vs. `distance` or `displacement`).

But it's just a very slight preference.  I think `reimage` is pretty generally understood as well.

cxhernandez:
>Although Trajectory already has other methods of similar length (remove_solvent, restrict_atoms, center_coordinates, etc.).

Good point.

>For instance, imaging atoms (which is something I believe GROMACS does unless told otherwise) is something different (and usually less useful).

Ah okay, I wasn't aware of this. `image_molecules` is probably a better name then.

Issue #1034, opened 2016-02-05
Is there a way to load  a part of trajectory? or lazy loading?


lambdalisue:
I noticed that `md.load()` takes times more than I expected. I only need a part of trajectory so I wrote a  code something like:

```python
# It seems md.load() loads everything so takes long time
t = md.load(netcdf, top=prmtop)
distances = md.compute_distances(t[1000:2000], atoms_pair)

# It returns a generator so not subscriptable but I just need coordinates from 1000
# so I don't want to iterate over until 1000.
t = md.iterload(netcdf, top=prmtop)
distances = md.compute_distance(t[1000:2000], atoms_pair)
# => 'generator' object is not subscriptable
```

But it seems that `md.load()` load everything first. **So I wonder if there is a way to load a part of trajectory or perform some lazy loading within MDTraj** (not by making a copy of trajectory file by using other software.)

Additionally, I'm not sure if I could but if I made a PR for a lazy loading (such as load fundamental informations like topology or nframes and load coordinates whenever the users access to `trajectory.xyz` attribute first time)  are you interested in?
I personally use netcdf for Amber so even I could, I'll make only the one for that though.

rmcgibbo:
There is a function called `md.iterload`, which might service your purposes. There is an example of one way to use it here: http://mdtraj.org/latest/examples/iterload.html. There is also the function `md.open`, which will open the underlying trajectory file object, which is a file-like object for accessing the trajectory. 

One of the reasons that there isn't exactly a complete lazily-loaded trajectory object is that for most trajectory files, you can't do a random `seek` to a given frame in the trajectory on `O(1)` time, so the APIs exposed by the low-level classes like `md.NetCDFTrajectoryFile`, `md.DCDTrajectoryFile`, etc, doesn't give random access, but instead gives sequential access with `.read()`, `.seek()`, `.tell()`, etc.

lambdalisue:
> One of the reasons that there isn't exactly a complete lazily-loaded trajectory object is that for most trajectory files, you can't do a random seek to a given frame in the trajectory on O(1) time, so the APIs exposed by the low-level classes like md.NetCDFTrajectoryFile, md.DCDTrajectoryFile, etc, doesn't give random access, but instead gives sequential access with .read(), .seek(), .tell(), etc.

Then, I'm wondering why you don't use `f.seek()` to skip parsing unwilling part of the trajectory in the core part of MDTraj.

[...]

PR #1060, opened 2016-02-25
Improve load_frame to support partial loading

lambdalisue:
I'm not really sure if you guys like this approach but I think partial loading is fundamental feature so I would like to ask you to review.

While there are `load_frame` function and it seems @rmcgibbo is not comfortable to add `load_partial` function which I proposed in https://github.com/mdtraj/mdtraj/issues/1034, I decided to make `load_frame` function to support `slice` instance. So when `index` argument of `load_frame` is an instance of `slice`, it returns trajectories within the range specified by `slice`.

Note that I didn't touch docstrings while I'm not sure if this PR is acceptable. Let me know if this PR is acceptable then I'll modify docstrings (but sometime I feel that my English is not clear enough to explain things so I'd prefer you to write if possible...)
#### Usage

``` python
import mdtraj as md
t = md.load_frame('traj.h5', slice(10, 20, 2))
print(t)
# <mdtraj.Trajectory with 5 frames, 22 atoms>
```
#### Supported formats

Same formats as `load_frame` (single frame version) supports.

### DIFF
--- a/mdtraj/formats/arc.py
+++ b/mdtraj/formats/arc.py
@@ -82,8 +82,15 @@ def load_arc(filename, stride=None, atom_indices=None, frame=None):
 
     with ArcTrajectoryFile(filename) as f:
         if frame is not None:
-            f.seek(frame)
-            n_frames = 1
+            if isinstance(frame, slice):
+                start  = 0 if frame.start is None else frame.start
+                stop   = frame.stop
+                stride = stride if frame.step is None else frame.step
+                f.seek(start)
+                n_frames = stop - start
+            else:
+                f.seek(frame)
+                n_frames = 1
         else:
             n_frames = None
         return f.read_as_traj(n_frames=n_frames, stride=stride,
--- a/mdtraj/formats/binpos/binpos.pyx
+++ b/mdtraj/formats/binpos/binpos.pyx
@@ -122,8 +122,15 @@ def load_binpos(filename, top=None, stride=None, atom_indices=None, frame=None):
 
     with BINPOSTrajectoryFile(filename) as f:
         if frame is not None:
-            f.seek(frame)
-            n_frames = 1
+            if isinstance(frame, slice):
+                start  = 0 if frame.start is None else frame.start
+                stop   = frame.stop
+                stride = stride if frame.step is None else frame.step
+                f.seek(start)
+                n_frames = stop - start
+            else:
+                f.seek(frame)
+                n_frames = 1
         else:
             n_frames = None
 
--- a/mdtraj/formats/dcd/dcd.pyx
+++ b/mdtraj/formats/dcd/dcd.pyx
@@ -135,8 +135,15 @@ def load_dcd(filename, top=None, stride=None, atom_indices=None, frame=None):
 
     with DCDTrajectoryFile(filename) as f:
         if frame is not None:
-            f.seek(frame)
-            n_frames = 1
+            if isinstance(frame, slice):
+                start  = 0 if frame.start is None else frame.start
+                stop   = frame.stop
+                stride = stride if frame.step is None else frame.step
+                f.seek(start)
+                n_frames = int((stop - start) / stride)
+            else:
+                f.seek(frame)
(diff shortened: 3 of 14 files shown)


lambdalisue:
Note: Even this PR is not acceptable. I think https://github.com/lambdalisue/mdtraj/commit/1a54ad4c41b2ed6aabc496d7c9fc558b4e199d89 is a bug so at least that commit should be merged.

rmcgibbo:
I like this.

rmcgibbo:
I need to think about the API a little bit more though. It becomes a little bit redundant between `frame` and `stride` for the `load_XXX` functions now, since both can encode a stride/step.

lambdalisue:
Glad to hear ;-)

Yes. It is a bit redundant but it is required to keep backward compatibility...
Or maybe use `list` or `tuple` which indicate `(start, stop)` instead of using `slice` instance?

rmcgibbo:
Maybe the ideal thing would be to have keyword arguments that were `start`, `stop`, and `stride`, defaulting to `None` and mimicking the semantics of `range()`.

mpharrigan:
Neat!

I'm sure we'll be able to find an API that everyone is happy with. It seems a little strange that `load_frame` now might return multiple frames. What if we kept `load_frame` the same, and moved this to a function like `load_frames` (maybe with the start, stop, and stride parameters like @rmcgibbo suggested)

But besides mundane naming concerns, this would be great to have. And the bugfix is important too!

lambdalisue:
I think there are two topics. API for `load_frame(s)` and API for `load_XXX`.

So what about

```python
# If I don't need to keep the order of argument, `stride` should go the last like
#def load_XXX(filename, atom_indices=None, start=None, stop=None, stride=None):
def load_XXX(filename, stride=None, atom_indices=None, start=None, stop=None):
    # brabrabra

def load_frames(filename, start, stop=None, stride=None, top=None, atom_indices=None, **kwargs):
    # brabrabra
    return load_XXX(...)

# OR

def load_frame(filename, index_or_start, stop=None, stride=None, top=None, atom_indices=None, **kwargs):
    # brabrabra
    return load_XXX(...)
```

lambdalisue:
ping

rmcgibbo:
Sry my phd. defense is tomorrow so I can't look at this now.

swails:
Good luck! What's next?

lambdalisue:
Wow. Good luck :-]

rmcgibbo:
Starting at DESRES in nyc in about 2 weeks.

hainm:
:+1:

marscher:
Thats cool @rmcgibbo. All the best for your new job then!

lambdalisue:
... so what do I need to do rest?

cxhernandez:
@mpharrigan: bump

mpharrigan:
Thought @rmcgibbo was going to share his thoughts on this one.

I like the api suggested https://github.com/mdtraj/mdtraj/pull/1060#issuecomment-188662053. I agree it's unfortunate that we (probably) can't order the arguments (`start`, `stop`, `stride`) in the `load_xxx` functions. Maybe we can do a deprecation cycle during which we force people to specify stride or atom_indices with a keyword argument (although I don't know how to check if an argument is passed positionaly or with a keyword)

Do you have any thoughts @cxhernandez ?

mpharrigan:
To be clear, I like the signature where the first parameter is named `start` :)

cxhernandez:
I actually would prefer a scheme where you input a range or slice for `load_xxx` and consolidate 3 kwargs into one. It seems cleaner and would let users have more control over how they load frames (e.g. I could load trajectories using `np.logspace` this way).

Backwards compatibility would still be an issue though....

hainm:
I think just add `frame_indices` (like in `pytraj`)

```python
md.load(..., frame_indices=...)
```

it can be `range(0, 100, 10)` or random integer array or iterator.

rmcgibbo:
This seems outdated. Feel free to re-open if I'm mistaken.

Issue #1016, opened 2016-01-05
atom-names in MET residue


gph82:
Hi, perhaps this is a no-brainer (it's late...), but why are the hydrogens HB1 and HB2 being renamed to HB2 and HB3, respectively? 

Compare lines/atoms 10 & 11 between the mdtraj's output and a the command line interface ("head"). Lines 13 & 14 work as well
![snap](https://cloud.githubusercontent.com/assets/7518004/12123645/6a47a62a-b3e2-11e5-9061-f9c49cee9b44.png)

swails:
[Because that's how the PDB names them](http://www.rcsb.org/pdb/files/ligand/MET.cif) and OpenMM adheres to that standard as much as possible. And MDTraj uses (a modified copy of) OpenMM's PDB parser/structure.

gph82:
Hi, thanks very much, was not aware that that was the PDB standard. I familiarized with gromacs' topology files, e.g. amber99sb-ildn.ff/aminoacids.rtp:
```
[ MET ]
 [ atoms ]
     N    N           -0.41570     1
     H    H            0.27190     2
    CA    CT          -0.02370     3
    HA    H1           0.08800     4
    CB    CT           0.03420     5
   HB1    HC           0.02410     6
   HB2    HC           0.02410     7
    CG    CT           0.00180     8
   HG1    H1           0.04400     9
   HG2    H1           0.04400    10
    SD    S           -0.27370    11
    CE    CT          -0.05360    12
   HE1    H1           0.06840    13
   HE2    H1           0.06840    14
   HE3    H1           0.06840    15
     C    C            0.59730    16
     O    O           -0.56790    17
```

Been using mdtraj to work with gromacs for over 2y and only now I noticed this.

Good to know!

swails:
It seems that GROMACS is using some weird PDB v.2/PDB v.3 hybrid.  For instance, PDB v.2 used names `1HB` and `2HB`, whereas PDB v.3 uses `HB2` and `HB3`.

gph82:
I see. Well now that you mention, this has popped on the newer 5.xx (>4.6) versions of gromacs. I'll look more into it tomorrow. I'm hopping it's just a naming convention, which grompp can safely ignore the mismatched-naming warnings. Thanks

gph82:
[...]

PR #1061, opened 2016-02-29
[load_pdb] new optional argument standard_names

gph82:
standard_names = False will keep all atom names and residue names intact, even if these are not standard. Addresses #1016


mdtraj/formats/pdb/pdbfile.py
```diff
@@ -211,6 +215,10 @@ class PDBTrajectoryFile(object):
     force_overwrite : bool
         If opened in write mode, and a file by the name of `filename` already
         exists on disk, should we overwrite it?
+    standard_names : bool, default=True
+        If set to false, non-standard atomnames and residuenames will NOT be
```
rmcgibbo: Can you change the order of the clauses in this sentence? Start with 'If True, we standardize..., If False, this step is skipped.'

rmcgibbo:
LGTM. Thanks for adding this, @gph82!

gph82:
yeah, can do

Issue #1054, opened 2016-02-20
Implement doc changes from msmbuilder


mpharrigan:
- [x] default page is latest release, not development (use index.html from msmbuilder)
 - [x] re-write links in past docs to point to `development/`
 - [x] use travis built-in deployment
 - [x] use updated sphinx
 - [x] use updated notebook runner
 - [x] use updated msmb theme

mpharrigan:
In msmbuilder, I went through and changed all the "latest" links to "development", but then realized that going forwards we should just include "development" in `versions.json`. There are now *two* "development" options in the version selectors for old msmbuilder docs. So for this, I'm just going to *remove* the "latest/" link and put it into `versions.json`.

mpharrigan:
http://mdtraj.org/development/ is now live, although nothing links to it

mpharrigan:
I'll finish this shortly before 1.7

mpharrigan:
Done

PR #1067, opened 2016-03-11
Port doc building and deployment maintenence from msmbuilder

mpharrigan:
To change not in code

 - [ ] redirect mdtraj.org to latest stable release
 - [ ] move old links from `latest/` to `development/`

addresses #1054


mpharrigan:
@rmcgibbo you have a shiftx2 package in your anaconda. Is that recipe somewhere? Can we move to omnia?

rmcgibbo:
I don't think so. I bet the "recipe" was just copying the precompiled
binaries into a conda tarball

mpharrigan:
https://github.com/rmcgibbo/conda-shiftx2

PR #1072, opened 2016-03-17
[compute_contacs] new opt arg periodic=True

gph82:


rmcgibbo:
Can you put the new kwarg at the very end of the signature? Otherwise LGTM.

PR #1073, opened 2016-03-17
Chains

evanfeinberg:
This PR contains ability to (1) read and retain `chain.id` from trajectories with PDB topology file, (2) read and retain `chain.id` from `.h5` trajectories, and (3) retain `chain.id` info when saving a `.h5` trajectory to disk. It is a natural extension of #876


mpharrigan:
This doesn't include any commits other than those in #876

evanfeinberg:
Made PR from wrong branch. Will remake.

Issue #1070, opened 2016-03-14
Test cases for dssp are failing


peastman:
I get two test failures when I run test_dssp.py.  I need to figure out what's going on with that before I can start cleaning up the SSE code for them.

I notice that AppVeyor shows the test as passing, but only because mkdssp isn't installed, which means it skips the relevant tests.  And travis never even gets as far as running tests.  It fails with the message

```
Error:  Package missing in current linux-64 channels: 
  - shiftx2
```

In spite of that it still shows the build passing on the front page, which seems like a problem...

rmcgibbo:
Yes, this seems problematic. I think Matt was just working on the shiftx2
recipe on omnia, so this might be recent, or maybe not. Have you tried
running the tests locally though?

On Mon, Mar 14, 2016 at 6:47 PM, peastman <notifications@github.com> wrote:

> I get two test failures when I run test_dssp.py. I need to figure out
> what's going on with that before I can start cleaning up the SSE code for
> them.
>
> I notice that AppVeyor shows the test as passing, but only because mkdssp
> isn't installed, which means it skips the relevant tests. And travis never
> even gets as far as running tests. It fails with the message
>
> Error:  Package missing in current linux-64 channels:
>   - shiftx2
>
> In spite of that it still shows the build passing on the front page, which
> seems like a problem...
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mdtraj/mdtraj/issues/1070>.
>



-- 
-Robert

peastman:
Yes.  I get two failures in test_dssp.py.

rmcgibbo:
Got it. Probably something is wrong then. I would put it at 70/30 that the
test case itself is buggy, but it could also be the implementation.

On Mon, Mar 14, 2016 at 7:11 PM, peastman <notifications@github.com> wrote:

> Yes. I get two failures in test_dssp.py.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mdtraj/mdtraj/issues/1070#issuecomment-196559085>.
>



-- 
-Robert

mpharrigan:
I wad playing with it, but none of those changes have made it into master yet

peastman:
It looks to me like this is a real issue in the code.  The test in question calls `compute_dssp()`, then executes the `mkdssp` application, reads in the results, and compares them.  On just a small number of residues, mdtraj gives different assignments:

```
 32: 'T' 'T'
 33: 'T' 'T'
[...]

Issue #1066, opened 2016-03-10
Support for extreme box shapes


peastman:
I'm looking into rewriting the SSE based geometry code (see #1010).  I notice that the calculation for triclinic boxes is done in a very expensive way, many times slower than what's done in OpenMM (or in the neighborlist code I wrote for mdtraj).  The reason it's doing that is to support periodic boxes with extremely sharp angles.  But it still doesn't fully do that, because it assumes the nearest copy will be in one of the 27 adjacent boxes.  When the angles get extreme enough, that isn't guaranteed to be true.

So we have three options for how to do this calculation:

1. The method used in OpenMM.  It's fast and works for all typical boxes shapes, but can fail in extreme cases.

2. The method currently implemented in `compute_distances()`.  It's much much slower.  It works for some cases that method 1 fails on, but still fails on others.

3. Write something that's guaranteed to always find the correct result no matter the box shape, but might be even slower.

Which one of these do we want to do?  And if we pick either 1 or 2, we should add error checking to detect boxes that won't work correctly.

cxhernandez:
Would it be possible to implement something that does method 1 if the box is standard and falls back on method 3 if it's something more exotic?

peastman:
Yes, it's a possibility.

rmcgibbo:
As long as we detect whether the box shape is extreme and give the right
error message instead of an incorrect solution and no common MD codes
actually use these extremely skewed boxes, then I support getting rid of it.

swails:
What does OpenMM do?  I know there are fast methods you can use when running MD that won't work in the general case.  The original code (less the transposed inverse matrix) was sufficient for implementing MIC dynamics, but was only *guaranteed* to work for particles with at least one image within the maximum allowable cutoff.  Any particle whose closest image lies outside that range may, or may not, give the true minimum distance under those conditions.  That's why random atom-atom distances in a truncated octahedron was only correct ~75% of the time (and why I added code to loop over the 26 adjacent unit cells).  I *think* that is the code OpenMM uses, but @peastman can correct me if I'm wrong.

I agree with @rmcgibbo's assessment.

peastman:
OpenMM uses the method given by equations 3.13 to 3.16 in the Gromacs manual.  It uses just a single line to subtract off the appropriate multiple of each box vector.  And it *claims* that as long as the requirements on the box vectors (equations 3.1 to 3.3) are satisfied, that will always be correct.

[...]

Issue #1019, opened 2016-01-08
Triclinic Distance Bug


rmcgibbo:
I found another bug in `md.compute_distances` with triclinic unit cells. Unfortunately, I don't think #930 was quite correct in the general case.

```
import numpy as np
import mdtraj as md
from scipy.spatial.distance import euclidean

t = md.Trajectory(
    xyz=np.array([[0.5338, 1.5336, 0.9745], [1.4097, 1.9486, 1.1537]]),
    unitcell_lengths=np.array([[2., 2., 2.]]),
    unitcell_angles=np.array([[46.8, 34.4, 78.7]]),
    topology=None)


def simple_brute_force(v1, v2, a, b, c):
    d = float('infinity')
    for i in range(-2, 2):
        for j in range(-2, 2):
            for k in range(-2, 2):
                v2_p = v2 + (i * a + j * b + k * c)
                d = min(d, euclidean(v1, v2_p))
    return d


print('md.compute_distances (opt=True)', md.compute_distances(t, [[0, 1]], opt=True))
print('md.compute_distances (opt=False)', md.compute_distances(t, [[0, 1]], opt=False))
print('simple_brute_force',
      simple_brute_force(v1=t.xyz[0, 0],
                         v2=t.xyz[0, 1],
                         a=t.unitcell_vectors[0, 0],
                         b=t.unitcell_vectors[0, 1],
                         c=t.unitcell_vectors[0, 2]))
```

```
md.compute_distances (opt=True) [[ 0.74414033]]
md.compute_distances (opt=False) [[ 0.74414021]]
simple_brute_force 0.6186748147010803
```

swails:
This isn't an orthorhombic unit cell...

rmcgibbo:
Sorry, I got those words confused. I meant triclinic.

rmcgibbo:
I mean the complicated one.

swails:
Well one difference between your code and the code in geometry is that you loop over displacements -2, -1, 0, and 1.  Whereas [the code in mdtraj right now](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/geometry/distance.py#L239-L248) only loops over -1, 0, and 1.  Basically, the idea is that the "standard" MIC code (which works for MD because the distances it gets *wrong* must be too far away to lie within the maximum permissible cutoff) gets the particle *close enough* that you only have to test images in the 27 adjacent unit cells.

[...]

PR #1077, opened 2016-03-22
Rewrote SSE geometry code

### LINKED ISSUE #1010: Refactor geometry kernels
The core geometry functions (`compute_distances`, `compute_angles`, etc) are written in a style in C that is not very inviting. The preprocessor is used in a kind of dirty way to get both the periodic and non-periodic versions of the kernels compiled, and the explicit use of SSE intrinsics is a little challenging.

It should probably be refactored into a more sensible C++ using [OpenMM's](https://github.com/pandegroup/openmm/blob/eb1293f52263e49536cb2004799fe9cd293a0712/openmmapi/include/openmm/internal/vectorize_sse.h) sse wrappers to simplify the visual look of all of the SSE code.

peastman:
I've rewritten all the optimized SSE based code, replacing the direct calls to intrinsics with wrapper classes.  This fixes #1010.  The main goal of this was to make the code easier to understand and maintain, but in the process I fixed several bugs and made other improvements:

- There was a bug in the implementation of dssp that caused it to sometimes misclassify residues.  See #1070.

- For triclinic boxes, `compute_distances()` and `compute_displacements()` would sometimes identify the wrong periodic copy (not the closest one).  See #1066.  I've improved that.  I'm still not 100% certain it is always guaranteed to find the closest copy, but it's certainly better than it was.  This may also address #1019, but I haven't tested that.

- If the box vectors changed over the trajectory, `compute_neighbors()` would use the vectors from the first frame for every frame.  I fixed that.

- `compute_angles()` and `compute_dihedrals()` didn't have optimized code for triclinic boxes, so in that case they would fall back to the very slow Python version.  The SSE code for them now supports triclinic boxes.


mpharrigan:
Wow! Nice work

mdtraj/geometry/src/geometry.cpp
```diff
@@ -0,0 +1,313 @@
...
+ * not. Most of the code between these versions is shared, so I've written
+ * the parts of the two functions which are different using #ifdefs. So we
+ * just include these files _twice_ here, toggling the variable that controls
+ * the ifdef.
+ *
+ * Note that these kernel files are really not capable of being compiled
+ * independently -- they're not header files at all -- and they're really just
+ * meant to be #included here.
+ **/
+#undef COMPILE_WITH_PERIODIC_BOUNDARY_CONDITIONS
```
marscher: Nice hack. Does this increase maintainability as you stated?

marscher:
This is major improvement! Thank you

mdtraj/geometry/src/geometry.cpp
peastman: It's not my hack!  I just copied that directly from the old geometry.c file.  So it still works the same way as before, just with different code in the included files.  (Personally, I don't find it all that hacky, but whatever.)

swails: > (Personally, I don't find it all that hacky, but whatever.)

I remember thinking exactly the same thing when I saw it in `geometry.c` :smiley:.  This is a fairly common technique I've seen used in many places including several codes in Amber as well as OpenMM (in a sense, anyway).

rmcgibbo: I did write it, but I guess I'm the only one who finds it ugly. :)

rmcgibbo:
Big +1 from me. Thanks, @peastman!

Issue #1078, opened 2016-03-22
Possible regression with len(xtc_file)


mpharrigan:
I'm noticing some issues with the new xtc code #1038 

1. Just getting the length is slower than it used to be. It spends some time making the index rather than just telling you the length
2. Seems to fail on small number of frames

I'm going to write little test cases for both of these, because they're just conjecture at this point

mpharrigan:
```python
import mdtraj as md
import numpy as np
from subprocess import call

with md.formats.XTCTrajectoryFile('smalltraj.xtc', 'w') as xtc:
    xtc.write(np.random.uniform(size=(5, 100000, 3)))

call(['ls', '-lh', 'smalltraj.xtc'])

f = md.open('smalltraj.xtc')
print(len(f))
```

```
$ python smallfile.py 
-rw-r--r-- 1 harrigan harrigan 1.9M Mar 22 14:51 smalltraj.xtc
Traceback (most recent call last):
  File "smallfile.py", line 11, in <module>
    print(len(f))
  File "mdtraj/formats/xtc/xtc.pyx", line 698, in xtc.XTCTrajectoryFile.__len__ (mdtraj/formats/xtc/xtc.c:10331)
  File "mdtraj/formats/xtc/xtc.pyx", line 675, in xtc.XTCTrajectoryFile.offsets (mdtraj/formats/xtc/xtc.c:9884)
  File "mdtraj/formats/xtc/xtc.pyx", line 649, in xtc.XTCTrajectoryFile._calc_len_and_offsets (mdtraj/formats/xtc/xtc.c:9560)
IndexError: Out of bounds on buffer access (axis 0)
```

mpharrigan:
I can't reproduce (1) in a test environment. I suspect something somewhere is being clever and caching the files

marscher:
Well maybe just one line of code needs to be adopted to really fix this. Please hold on

marscher:
Just a comment to (1). The old code used to interpret all atom coordinates for every frame just to obtain the length, however the new code just reads one integer from a frame ignoring how many atoms there are. So do you really believe this could be any slower?

marscher:
The error is caused by a simple rounding error during increasing the size of the offsets array. I'll prepare a PR.

mpharrigan:
Ok, your comments on (1) have sold me on that as well

PR #1079, opened 2016-03-22
Half-fix short xtc file issues

mpharrigan:
see #1078 

This puts `__len__` back to the old method, which works for trajectories with a small number of frames. It's probably faster because it doesn't allocate the offsets array

I added a test (the test from #1078) that fails. @marscher : can you build off this pull request and try to fix that problem?

### DIFF
--- a/mdtraj/tests/test_xtc.py
+++ b/mdtraj/tests/test_xtc.py
@@ -281,8 +281,14 @@ def test_ragged_2():
         assert_raises(ValueError, lambda: f.write(xyz))
 
 
-def test_short_traj():
+def test_short_traj_len():
     with XTCTrajectoryFile(temp, 'w') as f:
         f.write(np.random.uniform(size=(5,100000,3)))
     with XTCTrajectoryFile(temp, 'r') as f:
         assert len(f) == 5, len(f)
+
+def test_short_traj_offset():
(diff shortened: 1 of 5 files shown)


mdtraj/tests/test_xtc.py
```diff
@@ -279,3 +279,16 @@ def test_ragged_2():
     with XTCTrajectoryFile(temp, 'w', force_overwrite=True) as f:
         f.write(xyz, time=time, box=box)
         assert_raises(ValueError, lambda: f.write(xyz))
+
+
+def test_short_traj_len():
+    with XTCTrajectoryFile(temp, 'w') as f:
+        f.write(np.random.uniform(size=(5,100000,3)))
+    with XTCTrajectoryFile(temp, 'r') as f:
+        assert len(f) == 5, len(f)
```
marscher: len triggers offset calculation, so in fact the second test is equivalent.

marscher:
for a proper fix merge #1080

mdtraj/tests/test_xtc.py
mpharrigan: In this PR, I changed __len__ to not trigger the offset calculation, see changes. but yeah, with the unchanged code, it would

mpharrigan:
Closed in favor of #1080

PR #1080, opened 2016-03-22
[xtc] fix a bug in calc of offsets array for short trajs.

### LINKED ISSUE #1078: Possible regression with len(xtc_file)

marscher:
Fixes #1078. For short trajs there is a rounding error (forced cast
to integer) to the next smallest integral. Passed new_size to ceil.


mpharrigan:
Thanks!

PR #1085, opened 2016-04-06
[xtc] fix error status report during xdr_seek

marscher:
I consider this a minor issue, since the program crashes in this case any ways. Now it prints a proper message instead of "String format error".


mpharrigan:
Thanks!

PR #1087, opened 2016-04-11
[WIP] Remove all traces of broken segment_id code

jchodera:


rmcgibbo:
I would prefer not to merge this.

jchodera:
Alternatives?  https://github.com/mdtraj/mdtraj/commit/c4d43469705f395e21d2a6dc789223f4e0bbc6d3 committed a bunch of broken code. Completely excising it and having someone else start again with working code might be a good alternative to continuing to distribute code that simply isn't compatible with OpenMM.

rmcgibbo:
It isn't completely broken. There is a single bug which can be fixed properly.

jchodera:
Unless you'd like to identify the

PR #1088, opened 2016-04-12
[WIP] OpenMM does not yet support segment_id

jchodera:
Attempting to address #1047.

Will add tests in a moment.

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -31,6 +31,7 @@ test:
     - ipywidgets
     - matplotlib
     - pandas
+    - openmm
     - shiftx2 # [not win]
 
   commands:
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -369,7 +369,8 @@ class Topology(object):
         for chain in value.chains():
             c = out.add_chain()
             for residue in chain.residues():
-                r = out.add_residue(residue.name, c, residue.segment_id)
+                segment_id = "" # OpenMM does not yet support segment_id
+                r = out.add_residue(residue.name, c, segment_id)
                 for atom in residue.atoms():
                     if atom.element is None:
                         element = elem.virtual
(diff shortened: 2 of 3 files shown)


jchodera:
It looks like there *is* a test for `Topology.from__openmm()`, but it may not have been run because the [nosetests don't install the `openmm` package](https://github.com/mdtraj/mdtraj/blob/master/devtools/conda-recipe/meta.yaml#L25-L34).

Is it OK if I add `openmm` in `test: requires:`?

jchodera:
Hm, appveyor seems to be broken:
```
error: command 'C:\\Users\\appveyor\\AppData\\Local\\Programs\\Common\\Microsoft\\Visual C++ for Python\\9.0\\VC\\Bin\\cl.exe' failed with exit status 2 
```

jchodera:
I've added `openmm` to the nosetests so this doesn't break again.

jchodera:
There is some code in `MDTraj` that is apparently supposed to work on either MDTraj or OpenMM `Topology` objects even though they are fundamentally incompatible. I think this is a deeper issue, but I am applying a band-aid for now.

The HDF5 failure is more serious.  There is a cast of `residue.resSeq`---which is a *string* for OpenMM---to an `int`, which is fundamentally not allowed. I don't know how to fix that without rearchitecting your HDF5 data model.

mpharrigan:
Thanks @jchodera 

Looks like the addition of openmm uncovered some new test failures. Upon first glance it looks like the reporter objects are using openmm topology and assuming it's the same as an mdtraj topology

mpharrigan:
Subsumes #1048

jchodera:
Well, there's code in there that really states that is a fundamental assumption:
```
Note
    ----
    This really should be a copy constructor (class method) on Topology,
    but I want it to work on either the mdtraj topology OR the OpenMM
    topology. An inplace version for the topology object we have here
    is also available.
```

mpharrigan:
Could you point me to that note? Can it use the from_openmm method you just fixed?

jchodera:
I still think you should rip out https://github.com/mdtraj/mdtraj/commit/c4d43469705f395e21d2a6dc789223f4e0bbc6d3 wholesale.  It wasn't a mostly-working PR with a bug or two. It fundamentally violates deep assumptions made within MDTraj.

jchodera:
Here you go!

https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/topology.py#L69-L94

If you figure out there are more minor changes that would actually make things work, let me know---I'm happy to make them.

PR #1089, opened 2016-04-12
Fix segment_id issues take 3

mpharrigan:
Based off of the first two commits of #1088 

@jchodera let me know what you think

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -31,6 +31,7 @@ test:
     - ipywidgets
     - matplotlib
     - pandas
+    - openmm
     - shiftx2 # [not win]
 
   commands:
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -369,7 +369,8 @@ class Topology(object):
         for chain in value.chains():
             c = out.add_chain()
             for residue in chain.residues():
-                r = out.add_residue(residue.name, c, residue.segment_id)
+                segment_id = "" # OpenMM does not yet support segment_id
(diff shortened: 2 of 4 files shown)


jchodera:
Thanks!

In cases like this, is it possible to refer to which `Topology` object (mdtraj or openmm) the function expects? Now that they are inequivalent, it's often ambiguous as to which one(s) is(are) supported:
```python
     Parameters
      ----------
     topology : topology
         The base topology
```

mpharrigan:
The assumption is mdtraj.Topology. If you find cases where it expects something different, feel free to update the docstrings

mpharrigan:
Looks like travis failure is unrelated

jchodera:
I think you can just retrigger the build.

jchodera:
Other parts of the code in the same file are very specific:
```
         Parameters
         ----------
         traj : MDTraj.Trajectory, optional, default=None
             If specified, use the first frame from this trajectory to
             set the unitcell information in the openmm topology.
```
Being vague in that particular function seems inconsistent with the rest of the documentation.

mpharrigan:
Which function?

jchodera:
Here: https://github.com/mdtraj/mdtraj/pull/1089/files#diff-ce95bbb0db9dbb5dce66ba686ed67f2dR80

This one also used to be both `OpenMM` and `MDTraj` `Topology` compatible but now seems only `MDTraj`: https://github.com/mdtraj/mdtraj/pull/1089/files#diff-ce95bbb0db9dbb5dce66ba686ed67f2dR214

There are probably a few others that are now only one or the other due to the divergence.

rmcgibbo:
This lgtm

PR #1090, opened 2016-04-12
Test against latest numpy

mpharrigan:


mpharrigan:
I'll merge this and we can fix windows in subsequent prs

mpharrigan:
Actually, the relevant numpy 1.11 packages don't exist for windows. I'll revert this in another pr

PR #1100, opened 2016-04-24
Update geometry.cpp for Windows

rmcgibbo:
This file doesn't currently compile on our windows CI.


mpharrigan:
Nice. looks like there's still a problem with cbrt

Issue #1084, opened 2016-04-05
image_molecules IndexError


jadeshi:
Wondering if anyone else has come across this error when trying to use image_molecules. The trajectory is an xtc for which I supplied a pdb topology with PBC's, and loads into mdtraj fine.

```
IndexError                                Traceback (most recent call last)
<ipython-input-46-7ee548d6ab04> in <module>()
----> 1 traj.image_molecules()

/home/jadeshi/miniconda/lib/python2.7/site-packages/mdtraj-1.7.0.dev0-py2.7-linux-x86_64.egg/mdtraj/core/trajectory.pyc in image_molecules(self, inplace)
   1901             used_anchors = [0]
   1902             available_anchors = list(range(1, num_anchors))
-> 1903             min_anchor_dist = anchor_dist[0, :]
   1904 
   1905             # Add in anchors one at a time, always taking the one that is nearest an existing anchor.

IndexError: index 0 is out of bounds for axis 0 with size 0
```

Thanks!

mpharrigan:
I'm seeing this as well, incidentally on the system I used to originally test the code. @peastman do you have any ideas?

peastman:
Does that mean this used to work, and then recently stopped working?  If someone can provide me with a file that shows the problem, I'll look into it.

mpharrigan:
@jadeshi I could reproduce this when using a gro topology file, but not with pdb. From ipython, can you enter into pdb (python debugging) by entering `%pdb on` (so it will enter the debugger when an exception happens) and at the pdb prompt, type and report the values of

`atoms_cutoff`
`num_anchors`

peastman:
Can you send me the files to reproduce this?

mpharrigan:
I sent you mine, but I suspect @jadeshi 's issue might be different if he is using a pdb as he said

peastman:
It looks like there are actually two problems.  One is clearly a bug.  The other isn't exactly a bug, but it could still possibly be handled better.

The first problem is that if it doesn't identify any anchor molecules, it throws this exception.  That will happen, for example, if every molecule in the topology is exactly the same size.  So if you called this function for a box of water, it would fail.  I can fix that.

[...]

PR #1102, opened 2016-04-26
Give more helpful error message when image_molecules fails

mpharrigan:
I ran into #1084 when using a `.gro` file as my topology. It looks like it was putting every atom in its own "molecule" and thereby finding no anchor molecules. 

This error message will have to do for now, since there's pressure to cut a new release

### DIFF
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -1863,7 +1863,7 @@ class Trajectory(object):
         num_anchors = len(anchor_molecules)
         if num_anchors == 0:
             raise ValueError("Could not find any anchor molecules. Based on "
-                             "out heuristic, those should be molecules with "
+                             "our heuristic, those should be molecules with "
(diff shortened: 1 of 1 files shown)


mdtraj/core/trajectory.py
```diff
@@ -1861,6 +1861,12 @@ def image_molecules(self, inplace=False):
         anchor_molecules = [m for m in molecules if len(m) > atoms_cutoff]
         other_molecules = [m for m in molecules if len(m) <= atoms_cutoff]
         num_anchors = len(anchor_molecules)
+        if num_anchors == 0:
+            raise ValueError("Could not find any anchor molecules. Based on "
+                             "out heuristic, those should be molecules with "
```
rmcgibbo: "out" -> "our"?

peastman:
Let me fix this properly.  It's very easy to do.

PR #1101, opened 2016-04-24
Fix geometry compilation for Windows

mpharrigan:
Supersedes #1100 
cc #1077 
fixes #1092 

I removed the "static" option because it was giving me compilation errors. something about dll's. Now it still gives a warning `mdtraj/geometry/src/geometry.cpp(31): warning C4273: 'roundf': inconsistent dll linkage`

Issue #1097, opened 2016-04-19
Compilation error on Mac OSX


tcmoore3:
I'm having issues building MDTraj from source. I'm running OS X 10.9.5. My `clang` version is:
 ```bash
(master) $ clang -v
Apple LLVM version 6.0 (clang-600.0.57) (based on LLVM 3.5svn)
Target: x86_64-apple-darwin13.4.0
Thread model: posix
```

The error I'm getting seems to be a namespace error:
```bash
mdtraj/geometry/src/geometry.cpp:227:9: error: use of undeclared identifier 'isnan'; did you mean
      'std::isnan'?
    if (isnan(existing_e0) || e < existing_e0) {
        ^~~~~
        std::isnan
/usr/include/c++/4.2.1/cmath:551:5: note: 'std::isnan' declared here
    isnan(_Tp __f) { return ::__gnu_cxx::__capture_isnan(__f); }
    ^
mdtraj/geometry/src/geometry.cpp:234:14: error: use of undeclared identifier 'isnan'; did you mean
      'std::isnan'?
    else if (isnan(existing_e1) || e < henergies[2*donor+1]) {
             ^~~~~
             std::isnan
/usr/include/c++/4.2.1/cmath:551:5: note: 'std::isnan' declared here
    isnan(_Tp __f) { return ::__gnu_cxx::__capture_isnan(__f); }
    ^
2 errors generated.
error: command 'clang' failed with exit status 1
```

It of course compiles when I change `isnan` to `std::isnan`, but that seems like a workaround rather than a solution.

marscher:
You're likely mixing clang with libstdc++, which leads to strange errors 
like the one you just observed.

/usr/include/c++/4.2.1 maybe is a left-over from a custom installation?

PR #1104, opened 2016-04-27
Fixed a compilation error

peastman:


mpharrigan:
@peastman : did you see #1101 ?

1. Can you go over the changes and make sure I didn't do anything dumb
2. I suspect this PR will break python 2.7 on windows, which doesn't have std::nan (?)

peastman:
`std::isnan()` is completely standard.  It's found in `cmath`:

http://en.cppreference.com/w/cpp/numeric/math/isnan

So Windows shouldn't have any problem with it, and it doesn't matter what Python version you're using.

mpharrigan:
Are you sure it works with visual studio 2008? The python version dictates the visual studio version

peastman:
Do you have reason to think it doesn't work?

If in doubt, `isnan(x)` can be replaced with `x != x`.

mpharrigan:
Yes. It wasn't compiling on appveyor. Please see #1100 and #1101 where @rmcgibbo and I implemented the nan check you describe above

Your link above also describes it as a c++11 feature.

Did you observe a compilation failure without this pr (i.e. without the namespace prepended)? It compiles on travis and on my machines (with gcc)

peastman:
If the issue is already being fixed elsewhere, I can close this.

mpharrigan:
Did you actually observe a compilation failure with the `std::`s omitted? Or can this be closed?

ctk3b:
Related to #1097 ?

peastman:
Yes.  When I pulled the latest changes a few days ago, it started failing to compile.  This change made it compile again.  It looks like the same problem as #1097.

mpharrigan:
Is it a clang thing?

mpharrigan:
I suspect we can fix it without breaking the windows bodge by putting `using std::isnan` at the top

mpharrigan:
We should also add travis mac build to the matrix

rmcgibbo:
> We should also add travis mac build to the matrix

Good idea. That should be relatively simple, although some of the external test dependencies like SPARTA+ and stuff might be more problematic.

peastman:
Or just define our own `isnan()` that checks for `x != x`.

marscher:
Am 29.04.2016 um 19:15 schrieb peastman:
> Or just define our own |isnan()| that checks for |x != x|.
x != x can be optimized away by some super clever compilers.

peastman:
Generally those optimizations have to be explicitly enabled.  For example, gcc has a `-ffinite-math-only` flag which tells the compiler it can assume numbers will never be inf or nan.  But that isn't included in any of the `-0#` optimization levels.  It's usually safe to assume math will be IEEE compliant unless you've specifically told the compiler it doesn't have to be.

marscher:
I'm pretty sure gcc and clang obey this. But can we ever be sure about M$ compilers?

peastman:
See https://msdn.microsoft.com/en-us/library/e7s85ffb.aspx.  The default behavior is `/fp:precise`, for which is specifies:

> The compiler correctly handles comparisons that involve NaN. For example, x != x evaluates to true if x is NaN and ordered comparisons involving NaN raise an exception.

marscher:
Ah thanks, didn't know that.

mpharrigan:
It would be really annoying if microsoft didn't provide `isnan` and also optimized away `x != x`

Issue #1086, opened 2016-04-06
Custom anchor_molecules for traj.image_molecules


jchodera:
Is there a way to use MDTraj to reimage a trajectory inside a periodic box so that all molecules are wrapped to be the minimum-image distance from a given atom selection?

mpharrigan:
You need the development release as it was recently added, but yes

http://mdtraj.org/development/api/generated/mdtraj.Trajectory.html#mdtraj.Trajectory.image_molecules

jchodera:
Thanks!  

From the docs:
>  It tries to be intelligent in deciding what molecules to center, so you can simply call it and trust that it will “do the right thing”.

Is there a way to specify what molecule should be at the "center" of the box via an optional selection argument?

mpharrigan:
Not currently, but PRs welcome

mpharrigan:
(It should be straightforward)

jchodera:
How about we just expose `anchor_molecules` as an optional argument in [`image_molecules`](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/trajectory.py#L1823)?  If specified by the user, this is used instead of the automatically-selected one.

mpharrigan:
Yup! That's what I was thinking

mpharrigan:
and maybe refactor the guessing logic into its own function

PR #1106, opened 2016-04-28
image_molecules in cython

mpharrigan:
On 11 frame of 82k atoms

before: 25.0 s
after: 2.8 s

I'm going to rebase and tidy a bit. Specifically, I don't think compiling in `geometry.cpp` is necessary here


mpharrigan:
Ok, I just moved this code to be part of `mdtraj.geometry._geometry`

mpharrigan:
This also addresses #1086

mpharrigan:
- [ ] docstring update
 - [ ] move `guess_anchor_molecules` to `Topology`
 - [ ] add a couple tests

jchodera:
Is this still in progress?

mpharrigan:
Yes, I'd still like to do the unchecked things above

jchodera:
What about grabbing the tests from #1183 and punting on the refactoring to move `guess_anchor_molecules` to a separate PR?

jchodera:
Any thoughts on that idea, @mpharrigan?

mpharrigan:
I'd rather just move the function now so we don't introduce an api we're immediately going to break. Do you want to make those changes off of this branch?

jchodera:
> I'd rather just move the function now so we don't introduce an api we're immediately going to break. Do you want to make those changes off of this branch?

Guessing the anchor molecules is still an internal-only function, right? There's still no API exposure.

> Do you want to make those changes off of this branch?

Would you want me to bring in my optional argument support (to let the user specify anchor molecules) and tests from #1183 into here?

In thinking about why we still have a bound peptide bouncing all over the place, I realized that the [current behavior with multiple anchors](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/trajectory.py#L1908-L1964) is probably actually what we need---to image the peptide to minimize the contact distance---so I'm not sure why it isn't working (unless it's somehow not picking up the peptide as an anchor).

mpharrigan:
>Guessing the anchor molecules is still an internal-only function, right? There's still no API exposure.
Nope! as currently written, it's not prefaced with an underscore. I'd imagine users would want to use this function to (a) see what's being guessed as an anchor molecule, (b) start with the guess, tweak it a bit, and feed into imaging function

>Would you want me to bring in my optional argument support (to let the user specify anchor molecules)

I think this already includes this change. What arguments in particular?

jchodera:
> I think this already includes this change. What arguments in particular?

You're right!

> edit: tests from #1183 appreciated!

OK, will create a PR to your repo branch with just the tests.

jchodera:
Any chance you could merge from master or rebase first?

mpharrigan:
Rebased

jchodera:
Thanks! PR created: https://github.com/mpharrigan/mdtraj/pull/5

jchodera:
Looks like the appveyor error is just an HTTP timeout:
```
Fetching package metadata: .....Error: HTTPError: 504 Server Error: GATEWAY_TIMEOUT: https://repo.continuum.io/pkgs/pro/win-32/
```

mpharrigan:
thanks @jchodera !

PR #1109, opened 2016-04-29
Run macosx tests

mpharrigan:


mpharrigan:
cc #1104 

The compilation error shows up clear as day. I can't replicate the build error on my linux system when using clang

mpharrigan:
Switching from `<cmath>` to `<math.h>` fixed the compilation error

tcmoore3:
What was wrong with using ```std::isnan```

mpharrigan:
visual studio 2008, needed for python 2.7 for abi compatibility, does not have `std::isnan`. We get around it by defining a macro for `isnan` in that case

PR #1118, opened 2016-05-14
[basesetup] copy contents of export_include and static lib to src tree.

### LINKED ISSUE #1117: [basesetup] StaticLib outputdir not correct for develop install mode
For libtheobald we need to point the output directory to the actual source tree instead of the build directory for mdtraj.capi to work. Currently the headers and static libs are being stored in the temporary build directory, so they can not be included properly if one only has a development version of mdtraj installed.

marscher:
Fixes #1117

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,6 @@ nosetests.xml
 
 .cache
 build.log
+
+# export of libtheolib
+/mdtraj/core/lib/*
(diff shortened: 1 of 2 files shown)


rmcgibbo:
Do you have any idea if the windows failure is related to this PR?

marscher:
No it isn't. It fails with some distance calculation error.

rmcgibbo:
Okay, let's keep an eye out for that one and see if it's recurring or stochastic.

rmcgibbo:
(Thanks for the fix!)

PR #1130, opened 2016-06-13
Automatically test examples to make sure they work

### LINKED ISSUE #1122: The examples and documentation seem outdated
For example, http://mdtraj.org/latest/viewer.html refers to `from mdtraj.html import TrajectoryView, enable_notebook`, but this fails:

```python

>>> import mdtraj
>>> from mdtraj.html import TrajectoryView, enable_notebook
ImportError: cannot import name TrajectoryView
```

peastman:
Fixes #1122.


peastman:
It looks like things are working nicely on travis, but appveyor reports lots of failures.  From looking at the log, I don't really understand what's causing them.  There are lots of messages about undefined symbols, and the error message "Unable to open/create file 'ala2.h5'" appears lots of times, and in one place there's an error complaining that OpenMM isn't installed.  But I can't tell what the root cause is.

mpharrigan:
guess: ala2.h5 is a symlink which don't exist on windows. problems ensue

peastman:
You're right.  I didn't know git could preserve symlinks like that.  There are two of them in the examples folder.  Shall I just copy over the real files instead?  That one is only 32K, and the other one is 145K, so it's not like we're saving a lot of space.

mpharrigan:
Sounds good to me

peastman:
Now we're down to a much smaller list of errors.  simulation-with-openmm.ipynb is failing because OpenMM isn't installed.  In the test dependencies in meta.yaml it specifies

    - openmm  # [not (win32 or (win and py2k))]

What's the reason for not installing it on Windows?

WebGL-Viewer.ipynb is also failing with an error that I don't really understand:

```
  File "C:\Miniconda3\envs\_test\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\Miniconda3\conda-bld\work\examples\test_examples.py", line 19, in check_one_notebook
    run_notebook(nb)
  File "C:\Miniconda3\conda-bld\work\examples\test_examples.py", line 37, in run_notebook
    reply = shell.get_msg(timeout=20)['content']
  File "C:\Miniconda3\envs\_test\lib\site-packages\jupyter_client\blocking\channels.py", line 57, in get_msg
    raise Empty
queue.Empty: 
```

mpharrigan:
Looks like we only build openmm conda packages for 64bit python3

https://anaconda.org/omnia/openmm/files

peastman:
How do we want to handle that then?  If there are examples that aren't expected to work on certain platforms?

mpharrigan:
Can we include logic to skip certain notebook tests? Maybe have a cell at the top of the notebook
```python
try:
  import openmm
except ImportError:
  print("This requires openmm!")
  ipython.equivalent_of_sys.exit()
```

peastman:
Ok, I think this is ready to merge.  I used the easy solution of only testing the examples on platforms that support OpenMM.  The other error was just a timeout that was set a little too short.

Issue #1129, opened 2016-06-12
PDB parsing issues


psolsson:
Parsing PDB-files where residues have atoms in a not strictly contiguous stretch gives rise to multiple residues. examples:
loading this one in mdtraj 1.7.2
[system-protein.pdb.zip](https://github.com/mdtraj/mdtraj/files/310935/system-protein.pdb.zip)
is understood as a topology with 78 residues; whereas it has only 39. 

Changing the atom ordering fixes the problem. As far as i can see there is no strict rules about the ordering of atoms according to the PDB format specification - the pdb parses well in other programs, vmd, pymol etc.

rmcgibbo:
Thanks for the bug report and the example file.

cclementi:
Hi, I would like to add to this issue, as I am having the same problem with the same file. 
This pdb file is created by vmd from a .mae file that is provided with the D.E. Shaw trajectories of the proteins simulated in their paper: Kresten Lindorff-Larsen, Stefano Piana, Ron O. Dror, and David E. Shaw, "How Fast-Folding Proteins Fold," Science, 334, 2011, 517-520.  All proteins from this set seem to have this issue. If the pdb file straight out of vmd is used as topology file in mdtraj to load the *.dcd files, it is understood as a longer protein because the atoms are not listed in the correct residue-after-residue order (all the H atoms are at the end). As Simon pointed out, other programs (like pymol) are able to interpret this pdb correctly. If the pdb is loaded and saved by pymol in a format that is mdtraj-friendly (with the atoms in the correct order), then it's not consistent with the corresponding *.dcd files anymore and if the *.dcd files are loaded in mdtraj the molecule appears all "scrambled".  
I hope I haven't confused things more. I can provide the files if needed.
Thanks!

rmcgibbo:
Yes, this sounds like a problem. I think other users like @evanfeinberg may have come up with solutions. Unfortunately I don't really have time to fix this, but maybe other developers do. If either of you want to take a stab at fixing it in a pull request, that would be great too.

rmcgibbo:
One thing which may be worth checking: the PDB-parsing code in mdtraj mostly originates from OpenMM, but it might be slightly outdated with respect to any bugfixes that have been recently added to OpenMM. So it may be worth checking if this issue exists in OpenMM, and it doesn't, there maybe be a very simple patch that could be applied.

evanfeinberg:
[...]

PR #1135, opened 2016-06-28
Fix parsing errors for pdb and psf files with nonsequential atoms

Eigenstate:
This should fix issue #1129 , and similar issues parsing psf files.

The origins of the out of order residues is actually from Maestro's add hydrogens, which puts them at the end of the mae file. Converting tends to be done in atom order, so these hydrogens stay at the end of the psf file and away from the rest of the residue.

The parsing functions for pdb and psf were defining a new residue every time they encountered a change in resid. Now, they check if the resid of the current atom has already been defined, and if so add atoms to that residue. Otherwise it behaves as normal and creates a new residue.

I tested this on my out of order pdb and psf files from Anton trajectories.


cxhernandez:
@mdtraj/developers: should we add the topology from #1129 as a test file and create a test for this bug?

peastman:
We need to be a bit careful about this.  There are so many nonstandard PDB files out there, and they can be nonstandard in so many ways, it's really easy to break the reading of some files when you try to fix others.

PDB files only have four columns for the residue ID, which means a standard PDB file can only have up to 9999 residues.  For more residues, you're officially supposed to split it into multiple files, but hardly anyone does that.  Instead they use various tricks (intentionally or accidentally) to allow more.  Some that I've seen include:

- Writing the residue ID mod 10000.
- Spilling over into the next column (which is supposed to be for the insertion code).
- Just writing **** as the residue ID.

This PR will fail if it gets a file that does any of those, because it will encounter a new residue that has the same value in the residue ID field as a previous one.

cxhernandez:
> This PR will fail if it gets a file that does any of those because it will encounter a new residue that has the same value in the residue ID field as a previous one.

Do we want to encourage non-standard PDBs like that in cases where they should clearly be using PDBx/mmCIF or something more modern? Also, do we want to sacrifice an increasing number of Desmond/Maestro users because people are using the incorrect format to store their topologies?

cc @mpharrigan @kmckiern, who actually work on large systems, for their input

peastman:
Files of this sort are _really_ common.  To put it into perspective, consider that any PDB file with more than 9999 residues or 99,999 atoms is _by definition_ nonstandard.  It has to do something to work around the limited number of columns, and chances are high that "something" will cause this PR to behave incorrectly.  That includes, among other things, files written by mdtraj itself!  It writes the residue ID mod 10,000:

https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/pdb/pdbfile.py#L349

mpharrigan:
Yeah I think it's more common to have > 10000 residues than using a maestro pdb

Eigenstate:
That makes sense. How about just taking the fix for parsing psf files and then having Maestro/Desmond users provide a psf?

mpharrigan:
That sounds good to me. Maybe we could add a flag for pdb parsing

PR #1136, opened 2016-06-30
Added insert_atom function

leeping:
Hi Robert, it's been a while.  I've been using mdtraj fairly regularly, and I found that a function that can insert atoms into a trajectory at a specific position is often helpful (rather than adding it to the end).  The following function should do this, and I think the interface is consistent with your existing `add_atom` function.  This PR doesn't override `add_atom` because I think people may be using it already without explicitly naming the `serial` keyword argument.

Let me know what you think is the best way to test.


mdtraj/core/topology.py
```diff
@@ -607,6 +607,54 @@ def add_residue(self, name, chain, resSeq=None, segment_id=""):
...
+            atom = Atom(name, element, self._numAtoms, residue, serial=serial)
+            self._atoms.append(atom)
+        else:
+            atom = Atom(name, element, index, residue, serial=serial)
+            for i in range(index, len(self._atoms)):
+                self._atoms[i].index += 1
+            self._atoms.insert(index, atom)
+            for i, a in enumerate(self._atoms):
+                if i != a.index:
+                    raise RuntimeError('Spoo!')
```
ctk3b: Should probably be a unit test or alternatively display a more informative error message :)

leeping: Ah, sorry, that was left in during development. :) I'll update it.

leeping:
I also added a function for deleting atoms.  I added these functions while writing a script to swap out the water molecules in a trajectory, which sometimes requires adding or removing virtual sites (for example, going from TIP3P to TIP4P-Ew and vice versa).  

Perhaps a good unit test would be to add the virtual sites, check the number of atoms matches up with our expectations, and then make sure that deleting these atoms brings us back to the original. What do you think?

Thanks,

- Lee-Ping

ctk3b:
That sounds very reasonable to me.

And +1 from me in general for this addition.

rmcgibbo:
Better late than never! I just merged this :)

PR #1137, opened 2016-07-01
Allow standard_names as a keyword argument to md.load()

leeping:
This PR allows `standard_names` as a keyword argument to `md.load()`, which I think is the normal expected behavior.  I also noticed that  `standard_names` is not an input argument to some other parsers of topology-containing file types, such as GROMACS `.gro`.  It may be good to make some of these changes in the future.


gph82:
Hi @leeping , I did not think of that when committing #1061, but yes, that would be a good idea.

PR #1139, opened 2016-07-08
Optimize baker_hubbard and wernet_nilsson functions

connorbrinton:
Hi MDTraj team,

Here are some changes I've made to the hbond-finding functions to make them run about six times faster, as well as adding the option sidechain_only, to consider only sidechain-sidechain interactions. I still have to write a test for the code, but in my ad-hoc testing all outputs have been identical to previous versions of mdtraj, just faster. Feel free to integrate this code however you like.

Note: I also haven't added documentation for the sidechain_only option, which should be done if it's something you want to support.

Thanks,
Connor

### DIFF
--- a/mdtraj/geometry/hbond.py
+++ b/mdtraj/geometry/hbond.py
@@ -26,7 +26,7 @@
 ##############################################################################
 
 from __future__ import print_function, division
-from itertools import product
+import itertools
 import numpy as np
 from mdtraj.utils import ensure_type
 from mdtraj.geometry import compute_distances, compute_angles
@@ -37,7 +37,6 @@ __all__ = ['wernet_nilsson', 'baker_hubbard', 'kabsch_sander']
 ##############################################################################
 # Functions
 ##############################################################################
-
 
 def wernet_nilsson(traj, exclude_water=True, periodic=True):
     """Identify hydrogen bonds based on cutoffs for the Donor-H...Acceptor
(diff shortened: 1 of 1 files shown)


connorbrinton:
Also, just so you know what to look for, the following were the optimizations I made:
(1) Generate donor-hydrogen-acceptor lists using numpy (instead of using itertools.product)
(2) Calculate angles using distances (from compute_distances, which is really fast) and the law of cosines (rather than using compute_angles)
(3) Don't calculate angles for bonds that don't satisfy the distance criteria (this one is the biggest improvement)

mdtraj/geometry/hbond.py
```diff
@@ -38,8 +37,7 @@
 # Functions
 ##############################################################################
 
-
-def wernet_nilsson(traj, exclude_water=True, periodic=True):
+def wernet_nilsson(traj, exclude_water=True, sidechain_only=False, periodic=True):
```
mpharrigan: Can you add `sidechain_only` to the end of this list. This makes sure it maintains API compatibility when using positional arguments

mpharrigan:
Looks like this also does some badly-needed refactoring cc #705

mpharrigan:
Maybe also related #522

connorbrinton:
I moved the sidechain_only option, let me know if there's anything else that should be changed.

connorbrinton:
Looks like there are tests for hbonds, and that some of them didn't pass. I'll figure out why, and fix the code if necessary.

connorbrinton:
So the failed tests were definitely my fault (oops), but I found the things I had done wrong and fixed them, so the tests are passing now. Hooray!

connorbrinton:
I spoke too soon last Friday saying that all the tests were passing, because some of the 64-bit builds hadn't finished yet. They gave different inputs (empty trajectories) to the hbond functions, so they exposed problems with empty trajectories, which I fixed this morning.

connorbrinton:
@mpharrigan could you check the AppVeyor and Travis CI builds to see if the two failed builds are because of my changes?  From what I could tell from the console the AppVeyor build failure was because of a build system problem and the Travis CI build failure was because of a race condition in a SASA test. But if they're because of something I did, I'll fix it. Thanks!

mpharrigan:
This looks good from a cursory review. @peastman do you want to review and merge?

peastman:
I'm not familiar with this part of the code, so I can't comment much on it.  I skimmed through the changes and nothing jumped out that looked obviously wrong.  But since I don't know the algorithm it's supposed to be implementing, I don't have any input on whether it's implemented correctly.

@mpharrigan if you think it looks good, I'm happy to follow your judgement on it.

mpharrigan:
I'm not that familiar either! Gonna go ahead and merge

PR #1153, opened 2016-07-28
[formats/XTC] in case of an out of bounds seek, raise IOError

marscher:
This is the standard exception in mdtraj being raised for invalid seek operations.

Before this change, one will get an IndexError (thrown by numpy), which could be confusing.


peastman:
Should it also check for `absolute < 0`?

marscher:
of course :)

peastman:
You have a typo now - tests are failing.

marscher:
now it works, but fails since scipy.stats.nanmean has been removed #1158.

PR #1158, opened 2016-08-01
scipy.stats.nanmean deprecated

smsaladi:
in favor of numpy.nanmean

http://docs.scipy.org/doc/scipy-0.17.0/reference/generated/scipy.stats.nanmean.html

Seems to be API compatible.

### DIFF
--- a/mdtraj/geometry/tests/test_drid.py
+++ b/mdtraj/geometry/tests/test_drid.py
@@ -4,7 +4,6 @@ import mdtraj as md
 from mdtraj.testing import get_fn, eq
 from mdtraj.geometry import compute_drid
 import scipy.special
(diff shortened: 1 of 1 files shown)


rmcgibbo:
Looks like np.nanmean was added in 1.8.0. Currently I think we support 1.7 and maybe even 1.6. http://docs.scipy.org/doc/numpy-dev/reference/generated/numpy.nanmean.html

mpharrigan:
Can we change numpy required version to 1.8

mpharrigan:
It looks like this function was actually removed in scipy 0.18

smsaladi:
How about this?

mpharrigan:
Oh yeah that looks good

PR #1161, opened 2016-08-03
fix typos in examples

smsaladi:


mdtraj/formats/dcd/dcd.pyx
```diff
@@ -170,15 +170,15 @@ cdef class DCDTrajectoryFile:
     >>> f.close()
 
     >>> # read all of the data with automatic closing of the file
-    >>> with DCDTrajectoryFile('mytrajectory.dcd') as f:
```
smsaladi: Actually, I'm not sure if this line was purposefully written like this or not...

mdtraj/formats/dcd/dcd.pyx
```diff
@@ -170,15 +170,15 @@ cdef class DCDTrajectoryFile:
...
+    >>> with DCDTrajectoryFile('mytrajectory.dcd', 'r') as f:
```
smsaladi: (Proposed change)

mpharrigan: Right, since the 'r' is default. Don't think it matters that much

mpharrigan:
Thanks!

PR #1162, opened 2016-08-03
scipy.stats.nanmean deprecated part two

mpharrigan:
includes #1158 

Bumps travis version and make a note in the docs


mpharrigan:
Nevermind, see latest commit in #1158

Issue #1154, opened 2016-07-29
Timestamps from DCD files


smsaladi:
I posted this [question on the forums ](http://discourse.mdtraj.org/t/timestamps-from-dcd-trajectories/159) about a month ago (though it was down for a little bit #1150), but it doesn't look like the usergroup has thoughts. Might be related to #1148 

I am trying to load two DCD trajectory files from NAMD where one is a continuation of another.

```
traj_files = glob.glob("./itasser_4/itasser_4_autopsf_tip3p_NaCl.prod_tp*.dcd")
print(traj_files)
# In case glob, does not load provide the files in the correct order
traj_files.sort()

t = md.load(traj_files,
           top='./itasser_4/itasser_4_autopsf_tip3p_NaCl.pdb',
           discard_overlapping_frames=True)
print(t)
```

> ['./itasser_4/itasser_4_autopsf_tip3p_NaCl.prod_tp00.dcd', './itasser_4/itasser_4_autopsf_tip3p_NaCl.prod_tp01.dcd']
> < mdtraj.Trajectory with 6575 frames, 16231 atoms, 4956 residues, and unitcells>

However, it looks like the timestamps are not being interpreted correctly --

```
print(t.time)
print(t.time.size)
print(np.unique(t.time).size)
```

> [   0    1    2 ..., 1505 1506 1507]
> 6575
> 5067

Is there something I should be doing such that the timestamps for each frame are correctly read?

mpharrigan:
From looking at the code, mdtraj does *not* load in time information from files. Does `dcd` contain a time field?

mpharrigan:
ref: https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/dcd/dcd.pyx#L391

mpharrigan:
dcd does not store time information (besides timestep, but it even does that weirdly)

http://www.ks.uiuc.edu/Research/namd/2.9/ug/node11.html

smsaladi:
Might not be interpreting this correctly, but I think the format stores the first step "step 1", step interval, and timestep:

http://www.ks.uiuc.edu/Research/vmd/plugins/molfile/dcdplugin.html

mpharrigan:
Ah, I see. As written it looks like mdtraj doesn't read that info for dcd

mpharrigan:
I think implementing this would be a matter of allocating a time array and reading `self.timestep.physical_time` for each frame in this loop

[...]

PR #1163, opened 2016-08-03
Properly read time information from DCD files generated by NAMD

smsaladi:
#1154

### DIFF
--- a/mdtraj/formats/dcd/dcd.pyx
+++ b/mdtraj/formats/dcd/dcd.pyx
@@ -27,6 +27,7 @@
 ##############################################################################
 
 import os
+import warnings
 import numpy as np
 cimport numpy as np
 np.import_array()
@@ -68,8 +69,8 @@ cdef ERROR_MESSAGES = {
 ##############################################################################
 
 @FormatRegistry.register_loader('.dcd')
-def load_dcd(filename, top=None, stride=None, atom_indices=None, frame=None):
-    """load_dcd(filename, top=None, stride=None, atom_indices=None, frame=None)
+def load_dcd(filename, top=None, stride=None, atom_indices=None, frame=None, as_namd=False):
+    """load_dcd(filename, top=None, stride=None, atom_indices=None, frame=None, as_namd=False)
 
     Load an DCD file from disk.
 
@@ -95,6 +96,8 @@ def load_dcd(filename, top=None, stride=None, atom_indices=None, frame=None):
         Use this option to load only a single frame from a trajectory on disk.
         If frame is None, the default, the entire trajectory will be loaded.
         If supplied, ``stride`` will be ignored.
+     as_namd : bool, default=False
+        Read/write timestep (delta) in header in NAMD internal units.
 
     Examples
     --------
@@ -131,7 +134,7 @@ def load_dcd(filename, top=None, stride=None, atom_indices=None, frame=None):
     topology = _parse_topology(top)
     atom_indices = cast_indices(atom_indices)
 
-    with DCDTrajectoryFile(filename) as f:
+    with DCDTrajectoryFile(filename, as_namd=as_namd) as f:
         if frame is not None:
             f.seek(frame)
             n_frames = 1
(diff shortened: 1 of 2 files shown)


peastman:
I don't understand what the `as_namd` option is for.  The timestep is *always* stored in that particular set of units.  That's part of the definition of the file format.  We should always convert it to standard units when reading files, and we should always convert to those units when writing files.  If we didn't, we would be generating a corrupt file.

mpharrigan:
>  The timestep is always stored in that particular set of units. 

Unfortunately, I don't think this is true. Here's what mdanalysis has to say:

https://pythonhosted.org/MDAnalysis/documentation_pages/coordinates/DCD.html

smsaladi:
Given one can set the timestep as they wish when setting up a system with NAMD (any non-negative decimal [NAMD docs](http://www.ks.uiuc.edu/Research/namd/2.6/olddocs/ug/node26.html)), any thoughts on how to parse this value robustly since files generated by NAMD need delta multiplied by 48.88821 to get to fs. In lieu of something automatic, I put in the `as_namd` option and a warning if the timestep is not a whole number (after parsing).

peastman:
You're looking at the wrong documentation.  That describes the user-level options, for which the default time step is 1 fs.  That has nothing to do with how it's represented in the file.  That is described at http://www.ks.uiuc.edu/Research/namd/2.9/ug/node11.html, which states

> The timestep is stored in the DCD file in NAMD internal units and must be multiplied by TIMEFACTOR=48.88821 to convert to fs.

smsaladi:
Right... I was just making the point that a user can set the timestep to whatever they non-negative decimal they want, which makes inferring whether the timestep read from the DCD file needs a factor multiplied or not

peastman:
It *always* needs a factor multiplied.  If someone wrote a DCD file that didn't include the scale factor, they had a bug in their DCD writer.  That scale factor is part of the definition of the file format.

smsaladi:
Not so sure about that, open_dcd_write writes `delta = 1` by default. https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/dcd/src/dcdplugin.c#L985

I don't think "NAMD internal units" is the default for the DCD format -- it wouldn't make sense for X-PLOR or CHARMM to store timestep in that format.

There just seems to be a lack of consistency, so I would be hesitant to always multiply by that scale factor.

Seems like this was discussed sometime ago https://github.com/MDAnalysis/mdanalysis/issues/187 

At that time, `orbeckst` concluded -- 
> I know of no reliable way to detect that a DCD was written by CHARMM 
or NAMD so that we could automatically get the time units right.

peastman:
In the MDAnalysis thread, the person was mistaken about the time units.  (He said his information might not be entirely correct.)  His table claimed that NAMD stored the time in ps, when in fact it uses AKMA units as described in the NAMD documentation.

Also, see the current documentation on their DCDWriter class: http://www.mdanalysis.org/mdanalysis/documentation_pages/coordinates/DCD.html#MDAnalysis.coordinates.DCD.DCDWriter.  Here's how it describes the `delta` parameter:

> timestep (MD integrator time step (!), in AKMA units); default is 20.45482949774598 (corresponding to 1 ps).

smsaladi:
Ok, I see your point. Following `MDAnalysis`, it would be reasonable to assume that the timestep in the file is in AKMA units then and multiply by that factor by default. If this is the case, then `MDtraj` ought to write DCD files with the timestep in AKMA units as well, but, currently `dcdplugin.c` calling `open_dcd_write` which writes the timestep as `1` with no argument to change this (though could interface directly with `write_dcdheader` -- will be a bit inelegant since it's a static method).

Then again ... 
1) This would introduce an backward-incompatible change (rather significant I think) since previously timestep was assigned as 1. 
2) `dcdplugin.c` writes timestep as `1` by default (which is used by VMD for writing and other utilities like catdcd)

In light of this, it might still be useful to have an argument to let the user specify how to read the timestep directly (and throw a warning if the timestep doesn't look right), maybe `akma=False` by default with a warning for one or two releases and then changing to `akma=True` by default.

peastman:
That means there's a bug in the DCD writer, and it's producing corrupt files.  The solution is to fix that!  It doesn't mean we should try to define a new variant of DCD files and then provide an argument for selecting whether it's a "real" DCD file or our nonstandard version.

It's already enough of a messed up format because other people have done essentially that in the past, and now we're stuck trying to support multiple subtly different formats that all have the same name.  Let's not make it worse for ourselves and everyone else by endorsing yet another variation.

smsaladi:
Do you have any thoughts about the error on windows build? Not sure how to deal with this.

Also, as currently, implemented, mdtraj will write dcd files with a delta of `1` because `dcdplugin.c` does it. There is no argument to change this to `open_dcd_write`

mpharrigan:
Was there a consensus reached here?

PR #1165, opened 2016-08-12
There should never be zero bins in Voxels class

tovrstra:
I got segfaults when using a large cutoff radius on a relatively small unit cell. This is a corner cases that is probably rarely met in practice but it should still be fixed.


mpharrigan:
Thanks!

PR #1167, opened 2016-08-16
Fix corner case in distancekernel.h

tovrstra:
This fixes an annoying problem that only occurs only in corner cases, which happen sufficiently frequently in my use case to hunt it down. See comment in code for details.

Please take a good look. This may degrade performance a little. Similar problems may exist elsewhere in the code.


peastman:
Nice work tracking that down.  Your change looks good to me.

setup.py
```diff
@@ -222,6 +222,9 @@ def geometry_extensions():
                      'mdtraj/geometry/src/_geometry.pyx'],
             include_dirs=['mdtraj/geometry/include',
                           'mdtraj/geometry/src/kernels'],
+            depends=['mdtraj/geometry/src/kernels/anglekernels.h',
```
marscher: what is the purpose of listing the same file 3 times?

mpharrigan:
For my own edification, what was the problem here? Was the 4th element not getting set to zero somehow?

peastman:
I expect it was probably producing a segfault?  The fourth element of the vector is never actually used, so it shouldn't matter what gets loaded into it.  But if the vector happens to be right at the end of a block of memory that's been allocated to the process, trying to read an extra value beyond it could produce an illegal memory access.

tovrstra:
I never got segfaults (yet), but when running my scripts through valgrind, I did get illegal read errors. I noticed this problem because some distances were Nan when I was expecting normal numbers. These turned out to be distances between some atom and the last atom, in the last frame of a trajectory. The problem is the fourth double read when accessing the Cartesian coordinates of that last atom.

I don't know exactly why I was sometimes getting Nan. Strictly speaking, that should not happen. I guess the SSE instructions on four doubles result in Nan when one of the inputs is Nan, even when it is not really used. (I'll write a script to test that hypothesis.)

tovrstra:
Please try the attached example as follows (on the master branch):

    valgrind python invalidread.py

This should show the invalid read errors in valgrind.

[invalidread.zip](https://github.com/mdtraj/mdtraj/files/421293/invalidread.zip)

tovrstra:
There are a few similar instances of the same pattern:

- `geometry.cpp`, on lines 144, 145, 146, 152, 154, 179
- `neighborlist.cpp`, on lines 167, 306, 360, 363

I may have missed some.

rmcgibbo:
Do you want to patch up the other instances of this bug too?

tovrstra:
I would not mind doing it. However, I'm a bit reluctant because I have very little experience with SSE instructions like these. My changes would not break things but they may not be the best solution.

I would be more comfortable, if I could reproduce the bug systematically and turn it into a (failing) unit test. Now, it only happens every once in a while (a few times in 10^5 repetitions of the same function) and many other users don't seem to have the problem.

rmcgibbo:
The code I posted in #1185 could be adapted to a proper unit test, and it's testing the same issue. The solution you posted in this PR is the best way.

marscher:
The float* constructor of fvec4 allows this faulty behaviour in general as it leaves the responsibility to the invoker, maybe it should be removed at all?

rmcgibbo:
No.

marscher:
What is your intention to proceed with this? It should be released soon, since it causes very ugly, hard to tackle bugs downstream.

tovrstra:
I'm a bit overloaded for the moment. I'll try to look into it in a few days.

tovrstra:
I have fixed all other cases. I was able to construct two unit tests that would fail without these fixes, i.e. because fvec4 was initialized with a pointer to a too-short array.

I couldn't make unit tests for all fixes, so I did add another safe guard to avoid that similar bugs would be re-introduced in future. In `mdtraj/geometry/include/vectorize_sse.h`, I've commented out the dangerous constructor and added some comments:

    /*  Note on the initialization of fvec4 with a pointer in MDTraj
        ------------------------------------------------------------
        The constructor below was often used with a pointer to a three-element array
        (instead of a four-element array) in previous versions of MDTraj. When the
        memory past the third element is garbage or outside the range assigned to the
        current process, one may get unexpected results (crash, NaNs, ...). In specific
        use cases, such a poor initialization can make MDTraj useless. To avoid that such
        bugs get re-introduced in future, the pointer constructor below is commented out.
    */
    //fvec4(const float* v) : val(_mm_loadu_ps(v)) {}

That may seem rather drastic but it is a nice way of enforcing a safe programming style.

tovrstra:
P.S. I have rebased my branch on master to test it (locally) w.r.t. the latest development version. Hence, I had to do a "force push". I hope this doesn't bother anyone.

marscher:
Thank you so much! :+1: for enforcing safe programming!

I don't think force pushes on your own fork would bother anybody.

tovrstra:
You're welcome. :) thanks to Robert for the unit test suggestion.

The tests on Travis are all failing, but it seems to be a Miniconda issue. The last line says:

    Miniconda MD5 mismatch

I'm not sure what that means.

cxhernandez:
> I'm not sure what that means.

see https://github.com/choderalab/pymbar/pull/240

cxhernandez:
Submitted a PR for the MD5 issue in https://github.com/mdtraj/mdtraj/pull/1192

tovrstra:
Rebased again after PR #1192 got merged.

Lnaden:
There are 2 separate MD5 issues. The first was that Miniconda switched to using HTTPS instead of HTTP. Right now though, MD5 checks on Miniconda3 pulls are currently broken, the "Latest" file is for 4.2.12, but the site has not updated this and still thinks the latest is 4.2.11, so the MD5s dont match. All Travis scripts which check on that will break until they update the site.

Lnaden:
It looks like they updated the site. The MD5 issue from the site mismatching should be resolved. Try re-running the travis tests, see if they get past the miniconda install.

mpharrigan:
I restarted the build

tovrstra:
I've looked at the remaining failing tests on Travis. They seem to be unrelated to this PR. One failed because it could not download a file from CONDA:

    conda                     4.2.12                   py35_0  
    Fetching package metadata ......
    CondaHTTPError: HTTP 409 Conflict
    for url <https://repo.continuum.io/pkgs/pro/noarch/repodata.json.bz2>
    An HTTP error occurred when trying to retrieve this URL.

Four others (all with numpy 1.11) failed because some names are undefined:

      File "/home/travis/miniconda3/conda-bld/conda-recipe_1478294040539/_t_env/lib/python2.7/site-packages/mdtraj/tests/test_io.py", line 90, in setUp
        hdfFile.createCArray("/", "arr_0", tables.Float32Atom(), self.data.shape, filters=io.COMPRESSION)
    AttributeError: 'File' object has no attribute 'createCArray'

and

      File "/Users/travis/miniconda3/conda-bld/conda-recipe_1478295235068/_t_env/lib/python3.5/site-packages/nose/failure.py", line 39, in runTest
        raise self.exc_val.with_traceback(self.tb)
      File "/Users/travis/miniconda3/conda-bld/conda-recipe_1478295235068/_t_env/lib/python3.5/site-packages/nose/loader.py", line 252, in generate
        for test in g():
      File "/Users/travis/miniconda3/conda-bld/conda-recipe_1478295235068/_t_env/lib/python3.5/site-packages/mdtraj/tests/test_io.py", line 144, in test_groups
        f = tables.openFile(temp, 'w')
    AttributeError: module 'tables' has no attribute 'openFile'

These are both related to PyTables. They should be fixed obviously, but probably not in this PR.

peastman:
I don't think that removing that constructor is the right solution.  It is *much* faster than the other one.  Whenever possible, we should arrange so that arrays will be sufficiently large so that we can safely use it.  That's how vectorized programming is supposed to be done.  The other constructor is just a workaround for the (hopefully rare) cases where we can't do that.

rmcgibbo:
I agree with Peter.

tovrstra:
OK. I agree too but I would rather make it work first and then make it fast.

So, if we want to fix this properly, the following attributes of the Trajectory class should be extended with at least one float:

    xyz : np.ndarray, shape=(n_frames, n_atoms, 3)
    unitcell_vectors : {np.ndarray, shape=(n_frames, 3, 3), None}
    unitcell_lengths : {np.ndarray, shape=(n_frames, 3), None}
    unitcell_angles : {np.ndarray, shape=(n_frames, 3), None}

In the C++ code, fvec4 pointer constructors can then be used safely on these arrays. Are there other arrays that should be extended? I would expect that mainly operations on the `xyz` array are performance critical. (Are there benchmarks for this?)

What is the preferred procedure? One could add a new `xyz_ext` attribute to the Trajectory class:

    xyz_ext : np.ndarray, shape=(n_frames*n_atoms*3+1)

All low-level code interacts with this `xyz_ext`. `xyz` should then become a "get-only" property as follows:

        @property
        def xyz(self):
            return xyz_ext[:-1].reshape(n_frames, n_atoms, 3)

Needles to say, quite a few other things will have to change as well. Is there a simpler way? (It may be possible to work with a `np.ndarray` subclass that over-allocates. That is a wild guess.)

marscher:
Or could we add a "safe" flag to the pointer ctor, which is true by default, but if its not, then we set the fourth element to zero? I guess this would not involve as much of a speed penalty like dropping the vectorized construction at all.

rmcgibbo:
This is low-level code. It doesn't require anything so drastic as either of above comments. One just has to be careful in dealing with the final 3 elements of the xyz array. I merged the PR but modified the comment added to vectorize_sse.h.

marscher:
Thank you. We are curious when you're going to release this :)

peastman:
This looks relevant: http://stackoverflow.com/questions/15922203/padding-large-numpy-arrays.  If you want your array to be padded, you create it as a larger array, then set your variable to a view into that array that leaves off the padding.

tovrstra:
Should we open a follow-up issue to re-introduce the fvec4 pointer constructor in a safe way in performance critical parts?

PR #1170, opened 2016-08-18
Load single-element path lists without a copy

connorbrinton:
I just got bit by this issue, and I want to make sure others don't run into it.

The `load` function can handle either a single filename or a list of filenames. When dealing with a list of filenames, it loads each trajectory and then joins them together sequentially. This joining process creates copies of the coordinate data.

When passing a one-element list containing only one filepath, `load` attempts to `join` the single loaded trajectory with an empty list (generated from a slice at trajectory.py:96), triggering a copy of the trajectory's coordinate data. Because of this, I was consistently running out of memory when trying to load a particularly large trajectory.

This pull request modifies `load` to "peel" single-element lists, so that they are treated as if the element were passed as the argument to load. Tests are included.

### DIFF
--- a/mdtraj/tests/test_load.py
+++ b/mdtraj/tests/test_load.py
@@ -1,30 +1,3 @@
-##############################################################################
-# MDTraj: A Python Library for Loading, Saving, and Manipulating
-#         Molecular Dynamics Trajectories.
-# Copyright 2012-2013 Stanford University and the Authors
-#
-# Authors: Robert McGibbon
-# Contributors:
-#
-# MDTraj is free software: you can redistribute it and/or modify
-# it under the terms of the GNU Lesser General Public License as
-# published by the Free Software Foundation, either version 2.1
-# of the License, or (at your option) any later version.
-#
-# This library is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU Lesser General Public License for more details.
-#
-# You should have received a copy of the GNU Lesser General Public
-# License along with MDTraj. If not, see <http://www.gnu.org/licenses/>.
-##############################################################################
-
-
-"""
-Tests of generic loading functionality.
-"""
-
 from mdtraj import load
 from mdtraj.testing import get_fn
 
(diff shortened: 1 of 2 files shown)


mpharrigan:
I'm worried about putting the check where you have it. What if we pass in an iterable that doesn't permit `len()`. It will also take that code path if we pass in a length-1 single filename. Should still work but seems odd. What if we're more explicit and put the check here:

https://github.com/connorbrinton/mdtraj/blob/d98d834aaadc49f0f4102f8f4acac96b09ea42b3/mdtraj/core/trajectory.py#L402

OR! we could use the new(er) `md.join()` method and make sure that doesn't do unnecessary copies. 

You also have a test failure

mdtraj/tests/test_load.py
```diff
@@ -0,0 +1,48 @@
+##############################################################################
+# MDTraj: A Python Library for Loading, Saving, and Manipulating
+#         Molecular Dynamics Trajectories.
+# Copyright 2012-2013 Stanford University and the Authors
+#
+# Authors: Robert McGibbon
```
mpharrigan: Feel free to put your name here and update the copyright date and/or remove this whole header.

mdtraj/tests/test_load.py
```diff
@@ -0,0 +1,48 @@
...
+"""
+Tests of generic loading functionality.
+"""
+
+from mdtraj import load
+from mdtraj.testing import get_fn
+
+def test_load_single():
+    """
+    Just check for any raised errors coming from loading a single file.
```
mpharrigan: Annoyingly, `nose` takes these docstrings as the name of the test. The "default" of `test_module.test_function` is better. Can you change these docstrings to ordinary comments?

mpharrigan:
Code for join is here: https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/trajectory.py#L520

Hopefully functools does this check for us?

mpharrigan:
It does:
```python
In [6]: def verbose_add(x,y):
   ...:     print("Adding")
   ...:     return x+y
   ...: 

In [7]: import functools

In [9]: functools.reduce(verbose_add, [2])
Out[9]: 2

In [10]: functools.reduce(verbose_add, [2,2])
Adding
Out[10]: 4
```

connorbrinton:
Sweet, I'll remove the manual check and use md.join instead.

connorbrinton:
I fixed the test too, and it looks like everything's passing now. Is there anything else you want me to change?

mpharrigan:
Looks great. Thanks!

Issue #881, opened 2015-06-26
Is chunk % stride == 0 requirement in iterload still necessary? 


marscher:
At least for xtc it is not, since the amount of frames is correct after loading all chunks. So my guess is that this is a relict from ancient times?

mpharrigan:
I think this is correct. It looks like you've been patching this in PyEMMA? You haven't run into any issues, correct?

PR #1171, opened 2016-08-19
Remove unnecessary restriction

mpharrigan:
fixes gh-881


marscher:
We used this setting for a about a year right now and it didn't caused any trouble.

mpharrigan:
Thanks!

PR #1174, opened 2016-09-11
Added a bunch of Van Der Waals values for all the elements that have

dr-nate:
known values. Corrected (seemingly) erroneous previous values for
H,C,O,N,F,S,P.

I added them because, while using the SASAFeaturizer in msmbuilder, it would error because there were no values for sodium (which are still in my filtered trajs).

Also, I don't know where the previous values came from, but they seem wrong. @msultan @rmcgibbo , can you guys comment on that (git blame blames you). 

Even further, I'm wondering if we should be using VdW radii for things like sodium, since the in the charged state the radii would be different. 🤔 Maybe I'm over thinking this.


msultan:
I think the difference comes from the ionic radius vs. atomic radius. I would think the former is more accurate given that we have charges on all of these atoms in our system.

dr-nate:
I think ionic radius makes sense for atoms that are ionized, but not for covalently bound atoms like a hydrogen in a hydrocarbon. Ideally, you could check the topology and know if the atom is ionized (though that might get tricky for atoms like sulfur in a sulfonyl).

We should also be clear about the radii we're using. The variable name is _ATOMIC_RADII, but then claims to get the values from a file named vdwradii.dat.  VdW are different from atomic radii, which are different from ionic radii, etc. And we should be clear about which one we're using.

Ultimately, I don't know how much we should worry about this because these radii also change based what molecule they're a part of. Usually people are looking for changes in SASA, not some super accurate value. Right? But I do think we should be explicit and consistent about where the values came from.

Also, the tests are failing because I forgot to update them. But I'd like to get some consensus on what values we should be using before I fix that.

dr-nate:
Here's how they chose vdW values in VMD: 

```
/*
 * corresponding table of VDW radii.
 * van der Waals radii are taken from A. Bondi,
 * J. Phys. Chem., 68, 441 - 452, 1964,
 * except the value for H, which is taken from R.S. Rowland & R. Taylor,
 * J.Phys.Chem., 100, 7384 - 7391, 1996. Radii that are not available in
 * either of these publications have RvdW = 2.00 Å.
 * The radii for Ions (Na, K, Cl, Ca, Mg, and Cs are based on the CHARMM27
 * Rmin/2 parameters for (SOD, POT, CLA, CAL, MG, CES) by default.
 */
static const float pte_vdw_radius[] = { 
    /* X  */ 1.50f, 1.20f, 1.40f, 1.82f, 2.00f, 2.00f,
    /* C  */ 1.70f, 1.55f, 1.52f, 1.47f, 1.54f,
    /* Na */ 1.36f, 1.18f, 2.00f, 2.10f, 1.80f,
    /* S  */ 1.80f, 2.27f, 1.88f, 1.76f, 1.37f, 2.00f,
    /* Ti */ 2.00f, 2.00f, 2.00f, 2.00f, 2.00f, 2.00f,
    /* Ni */ 1.63f, 1.40f, 1.39f, 1.07f, 2.00f, 1.85f,
    /* Se */ 1.90f, 1.85f, 2.02f, 2.00f, 2.00f, 2.00f,
    /* Zr */ 2.00f, 2.00f, 2.00f, 2.00f, 2.00f, 2.00f,
    /* Pd */ 1.63f, 1.72f, 1.58f, 1.93f, 2.17f, 2.00f,
    /* Te */ 2.06f, 1.98f, 2.16f, 2.10f, 2.00f,
    /* La */ 2.00f, 2.00f, 2.00f, 2.00f, 2.00f, 2.00f,
    /* Eu */ 2.00f, 2.00f, 2.00f, 2.00f, 2.00f,
    /* Er */ 2.00f, 2.00f, 2.00f, 2.00f, 2.00f, 2.00f,
    /* W  */ 2.00f, 2.00f, 2.00f, 2.00f, 1.72f, 1.66f,
    /* Hg */ 1.55f, 1.96f, 2.02f, 2.00f, 2.00f, 2.00f, 2.00f,
    /* Fr */ 2.00f, 2.00f, 2.00f, 2.00f, 2.00f, 1.86f,
    /* Np */ 2.00f, 2.00f, 2.00f, 2.00f, 2.00f, 2.00f, 2.00f, 2.00f,
    /* Md */ 2.00f, 2.00f, 2.00f, 2.00f, 2.00f, 2.00f, 2.00f, 2.00f,
    /* Mt */ 2.00f, 2.00f, 2.00f
};
```

Seems pretty reasonable. What do you think of using these values or something similar? We could use ionic radii for Li, Na, K, Cl, Ca, Mg, Cs, etc.

msultan:
I agree that the dominant use cases are comparisons instead of quantitative values. However, for oxygen, nitrogen, and sulphur we should use ionic radii since they do have significant partial charges within the protein/water FF. Everything else, i think we should default to atomic radius(VDW radii) and if thats not available then 2A.

dr-nate:
Ok, now updated with the full periodic table. My comment from the file explains why I chose the various radii I did

```
# These van der waals radii are taken from
# https://en.wikipedia.org/wiki/Atomic_radii_of_the_elements_(data_page)
# which references 
# A. Bondi (1964). "van der Waals Volumes and Radii". J. Phys. Chem. 68: 441. doi:10.1021/j100785a001 and doi:10.1021/jp8111556. 
# M. Mantina et al. (2009). "Consistent van der Waals Radii for the Whole Main Group". J. Phys. Chem. A. 113 (19): 5806–12. doi:10.1021/jp8111556
# Where no van der Waals value is known, a default of 2 angstroms is used.
# However, because certain atoms in biophysical simulations have a high 
# chance of being completely ionized, we have decided to give the 
# following atoms their ionic radii their ionic radii:
# +2: Be, Mg, Ca, Ba
# +1: Li, Na, K, Cs
# -1: Cl
# These ionic radii are were taken from:
# Shannon, R. D. Revised effective ionic radii and systematic studies of interatomic distances in halides and chalcogenides. Acta Crystallographica Section 
# For most atoms, adding electrons usually doesn't change the radius much 
# (<10%), while removing them changes it substantially (>50%). Further, 
# when atoms like N, S, and P, are positive, they are bound to atoms in such 
# a way that would "hide" their radii anyway. We have therefore chosen to just 
# use their vdW radii.
```

Further, I was having trouble getting a more current version of Gromacs (v5.1.1), which uses the exact same radii values that we are using, to agree with our tests for alanine dipeptide. The reason for this is multifactorial. First, my initial tests showed that they were 61% mismatch between Gromacs and the mdtraj code at 2 sig. figures. I changed from using the file `frame0.trr` (higher precision) to using `frame0.xtc` for doing the calculation in Gromacs, and that dropped the mismatch down to ~30%. I then realized Gromacs sets a default number of point to place the probe around the atom at 24, which is much lower than the 960 mdtraj uses. Increasing to 960 points in gromacs dropped the mismatch down to ~8%. I finally decided that this was pretty damn close and the rest is probably down to the precision of the files or how they place the points, so changed the test to require only 1 sig. fig. agreement. If anyone thinks this is too loose, please let me know and I can be flesh this out some more.

dr-nate:
I should point out that the tests are failing unrelated to the code I introduced. On Linux, they're failing in python2.7 because of some floating point issue in `mdtraj.geometry.tests.test_angles.test_angle_pbc_0`. On Mac OS X, Python2.7 they failing because `'ascii' codec can't decode byte 0xe2 in position 1769: ordinal not in range(128)`. And on Windows some DLL is failing to load.

mpharrigan:
@msultan are you content with this now?

cxhernandez:
I think the thumbs up is a yes.

mpharrigan:
I think the travis failures are relevant. The comments have non ascii em dash in citations (this has come up before). I'll fix it in the online editor

PR #1182, opened 2016-10-15
Add 'sidechain' and 'sidechain-heavy' options to compute_contacts

jchodera:
This PR adds two additional modes to `compute_contacts`:
```
+            'sidechain' : distance is the closest distance between any
+                two atoms in residue sidechains
+            'sidechain-heavy' : distance is the closest distance between
+                any two non-hydrogen atoms in residue sidechains
```
Code from @juliebehr.

I've also added tests, but I wasn't sure how those distances were computed in the first place so I used `pymol` to compute the test distances. In case that script is of use, it is attached here. This is run with `@compute_distances_pymol.txt` from within pymol.

[compute_distances_pymol.txt](https://github.com/mdtraj/mdtraj/files/531468/compute_distances_pymol.txt)


jchodera:
That's odd---`mdtraj` seems to have [an unusual definition of what qualifies as a protein sidechain](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/topology.py#L1391-L1395):
```python
    @property
    def is_sidechain(self):
        """Whether the atom is in the sidechain of a protein residue"""
        return (self.name not in set(['C', 'CA', 'N', 'O'])
                and self.residue.is_protein)
```
What about the hydrogens on these atoms (`HA` and `H`)?

jchodera:
OK, tests pass locally.

rmcgibbo:
If you want to add HA and H to that list, that seems reasonable to me.

jchodera:
> If you want to add HA and H to that list, that seems reasonable to me.

Let's do that in a separate PR, since a whole mess of tests will need to be updated then.

PR #1183, opened 2016-10-16
[WIP] Allow anchor molecules to optionally be specified in Trajectory.image_molecules

jchodera:
This PR allows the user to optionally specify which molecules are to be used as anchor molecules in `Trajectory.image_molecules()`. This is useful when the automatic scheme doesn't do the right thing. We ran into a case where a peptide bound to a protein ended up jumping all over the place as a result.

I've also added a test for the various features of `image_molecules()` since there didn't appear to be any tests for this method.


jchodera:
This isn't quite working as I had hoped yet, since the peptide is still jumping around. Let me see if I can improve the logic a bit.

mpharrigan:
See #1106

jchodera:
It looks like the problem we're having is that imaging the peptide based on its center of mass leads to it jumping around the box. Instead, we probably want to image it in a way that minimizes its minimum distance to the protein it is binding to. I'm trying to figure out the best way to do that, since it seems like that is the intention of the logic behind `image_molecules`.

Where does MDTraj initially image everything inside a unique periodic box when loading periodic trajectories (e.g. those from OpenMM where no imaging was done)? It seems like a lot of the functions that use PBCs (like `find_closest_contact`) make this assumption, but I can't find where this initial imaging is done.

jchodera:
Oh! #1106 looks like a huge improvement!

jchodera:
Deprecating this PR in favor of #1106

PR #1200, opened 2016-11-08
Move guess_anchor_molecules to Topology

### LINKED ISSUE #1197: Move guess anchor atoms to topology

mpharrigan:
fixes #1197

### DIFF
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -1111,6 +1111,30 @@ class Topology(object):
             molecules[atom_molecule[atom.index]].add(atom)
         return molecules
 
+    def guess_anchor_molecules(self):
+        """Guess anchor molecules for imaging
+
+        Returns
+        -------
+        anchor_molecules : list of atom sets
+            List of anchor molecules
+        """
+        molecules = self.find_molecules()
+
+        # Select the anchor molecules.
+        molecules.sort(key=lambda x: -len(x))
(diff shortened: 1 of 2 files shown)


mdtraj/core/trajectory.py
```diff
@@ -1857,32 +1857,6 @@ def _check_valid_unitcell(self):
     def _have_unitcell(self):
         return self._unitcell_lengths is not None and self._unitcell_angles is not None
 
-    def guess_anchor_molecules(self):
```
marscher: you could deprecate the method first and redirect it to the topology impl.

mpharrigan: It was never in a release. This function was added a couple of weeks ago and I said it should be on Topology, but the move wasn't performed at the time

marscher:
Ah okay ;)

mdtraj/core/topology.py
```diff
@@ -1111,6 +1111,30 @@ def find_molecules(self):
             molecules[atom_molecule[atom.index]].add(atom)
         return molecules
 
+    def guess_anchor_molecules(self):
+        """Guess anchor molecules for imaging
```
peastman: If I didn't already know, I'd be confused about what that description was talking about.  Perhaps add a "see also" referencing `Trajectory.image_molecules()`?

mpharrigan: Good point

PR #1208, opened 2016-11-10
Install path for zlib on linux (too)

mpharrigan:


mpharrigan:
There's a problem with https://github.com/conda-forge/mdtraj-feedstock/pull/2 due to the introduction of zlib dependency with TNG support. I thought these directories were included by default. They aren't. Things were probably working because zlib is on every system ever except the conda-forge docker container :) In particular, zlib is a dependency of python, so any system with python installed into its standard location will be able to build.

I've added a "patch" of this pr to https://github.com/conda-forge/mdtraj-feedstock/pull/2 so I don't have to make a new version for pypi (and can test if it works on the ci over there)

Issue #1218, opened 2016-12-09
Average structure


tabramy:
Hi,

I am trying to calculate RMSD of the ligand from the average structure in a protein-ligand trajectory (both after protein alignment and ligand alignment), however, I am not able to find a MDTraj built in function to calculate the average structure. Can you please let me know if such a built in function exists, for example similar to pytraj's pytraj.mean_structure.

If such a function does not exist, please let me know what is the best way to go around this using the MDTraj functions. Otherwise this has to be calculated by a user manually?

Thanks

mpharrigan:
cc #1159

PR #1221, opened 2016-12-15
Implemented compute_average_structure()

### LINKED ISSUE #1159: Consensus structure
It would be nice if mdtraj could be given an ensemble of similar structures and find a "consensus" structure that is an average of them. I saw this paper http://www.ncbi.nlm.nih.gov/pubmed/17369627

I'm not sure about the current literature for this and/or whether this functionality is available in other packages

peastman:
Fixes #1159 and #1218.

### DIFF
--- a/mdtraj/geometry/alignment.py
+++ b/mdtraj/geometry/alignment.py
@@ -269,3 +269,33 @@ def rmsd_qcp(conformation1, conformation2):
     max_eigenvalue = scipy.optimize.newton(f, E0, df)
     rmsd = np.sqrt(np.abs(2.0 * (E0 - max_eigenvalue) / n_atoms))
     return rmsd
+
+def compute_average_structure(xyz):
+    """Compute the average structure from a set of frames.
+
+    The frames are first aligned to minimize the RMSD from the average structure,
+    and then the average is returned.
+
+    Parameters
+    ----------
+    xyz : ndarray, shape = (n_frames, n_atoms, 3)
+        xyz coordinates of each atom in each frame
+
+    Returns
+    -------
+    average : ndarray, shape = (n_atoms, 3)
+        the average structure
+    """
+    n_frames = xyz.shape[0]
+    n_atoms = xyz.shape[1]
+    candidate = xyz[0]
+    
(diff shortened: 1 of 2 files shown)


peastman:
Does this look ok to everyone?  Shall I merge it?

mdtraj/geometry/alignment.py
```diff
@@ -269,3 +269,33 @@ def rmsd_qcp(conformation1, conformation2):
     max_eigenvalue = scipy.optimize.newton(f, E0, df)
     rmsd = np.sqrt(np.abs(2.0 * (E0 - max_eigenvalue) / n_atoms))
     return rmsd
+
+def compute_average_structure(xyz):
+    """Compute the average structure from a set of frames.
+
+    The frames are first aligned to minimize the RMSD from the average structure,
```
jchodera: Why not add an optional argument to allow users to specify whether or not the frames should first be RMSD-aligned? I could imagine both schemes could be useful.

mdtraj/geometry/alignment.py
```diff
@@ -269,3 +269,33 @@ def rmsd_qcp(conformation1, conformation2):
...
+    n_atoms = xyz.shape[1]
+    candidate = xyz[0]
+    
+    # In practice there is usually negligible improvement after the second iteration.
+    # We do three just to be safe.
+    
+    for iteration in range(3):
+        average = np.zeros((n_atoms, 3))
+        for frame in range(n_frames):
+            average += transform(xyz[frame], candidate)
```
jchodera: Are you just aligning to the first frame? Is this approach equivalent to selecting the reference structure for alignment that minimizes the RMSD, as the docstring suggests?

peastman: See #1159.  We discussed this at some length there.

jchodera: Instead of hardcoding three iterations (is there a proof for that?) why not check for actual convergence?

mpharrigan:
I'm going to merge this, @jchodera feel free to submit an additional pr to address the small issues you picked out

PR #1228, opened 2017-02-03
Fix conda test running

mpharrigan:
Use test/source_files to run example notebooks. Bring the rest
of it in sync with msmbuilder recipe and conda-forge recipe

addresses CI issue raised in #1227


mpharrigan:
Weird doc building issue. I can reproduce but don't have any idea why it's happening. I'll try updating sphinx (??)

PR #1229, opened 2017-02-07
Build docs via conda package

mpharrigan:
idk kinda neat (this includes the commits in #1228 )


marscher:
This is a really good idea to handle the doc building procedure. Did you yet figured out, why mdtraj can not be imported during the build of mdtraj-doc?
How did you convince conda-build to install the local version of mdtraj? There is no switch to conda-build to prefer these.

mpharrigan:
> Did you yet figured out, why mdtraj can not be imported during the build of mdtraj-doc?

I don't remember exactly why, but I think i had no ideas why the import wasn't working on travis and gave up on it for that reason

> How did you convince conda-build to install the local version of mdtraj? There is no switch to conda-build to prefer these.

The trick I use here is to set the version as the git hash, and then in the requirements, require that git hash

mpharrigan:
This may be possible now with enhanced jinja support for conda build variants. Specifically: 

https://conda.io/docs/user-guide/tasks/build-packages/variants.html#referencing-subpackages

PR #1233, opened 2017-02-10
[geometry] fix compatibility with old visual studio for Python 2.7

marscher:
Also re-enable Py27 in Appveyors build matrix.

This PR is already part of a patch set for release 1.8 under Omnia and conda-forge channels to enable the Python 2.7 support for windows.


mdtraj/geometry/include/math_patch.h
```diff
@@ -1,11 +1,14 @@
 // Math functions for old versions of visual studio
-
+#ifndef __MATH_PATH_H_
+#define __MATH_PATH_H_
 #ifdef _MSC_VER
 #if _MSC_VER < 1900 // prior to vs2015
-float roundf(float x) {
+#pragma once
```
mpharrigan: why is this pragma necessary?

mdtraj/geometry/include/math_patch.h
```diff
@@ -1,11 +1,14 @@
 // Math functions for old versions of visual studio
-
+#ifndef __MATH_PATH_H_
```
mpharrigan: nit pick: This should probably be `MATH_PATCH_H`. (path --> patch) double underscores are (supposed to be) reserved for compiler vendors

mdtraj/geometry/include/math_patch.h
marscher: it is probably redundant with the inline qualifier. I got trouble linking these symbols, because they were included twice. We can have another CI run without the pragma once.

mdtraj/geometry/include/math_patch.h
marscher: -Wpedantic 👍

mdtraj/geometry/include/math_patch.h
mpharrigan: I thought it might be redundant with the ifdef guards

mpharrigan:
thanks so much @marscher !

PR #1238, opened 2017-02-27
Skip renumbering if no bonds in mol2

mattwthompson:
Currently, reading in a mol2 without bonds fails when it tries to renumber the bonds to 0 index:

```
import mdtraj as md

md.load('li.mol2')
```

with mol2 file:

```
@<TRIPOS>MOLECULE
RES
1 0 1 0 1 
SMALL
NO_CHARGES
@<TRIPOS>CRYSIN
   10.0000    10.0000    10.0000    90.0000    90.0000    90.0000  1  1
@<TRIPOS>ATOM
       1 Li           0.0000     0.0000     0.0000 opls_404      1 RES    
@<TRIPOS>BOND
@<TRIPOS>SUBSTRUCTURE
       1 RES             1 RESIDUE    0 **** ROOT      0 
```

> Traceback (most recent call last):
  File "//anaconda/lib/python3.5/site-packages/mdtraj/core/trajectory.py", line 428, in load
    value = loader(filename, **kwargs)
  File "//anaconda/lib/python3.5/site-packages/mdtraj/formats/mol2.py", line 108, in load_mol2
    offset = bonds_mdtraj.min()  # Should this just be 1???
  File "//anaconda/lib/python3.5/site-packages/numpy/core/_methods.py", line 29, in _amin
    return umr_minimum(a, axis, None, out, keepdims)
ValueError: zero-size array to reduction operation minimum which has no identity

I changed this to only renumber bonds if ```mdraj_bonds``` is not empty. I also added a test with a mol2 file that has no bonds. All of the mol2 files currently being tested have bonds.

### DIFF
--- a/mdtraj/formats/mol2.py
+++ b/mdtraj/formats/mol2.py
@@ -105,8 +105,10 @@ def load_mol2(filename):
     atoms_mdtraj["chainID"] = np.ones(len(atoms), 'int')
 
     bonds_mdtraj = bonds[["id0", "id1"]].values
-    offset = bonds_mdtraj.min()  # Should this just be 1???
-    bonds_mdtraj -= offset
+
+    if bonds_mdtraj.any():
+        offset = bonds_mdtraj.min()  # Should this just be 1???
+        bonds_mdtraj -= offset
 
     top = Topology.from_dataframe(atoms_mdtraj, bonds_mdtraj)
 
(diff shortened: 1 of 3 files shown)


mpharrigan:
I'm not super familiar with the mol2 format. @peastman or @jchodera do you want to take a look at this?

Otherwise, looks good and includes a test so I'm ok with merging

rmcgibbo:
+1

PR #1242, opened 2017-03-07
MAE Reader

### LINKED ISSUE #887: MAE reader
Would it be helpful to anyone if I were to write a module that reads in Schrodinger Maestro format ".mae" files? This would obviate converting it to PDB or some other format first.

mpharrigan:
fixes #887 

Read only


cxhernandez:
Could we add a test for loading?

mpharrigan:
Does anyone want to volunteer to add a simple test @cxhernandez @evanfeinberg

Basically if you have a system where you have both a pdb and mae: load them both and check for equality

cxhernandez:
```python
In [12]: mae.top
Out[12]: <mdtraj.Topology with 1 chains, 3 residues, 22 atoms, 0 bonds at 0x7f465c0b8ac8>

In [13]: pdb.top
Out[13]: <mdtraj.Topology with 1 chains, 3 residues, 22 atoms, 21 bonds at 0x7f4628e18f98>
```

looks like the MAE topology doesn't have bond info. Not sure if this is an issue with how VMD outputs MAE files or not.

marscher:
@cxhernandez Any progress on the readers test? Thanks for reporting your status.

marscher:
MAE does not set unitcell lengths, is this intended? I can provide a fix for the Trajectory.__hash__ method if the unitcell is really missing or defined differently.

cxhernandez:
> Any progress on the readers test? Thanks for reporting your status.

No, sorry. I would proceed with cutting the new release as I won't have time to fix this issue by this week.

cxhernandez:
> Maestro is not aware of periodic boundary conditions, so it does not wrap distances in the way that, for example, PBC does.

via https://www.schrodinger.com/kb/421

So perhaps not?

rmcgibbo:
I assume this is outdated. Feel free to re-open if desired.

PR #1247, opened 2017-03-24
added diffrentialable contacts option

msultan:
Added option to calculate contacts using a diffrentialable scheme when using the residue contact distance option.

### DIFF
--- a/mdtraj/geometry/contact.py
+++ b/mdtraj/geometry/contact.py
@@ -219,7 +219,7 @@ def compute_contacts(traj, contacts='all', scheme='closest-heavy', ignore_nonpro
                 distances[:, i] = atom_distances[:, index : index + n].min(axis=1)
             else:
                 distances[:, i] = soft_min_beta / \
-                                  np.log(np.sum(np.exp(soft_min_beta/ \
+                                  np.log(np.sum(np.exp(soft_min_beta/
                                                        atom_distances[:, index : index + n]), axis=1))
 
     else:
--- a/mdtraj/geometry/contact.py
+++ b/mdtraj/geometry/contact.py
@@ -218,7 +218,9 @@ def compute_contacts(traj, contacts='all', scheme='closest-heavy', ignore_nonpro
             if not soft_min:
                 distances[:, i] = atom_distances[:, index : index + n].min(axis=1)
             else:
-                distances[:, i] = beta / np.log(np.sum(np.exp(beta/atom_distances[:, index : index + n]), axis=1))
+                distances[:, i] = soft_min_beta / \
+                                  np.log(np.sum(np.exp(soft_min_beta/ \
+                                                       atom_distances[:, index : index + n]), axis=1))
 
     else:
         raise ValueError('This is not supposed to happen!')
--- a/mdtraj/geometry/contact.py
+++ b/mdtraj/geometry/contact.py
@@ -40,7 +40,50 @@ __all__ = ['compute_contacts', 'squareform']
 # Code
 ##############################################################################
 
-def compute_contacts(traj, contacts='all', scheme='closest-heavy', ignore_nonprotein=True, periodic=True):
+##############################################################################
+# MDTraj: A Python Library for Loading, Saving, and Manipulating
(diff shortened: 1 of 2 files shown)


msultan:
Exact scheme is inspired from [here](http://plumed.github.io/doc-v2.0/user-doc/html/mindist.html)

mdtraj/geometry/contact.py
```diff
@@ -40,7 +40,50 @@
 # Code
 ##############################################################################
 
-def compute_contacts(traj, contacts='all', scheme='closest-heavy', ignore_nonprotein=True, periodic=True):
+##############################################################################
+# MDTraj: A Python Library for Loading, Saving, and Manipulating
```
mpharrigan: What happened here? Don't re-add the header

msultan: yea, fixing it now

mpharrigan:
This should be a new `scheme` not a new parameter. And it should probably be named "softmin" or something https://en.wikipedia.org/wiki/Softmax_function

mpharrigan:
Or maybe it shouldn't be a new scheme because it can apply to each of the other schemes (except CA). Maybe add an exception if it's set to true and scheme is alpha carbon

Probably should be called `softmin` or something

mdtraj/geometry/contact.py
```diff
@@ -71,6 +114,14 @@ def compute_contacts(traj, contacts='all', scheme='closest-heavy', ignore_nonpro
     periodic : bool, default=True
         If periodic is True and the trajectory contains unitcell information,
         we will compute distances under the minimum image convention.
+    diffrentialable_scheme : bool, default=False
+        If diffrentialable_scheme is true, we will use a diffrentiable version of
+        the scheme. The exact expression used
+         is d = \frac{\beta}{log\sum_i{exp(\frac{\beta}{d_i}})} where
+         beta is user parameter which defaults to 500
```
mpharrigan: this is not the actual beta default

mdtraj/geometry/contact.py
```diff
@@ -71,6 +114,14 @@ def compute_contacts(traj, contacts='all', scheme='closest-heavy', ignore_nonpro
...
+    beta : float, default=20
```
mpharrigan: units? I think nm?

msultan: yep, also good catch on the 500.

mdtraj/geometry/contact.py
```diff
@@ -71,6 +114,14 @@ def compute_contacts(traj, contacts='all', scheme='closest-heavy', ignore_nonpro
     periodic : bool, default=True
         If periodic is True and the trajectory contains unitcell information,
         we will compute distances under the minimum image convention.
+    diffrentialable_scheme : bool, default=False
```
mpharrigan: link to plumed user manual in the docs?

msultan:
Thanks for all the feedback @mpharrigan

msultan:
For CA, it doesnt matter if we use soft_min since the logsumexp cancels out. I dont think we need to explicitly encode that.

mpharrigan:
> For CA, it doesnt matter if we use soft_min since the logsumexp cancels out. I dont think we need to explicitly encode that.

Maybe a warning? My thought is if people are doing those combinations of options, it's probably a mistake

mdtraj/geometry/contact.py
```diff
@@ -218,7 +218,9 @@ def compute_contacts(traj, contacts='all', scheme='closest-heavy', ignore_nonpro
             if not soft_min:
                 distances[:, i] = atom_distances[:, index : index + n].min(axis=1)
             else:
-                distances[:, i] = beta / np.log(np.sum(np.exp(beta/atom_distances[:, index : index + n]), axis=1))
+                distances[:, i] = soft_min_beta / \
+                                  np.log(np.sum(np.exp(soft_min_beta/ \
```
mpharrigan: this `\` is unnecessary

mpharrigan:
errors seem unrelated

msultan:
tests seem to be failing for some random reasons now.

PR #1254, opened 2017-04-19
 Fix tranform calculation and Transform object to be more general.

dr-nate:
Currently, if you create a transform object with `compute_transformation`,
you will get a Transform object that will only be able to correctly
transform groups of atoms that are the same as the target, i.e. you currently
can't align an entire system including water when you just compute
the transformation matrix on protein.

Perhaps there was a reason why it was written this way, but it currently does not deliver the behavior I was expecting. This fix should be more general and what users expect. I will come up with some tests and will commit them also.

### DIFF
--- a/docs/whatsnew.rst
+++ b/docs/whatsnew.rst
@@ -6,6 +6,8 @@ These are new features and improvements of note in each release.
 v1.9 (Development)
 ------------------
 
+- PR #1254: Fixed/updated alignment code to be more general.
+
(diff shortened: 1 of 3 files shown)


dr-nate:
Ok, should be ready to merge.

cxhernandez:
@dr-nate: could you update the [changelog](https://github.com/mdtraj/mdtraj/blob/master/docs/whatsnew.rst) as well?

PR #1246, opened 2017-03-24
handle a single frame being passed to sparta+

smsaladi:
See #1245

PR #1257, opened 2017-05-01
Fix #1256: Add O1 as alternative definition for water oxygen.

### LINKED ISSUE #1256: Extend definition of standard water to perceive bonds in PDB files
As far as I understand, the PDB file format doesn't require CONECT records for water molecules, so MDTraj attempts to automatically perceive the topology of the water molecules in a PDB files using the standard definitions in `mdtraj.formats.pdb.data.residues.xml`, which requires water molecules to have residue name `HOH` and atom names `O`, `H1` and `H2`. However, naming conventions for water molecules are less standard than amino acids, and different tools will create different water records (e.g. leap uses `WAT` as residue name for waters).

Would it be possible to detect water molecules independently from the specific names used used for residue/atoms? Or maybe to extend this definition to be more flexible?

Silent failing in perceiving water molecules topology recently introduced in our tools a bug that was tricky to track down (see choderalab/openmoltools#253).

andrrizzi:
Should be ready for review.

### DIFF
--- a/mdtraj/formats/pdb/data/pdbNames.xml
+++ b/mdtraj/formats/pdb/data/pdbNames.xml
@@ -270,7 +270,7 @@
   <Atom name="H73" alt1="3H5M"/>
  </Residue>
  <Residue name="HOH" alt1="H20" alt2="WAT" alt3="SOL" alt4="TIP3">
-  <Atom name="O" alt1="OW" alt2="OH2"/>
+  <Atom name="O" alt1="OW" alt2="OH2" alt3="O1"/>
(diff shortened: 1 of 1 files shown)


andrrizzi:
It looks like there's a spartaplus test that is skept on Linux and failing on MacOSX. I doubt it's related to this PR though (it seems it's caused by some test file missing).

cxhernandez:
Thanks @andrrizzi! Seems like that test failure is a known issue: https://github.com/mdtraj/mdtraj/pull/1246#issuecomment-289172611

PR #1260, opened 2017-05-10
Make compute_directors() user-facing

tcmoore3:
Described in a9b629b.

@ctk3b you may want to look this over.


ctk3b:
+1 from me

Two suggestions: 
- move the tests into the existing [file](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/geometry/tests/test_order.py)
- add the compute to the [docs](https://github.com/mdtraj/mdtraj/blob/master/docs/analysis.rst)

mpharrigan:
This looks great! Thanks @tcmoore3 

Looks like the test failure is the one we already know about.

PR #1265, opened 2017-05-12
[xtc] approx_nframes returns at least one.

### LINKED ISSUE #1264: [xtc] IndexError "out of bounds on buffer access" for single frame 'trajectories'
This is a rare case, but it should be fixed. Actually I'm the person to blame here, I'll send in a PR when I have time. It happens, when trying to access the length of the file and seeking ops.

marscher:
Fixes #1264


mpharrigan:
Thanks for figuring this out and fixing it!

Issue #1267, opened 2017-05-26
"In" operator in Atom Selection DSL


jaimergp:
Hello,

is it possible to use `in` as an operator in the DSL? I need to restrain some atoms in my OpenMM simulation, and for the time being I have to use queries as this one:

`backbone or index 7799 or index 7849 or index 7800 or index 7801 or index 7851 or index 7853 or index 7856 or index 7857 or index 7861 or index 7862 or index 7863 or index 7864 or index 7865 or index 7866 or index 7766 or index 7767 or index 7872 or index 7874 or index 7774 or index 7775 or index 7776 or index 7777 or index 7778 or index 7761 or index 7783 or index 7784 or index 7785 or index 7838 or index 7839 or index 7840 or index 7841 or index 7842 or index 7846 or index 7847 or index 7848`

I am wondering if it would be possible to write that like 

`backbone or index in (7799, 7849, 7800, 7801, 7851, 7853, 7856, 7857, 7861, 7862, 7863, 7864, 7865, 7866, 7766, 7767, 7872, 7874, 7774, 7775, 7776, 7777, 7778, 7761, 7783, 7784, 7785, 7838, 7839, 7840, 7841, 7842, 7846, 7847, 7848)`

If not, how hard would it be to implement that? I could try to submit a PR if that's doable!

hainm:
' or index '.join(indices) would do the trick.

jaimergp:
Yep, that's no problem inside a Python script. The thing is I develop this OpenMM wrapper called [`ommprotocol`](https://github.com/insilichem/ommprotocol) that uses a YAML input file. I recently added support for MDTraj DSL queries in that input file, and the users here at the lab are not that tech-savvy to think of not typing `index` several times, instead of using something like SublimeText multiple cursors. 

I could write a preprocessor for the `in` operator, but IMO the effort would better invested in submitting that PR.

hainm:
ah, make sense to add.

mpharrigan:
I took a stab at implementing this at one point. The syntax was going to be

```
backbone or index 799 7895 2342
```
which I think is how VMD does it. It would also remove the need for the insertion of an implicit equality operation

```
resname ALA
```
would turn into `resname in ["ALA"]`. 

Lemme see if I can find the branch where I started working on this

mpharrigan:
added in #1268

PR #1268, opened 2017-05-26
Add "in" selection to selection language

mpharrigan:
This is incomplete. I don't have much time to work on it right now if someone wants to take a stab at debugging it


mpharrigan:
cc #1267

mpharrigan:
Ok, I think this works now

Issue #1230, opened 2017-02-07
Velocities in trajectories


EnricoRiccardi:
Dear developers,

We are a group working on rare events (www.pyretis.org). We developed a code to manage md simulations to efficiently sample transition events. 

To increase the versatility of our code, we were thinking to use mdtraj such to be able to interface with several MD engines (gromcacs, amber etc) with a very elegant and direct implementation. 

Mdtraj seems to be exactly what we were looking for. Our problem is that we need to deal also with the velocities in a trajectories/frame. Do you think it is feasible to have this modification in an official release? We understand that this might influence the code speed, but for us, it would be the essential element which would make us decide if to use mdtraj or not. 

Thank you for consideration in any case and looking forward for an answer. 
Bests,
Enrico 

ps. is there any plan to deal with the hight precision .g96?

John-Jumper:
I am a user of MDTraj as well and I was just about to file a similar feature request.

I would like to have a collection of additional timeseries as numpy arrays in the MDTraj trajectory.  I am thinking of simulation energy and replica exchange index for me.  The user would have to guarantee that all additional timeseries' first axis has length equal to n_frames.  The only requirements for MDTraj would be to (1) copy the additional data whenever the Trajectory is copied and (2) slice the additional timeseries when the trajectory is sliced (keeping the numpy dtype). 

Best regards,
John

mpharrigan:
Hi @EnricoRiccardi and @John-Jumper 

We don't currently have plans for adding support for velocities or arbitrary timeseries data. Pull requests are welcome, however

rmcgibbo:
These are both reasonable requests. The second is probably easier to implement than the first, because the crux of the first request will be modifying the appropriate loaders to load velocity information in the first place.

To my knowledge, neither of these features is currently a high priority for the existing development team, but PRs are always welcome.

EnricoRiccardi:
I would allow myself to make a small proposal for a simple way out. Would be possible to load a trajectory as a collection of ONLY velocities (or forces)?  
As an example, the function md.load() would load only the positions of a trj as it is now, and 
the function md.vload() would load only the velocities (md.fload() the forces).

The rest of the code would not see any difference. We can do scripts and other trick to do it ourself to handle this, but I think it would be easier if it could be in the official release of mdtraj. 

ps. also the request of  John-Jumper would be useful to us.

mpharrigan:
[...]

PR #1271, opened 2017-06-02
Extracting velocities/forces from TRR files (hidden API)

dwhswenson:
This PR is related to the questions in #1230, but doesn't attempt to change the core MDTraj API. Instead, this just provides options (within the "hidden" methods like `_read` and `_write`) to read/write velocities and forces in TRR files.

Basically, I wanted to have MDTraj's seek capabilities in TRR files, plus the Cython speed, without having to fork and maintain it myself (or add another requirement). So I've tried to implement it so that I can get the extra features I'd like, without my code getting in the way. Call them hidden features, or preparation for solving #1230 in the future.

One thing I would like to change, but will do as you prefer: Right now, the number of return values for `_read` depends on the values of the parameters `get_velocities` and `get_forces`. If both are false, it returns 5 values. If one is true, it returns 6 values. If both are true, it returns 7 values.

That's ugly, but was done with the intent of minimally modifying any other code. If you'd prefer, I'd be happy to make it always return 7 values, and with `None` in the location of anything that isn't retrieved. I consider methods beginning with `_` to be in hidden APIs, and therefore subject to change without worrying about backward compatibility. But you might feel differently, so I maintained backward compatibility.


rmcgibbo:
Nice work.

I think it makes a lot of sense to add this ability to the low-level TRRTrajectoryFile whether or not it ends up getting exposed in the Trajectory class.

As the modifications are to method with an underscore, prefix, I think it's totally fine to change the API to the (more natural) unconditional 7-elements return value. The tradeoff is between code simplicity and backward compatibility, and I'm more than happy to favor simplicity in this case.

dwhswenson:
Thanks @rmcgibbo. I switched to `._read` always returning a 7-tuple, and also added support for when `atom_indices` is used.

I think this is ready for review. It looks like the tests that it fails are a general problem right now with Linux builds on Travis (AppVeyor and Travis/Mac tests pass).

Summary of what's in here:

* `TRRTrajectoryFile._read` always returns 7 items: `xyz`, `time`, `step`, `box`, `lambda`, `vel`, `forces`.
* `TRRTrajectoryFile._read` takes boolean arguments `get_velocities` and `get_forces`: if `True`, it returns the numpy arrays in the slots for those items; if `False`, those slots in the return tuple are `None`. Default is `False` for both.
* If `get_velocities=True` is passed to `._read`, but no velocities are stored in the TRR file, it raises a `RuntimeError` (similarly for forces).
* Support for all of this whether `atom_indices` are specified to `._read` or not.
* New static method `._check_has_velocities_forces` checks for presence of velocities/forces in a file (returning 2-tuple `(has_velocities, has_forces)`, each nonzero and positive if answer is yes.)
* `TRRTrajectoryFile._write` can optionally take `vel` and `force` (numpy arrays).
* New unit tests cover all this behavior.

Also some changes to make a few parts more readable on 80 columns, and shifted the zero-filled allocation of buffers (only used if `atom_indices` is not `None`) to only occur when `atom_indices` is not `None`.

mpharrigan:
rebased in #1294

Issue #1269, opened 2017-05-31
Hydrogen-bond calculations, excluding water-water but still including water-other


ahy3nz:
Hi,

I'm studying lipid bilayers and am interested in the water-lipid hydrogen bonding information. However, it seems like the current hydrogen bonding implementations either fully include or fully exclude all water molecules in the hydrogen bonding calculations. Computing hydrogen bonds while including the solvent becomes computationally expensive due to all the water-water pairs. Is there built in functionality to exclude water-water pairs, but still account for water-lipid pairs (donor-acceptor or acceptor-donor)? If not, would this be a worthwhile contribution (this could be useful for optimizing protein hydrogen bonding, too)?

PR #1273, opened 2017-06-12
Exclude water, but include water-solute hydrogen bonding

ahy3nz:
Addressing issue #1269. Main idea is to preserve the original logic with exclude_water, but add a scenario in which waters are largely excluded, but still look at water-solute or solute-water hydrogen bonding. This was implemented mainly by changing what constitutes the donor or acceptor set to now account for water. Comments/improvements would be greatly appreciated


mdtraj/geometry/tests/test_hbonds.py
```diff
@@ -145,3 +145,23 @@ def test_wernet_nilsson_1():
...
+    hbonds1 = md.baker_hubbard(t, exclude_water=True, include_water_solute=True) 
+    assert np.shape(hbonds1)[0] > 0 
+
+    # No water, no water-solute
+    hbonds2 = md.baker_hubbard(t, exclude_water=True, include_water_solute=False)
+    assert np.shape(hbonds2)[0] == 0
+
+    # Yes water, yes water-solute (water-solute is redundant)
+    hbonds3 = md.baker_hubbard(t, exclude_water=False, include_water_solute=True)
+    assert np.shape(hbonds3)[0] > 0
```
ctk3b: Could probably also assert that hbonds3 contains more than hbonds1?

ahy3nz: Good point, adding that

mdtraj/geometry/tests/test_hbonds.py
```diff
@@ -145,3 +145,24 @@ def test_wernet_nilsson_1():
             # to make sure the criterion is giving back totally implausible stuff
             if len(hbonds) > 0:
                 assert np.all(md.compute_distances(t[frame], hbonds[:, [0,2]]) < 0.5)
+def test_water_solute():
```
tcmoore3: Should you test or donor-acceptor selections? I assume you can manually calculate how long the donors and acceptors lists should be.

ahy3nz: For example, for water-solute, verify the number of water donors and number of solute acceptors? And then do the same thing for solute-water?

tcmoore3: Yeah.

mpharrigan:
Another note: try to maintain backwards compatibility by adding new optional arguments to the end so the api doesn't change w.r.t. positional args

Issue #1281, opened 2017-07-31
md.rmsd segmentation fault for large n_atoms


justinrporter:
Hi all,

I've observed a kind of odd behavior for `md.rmsd` calls when the size of the data grows very large, ~especially for large numbers of atoms~. When `md.rmsd` is called, there is a segmentation fault in `inplace_center_and_trace_atom_major` somewhere.

In my case, I have a protein with ~3000 atoms, and when the number of frames is somewhere between 10^5 and 10^6, I see this error. Although I'm giving a specific example here, I've tried this on several other inputs with similar results.

The code looks something like this:
```python
from glob import glob
import mdtraj as md

top = md.load('/home/shared/protein_fah/fah_input/prot_masses.pdb').top
trj = md.load(glob('/home/shared/protein_fah/RUN0/CLONE*/*.xtc')[0:200],
              top=top, atom_indices=top.select('name N or name C or name CA or name CB'))

print(trj.xyz.shape)

print("computing the RMSD")
r = md.rmsd(trj, trj[0])

print("THE RMSD", r)
```

gives:

```
(1071347, 3061, 3)
computing the RMSD
/act/sge/default/spool/node08/job_scripts/491875: line 18: 32578 Segmentation fault      (core dumped) python -u scripts/nonsense.py
```

I load the core dump in gdb and get a stack trace that looks like the following:

```
$ gdb $( which python ) --core core.1323
(gdb) bt
#0  inplace_center_and_trace_atom_major._omp_fn.0 () at /opt/rh/devtoolset-2/root/usr/lib/gcc/x86_64-CentOS-linux/4.8.2/include/xmmintrin.h:905
#1  0x00002ac56adb9d13 in ?? () from /home/jrporter/miniconda2/envs/std/lib/python3.5/site-packages/numpy/core/../../../../libiomp5.so
#2  0x00002ac56ad89b17 in ?? () from /home/jrporter/miniconda2/envs/std/lib/python3.5/site-packages/numpy/core/../../../../libiomp5.so
#3  0x00002ac56ad891c5 in ?? () from /home/jrporter/miniconda2/envs/std/lib/python3.5/site-packages/numpy/core/../../../../libiomp5.so
#4  0x00002ac56adba193 in ?? () from /home/jrporter/miniconda2/envs/std/lib/python3.5/site-packages/numpy/core/../../../../libiomp5.so
#5  0x0000003c91407aa1 in start_thread () from /lib64/libpthread.so.0
[...]

PR #1283, opened 2017-08-01
Fix precentering overflow error in center.c

justinrporter:
Per #1281, when `n_frames * n_atoms * 3` is greater than the maximum size of a signed 32-bit integer, an overflow error causes a segmentation fault in `mdtraj/rmsd/src/center.c`.

This PR fixes this by converting the offending `int` to a ~`size_t`~ `long long`.


justinrporter:
I think this is ready to go @mpharrigan

mpharrigan:
Thanks!

PR #1291, opened 2017-08-25
added chi5 angle computation

marscher:
sometimes one is literally interested in all dihedrals, which includes chi5 as well. cc @giopina


marscher:
ERROR: mdtraj.tests.test_nmr.test_shiftx2

marscher:
This is pretty much unrelated to my changes, right?

mpharrigan:
Yes, that's a known failure. Thanks for this! lgtm

marscher:
I think it is assumed by the indices getter that CHI_N_ATOMS is a 2d list. Sorry for that. BTW the dihedral test case depends on PyMOL, is this installed during the test session?

PR #1266, opened 2017-05-12
Refresh tests

mpharrigan:
- Switch to pytest (nosetests no longer maintained?)
 - use "fixtures" instead of `yield` (later deprecated by pytest)
 - minor improvements
 - start switching to relative imports
 - update travis to test on python 3.6 and numpy 1.12
 - remove test data from package
 - (related to above) move `get_fn` out of mdtraj proper
 - move geometry tests to be with the rest of the tests
 - port test_examples as well. TODO: fix some of these?
 - remove nose from conda recipe

The biggest changes are in test_trajectory.py. I changed it to programatically generate NxN tests for all formats. Sometimes these tests don't make sense due to the nature of the format

I've marked several things as xfail (expected fail). My intent is to open issues for "true" failures and eventually fix them. Things that aren't supported by the file formats themselves should be marked skip.

PR #1292, opened 2017-08-31
Debug Shiftx2

mpharrigan:
Can't reproduce this locally, so hacking travis to only do the nmr tests. builds on #1266


mpharrigan:
The seemingly salient error that I can't reproduce locally

```
***************************************************************

[Error in module]

	 angles: 139

***************************************************************

[SHIFTY+] Error: execution halted due to unexpected errors.                

  ->> SHIFTX2 is calculating with only SHIFTX+'s prediction results
```

mpharrigan:
Just going to disable this on travis.

PR #1294, opened 2017-09-01
 Extracting velocities/forces from TRR files (hidden API) - Rebase

mpharrigan:
This is a rebase of #1271 

I also modified the tests to work with pytest after #1266 

cc @dwhswenson


mpharrigan:
One thing I noticed is that  `_check_has_velocities_forces` opens and closes a file handle and then `__cinit__` opens a file handle. Not the end of the world, but probably could be re-ordered to just use the file handle we open anyways

mpharrigan:
I made this change in f386b4f to be consistent with how `_estimate_n_frames_from_filesize` works

Issue #1274, opened 2017-06-22
image_molecules() sometimes fails


CharlieLaughton:
Using MDTraj 1.8.0, image_molecules() sometimes cannot fix PBC jumps. Attached is a single-frame PDB format file (t783_orig.pdb) that has this problem for me.

```
>>> import mdtraj as mdt
>>> t = mdt.load('t783_orig.pdb')
>>> t.image_molecules()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/users/charlie/.local/lib/python2.7/site-packages/mdtraj/core/trajectory.py", line 1941, in image_molecules
    anchor_molecules = self.topology.guess_anchor_molecules()
  File "/users/charlie/.local/lib/python2.7/site-packages/mdtraj/core/topology.py", line 1139, in guess_anchor_molecules
    .format(atoms_cutoff))
ValueError: Could not find any anchor molecules. Based on our heuristic, those should be molecules with more than 2121 atoms. Perhaps your topology doesn't give an acurate bond graph?
```

I can fix this problem by editing line 1132 in topology.py from:
```
anchor_molecules = [mol for mol in molecules if len(mol) > atoms_cutoff]
```
to:
```
anchor_molecules = [mol for mol in molecules if len(mol) >= atoms_cutoff]
```
Then:
```
>>> import mdtraj as mdt
>>> t = mdt.load('t783_orig.pdb')
>>> t2 = t.image_molecules()
>>> t2.save('t783_imaged.pdb')       
```
But if you look at t783_imaged.pdb (attached) it is a worse mess than before.

The problem seems to me to be in image_molecules.pxi, lines 25-30:
```
 for k in range(3):
            delta[k] = frame_positions[atom2, k]  - frame_positions[atom1, k]
            offset[k] = frame_unitcell_vectors[2, k]*roundf(delta[2]/frame_unitcell_vectors[2,2])
            offset[k] += frame_unitcell_vectors[1, k]*roundf((delta[1]-offset[1])/frame_unitcell_vectors[1,1])
            offset[k] += frame_unitcell_vectors[0, k]*roundf((delta[0]-offset[0])/frame_unitcell_vectors[0,0])
            frame_positions[atom2, k] = frame_positions[atom2, k] - offset[k]
```
For example, the first pass through this loop has k=0, so only delta[0] is set in line 26, yet delta[2] is referenced in line 27, etc.

I think the correct logic is:
```
 for k in range(3):
[...]

PR #1298, opened 2017-09-08
Issue1274

CharlieLaughton:
Hi,

As requested, a PR for #1274 

Regards,

Charlie


mpharrigan:
Thanks so much for the PR! Looks like there are strange integer type errors on 64 bit windows. I can investigate these later when I have time. Please give me a bump if it seems I've forgotten about it

CharlieLaughton:
Will do. Sorry I can't help with this one - only have access to Mac and CentOS machines, good luck tracking it down!

CharlieLaughton:
Hi - that bump as requested...

PR #1308, opened 2017-11-03
Add Bond Order information to Topology

### LINKED ISSUE #1307: Adding Bond Orders to Topology
I'd like to add Bond Order information to the `Topology` class to preserve additional structural information when available. 

The reason for this and the discussion of what information is needed can be found in the [OpenMM issues discussing this topic](https://github.com/pandegroup/openmm/issues/1443) and we are wanting to put something like this in [YANK to enable SMARTS string searches](https://github.com/choderalab/yank/issues/801) without having to constantly flip from MDTraj topologies to OpenMM and back again.

This would require extending information stored in the `bond` to have additional fields, and likely spin it off into its own class like `Atom` or `Residue` is. My current plan is to use the same class as the [OpenMM Topology for bond](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/topology.py#L446) which would have a `type` and a `order` fields that follow the logic outlined [here](https://github.com/pandegroup/openmm/issues/1443#issuecomment-256101964). The `type` could take either `None`, `'Aromatic'`, or `'Amide'`, and the `order` would take `None`, `1`, `2`, or `3`, instead of the OpenMM classes of `Single`, `Double`, etc.

The readers and writers for a specific file type would need to be extended to store this information if the file type supported it, such as `mol2` files.

Thoughts on the utility of this feature, implementation plan? 

cc @jchodera

Lnaden:
This PR ports the [OpenMM bond order](https://github.com/pandegroup/openmm/pull/1668) representation into MDTraj. Implements the `Bond` class to Topology and updates the Mol2 reader to use bond_order field.

This does change the API a bit since bonds are now represented by a class instead of just two atoms. However, most backwards compatibility should be preserved as each `Bond` is subclass of the `namedtuple` so you can iterate by `for a,b in top.bonds` and still get a `Atom` class for `a` and `b`. This behavior is only different in the `to_dataframe` output (see below).

There are a couple of points which I implemented which I would like some additional feedback on.

1. Equality/Less than/Greater than of a Bond: Right now I only compare Atom index and not Type/Order since defining these does not change which atoms are bound. I could compare these, but it seemed redundant unless you wanted to define two bonds for the same pair of atoms? The only place I see this being used is in `Topology` equality checks.
2. `to_/from_dataframe` representation of `Bond`. I changed the output of this to float to represent the `Bond.type` as fractional for things like Aromatic and Amide, without having to have a numpy array of mixed type or structured arrays. Since this is the main API breaking component, it would be good to get some feedback on this implementation and if others have a better way of making this conversion.

Fixes #1307

### DIFF
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -1621,14 +1621,10 @@ class Bond(namedtuple('Bond', ['atom1', 'atom2'])):
             return False
         if {other[0], other[1]} != {self[0], self[1]}:
             return False
-        # Optional consideration, but this is more like metadata on the bond,
-        # If the above checks out, but the below does not, then there are
-        # now two bonds between the same atoms, which makes no sense
-
-        # if other.type != self.type:
-        #     return False
-        # if other.order != self.order:
-        #     return False
+        if other.type != self.type:
+            return False
+        if other.order != self.order:
+            return False
         return True
 
     def __repr__(self):
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -1619,8 +1619,8 @@ class Bond(namedtuple('Bond', ['atom1', 'atom2'])):
     def _equality_tuple(self):
         # Hierarchy of parameters: Atom1 index -> Atom2 index -> type -> order
         return (self[0].index, self[1].index,
-                self.order if self.order is not None else 0,
-                float(self.type) if self.type is not None else 0.0)
+                float(self.type) if self.type is not None else 0.0,
+                self.order if self.order is not None else 0)
 
     def __deepcopy__(self, memo):
         return Bond(self[0], self[1], self.type, self.order)
@@ -1646,10 +1646,31 @@ class Bond(namedtuple('Bond', ['atom1', 'atom2'])):
         # Set of atoms making up bonds, the type, and the order
         return hash((self[0], self[1], self.type, self.order))
 
-    def __gt__(self, other):
-        # Uses the total_ordering to handle all other comparators
+    @staticmethod
+    def _other_is_bond(other):
+        # Ensure other type for inequalities is a bond
         if not isinstance(other, Bond):
             raise TypeError("Bond inequalities can only be compared with other bonds")
+
+    def __gt__(self, other):
+        # Cannot use total_ordering because namedtuple
+        # has its own __gt__, __lt__, etc. methods, which
+        # supersede total_ordering
+        self._other_is_bond(other)
         return self._equality_tuple > other._equality_tuple
 
+    def __ge__(self, other):
+        self._other_is_bond(other)
+        return self._equality_tuple >= other._equality_tuple
+
+    def __lt__(self, other):
+        self._other_is_bond(other)
+        return self._equality_tuple < other._equality_tuple
+
+    def __le__(self, other):
+        self._other_is_bond(other)
+        return self._equality_tuple <= other._equality_tuple
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
 
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -54,7 +54,6 @@ import numpy as np
 import os
 import xml.etree.ElementTree as etree
 from collections import namedtuple
-from functools import total_ordering
 
 from mdtraj.core import element as elem
 from mdtraj.core.residue_names import (_PROTEIN_RESIDUES, _WATER_RESIDUES,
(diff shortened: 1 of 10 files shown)


Lnaden:
cc @andrrizzi @jchodera as this is a feature related to our projects.

Lnaden:
The Travis failures appear to be OSX image problems with Ruby version on Brew, I think the fix is just `brew update` at the start, but that should be its own PR, or wait for travis to update their build image. The Appveyor builds are not running but I don't know why, it appears PyYaml dependencies are failing in an infinite loop, but its unclear why they would have started failing now and not in older PR's.

Lnaden:
I modified the `test_pickle` of `test_topology.py` to test equality of a topology with and without bond order. However, there may be gaps in this test because its only checking the atom indices for `Bond.__eq__`, not order or type as combination of two topologies with bonds between the atoms, but differing `order` and `type` properties is ambiguous (see Point 1 of the original post). 

With respect to the original post, any feed back on the two outstanding points I asked about?

rmcgibbo:
For 1. it seems to me that either it should be illegal to define two bonds between the same pair of atoms, or the comparison operators ought to check the type and order. For 2, I think the slightly-breaking way you did the dataframe is fine.

mdtraj/core/topology.py
```diff
@@ -1621,14 +1621,10 @@ def __eq__(self, other):
...
-        # now two bonds between the same atoms, which makes no sense
-
-        # if other.type != self.type:
-        #     return False
-        # if other.order != self.order:
-        #     return False
+        if other.type != self.type:
+            return False
+        if other.order != self.order:
+            return False
```
rmcgibbo: I don't think you have a total ordering now, since you changed eq w/o changing gt and lt. two bonds that share the same atoms but have different type or order are neither greater than, less than, nor equal to one another under the current code.)

Lnaden: Under that condition (same atom indices, but different type/order), that should probably throw an error somewhere since that means there are two `Bonds` between the same pair of atoms (as you mentioned before). I can't think of a use case where that may be helpful, unless you are defining a type of additional restraint through the "bond" framework of whatever simulation package you are using.

 I'm not sure where there would be a good time to check for multiple bonds on the same atoms. The `add_bond` method could check to see if there is already a bond between atom1 and atom2 in the Topology._bonds list, but will that add too much overhead? The only way you could have 2 bonds on the same pair of atoms is if you manually call `add_bond` right now; I don't think any of the programatic ways (e.g. `from_openmm`, `join`) can cause this case. 

If you don't mind the overhead of doing a `for a,b in self._bonds` each time `add_bond` is called, I can add this in and then not have to worry about the inequality. Alternately, I can add an absolute order to `order` and `type` properties for `Bond.__lt__` and `Bond.__gt__`.

rmcgibbo: As I said before, I'm really fine with *either* it being illegal to have multiple bonds between the same pair of atoms or comparison operators that properly handle multiple bonds between the same pair of atoms. I wouldn't worry about the overhead of checking -- it's easy to optimize if slow.

rmcgibbo: you ought not to have to write _all_ of the comparators either -- [@functools.total_ordering](https://docs.python.org/3/library/functools.html#functools.total_ordering) can fill in most of them

Lnaden:
I've added in the full inequality check. It should catch all the corner cases. Although come to think of it, I could probably do it simpler by just making a tuple of `(atom1.index, atom2.index, type, order)` casting type and order to floats, then just doing inequality on the tuple... I'll fix that.

The test which keeps failing looks like something to do with one of the notebooks accessing a remote file, I don't think thats something I broke.

rmcgibbo:
Yeah, that test is something I was trying to fix in #1315, so don't worry about that.

rmcgibbo:
+1 for the tuple idea too. much easier.

rmcgibbo:
I think there might be more tests failing, e.g. starting on line 2928 of https://travis-ci.org/mdtraj/mdtraj/jobs/309785921#L2928.

Lnaden:
So it did, that is something I broken then. I'll figure that out and make a new push once my local tests pass

Lnaden:
A couple things:
* `namedtuple` already has `__lt__`, `__gt__`, etc., so `@total_ordering` did nothing, so I implemented all the logical tests one could do, which includes `!=`, `>=`, and `<=`. 
* I realized that if I want people to be able to call `add_bond` and pass a `Singleton` to `Type`, I had to make those available to the `__init__` of the package.

Lnaden:
Grain of salt on the tests passing, only the AppVeyor tests ran, travis never hooked into the commit (f8624c4)

rmcgibbo:
The topology tests run on windows, so I'm sure it's fine.

PR #1322, opened 2017-12-26
update the calculation of chi5

xianqiangsun:


marscher:
LGTM! I can confirm that all of the CHI atom name structures have to be 2d to be processed correctly. Strangely the tests didn't catch this.

@xianqiangsun can you also add the following in devtools/conda-recipe/meta.yaml:

```
tests:
   require:
    pymol # [py2k]
```

marscher:
test_dihedral should also contain a 2d chi5_atoms (is also 1d at the moment).

marscher:
LGTM

PR #1323, opened 2017-12-26
update __init__.py

xianqiangsun:
updated __init__.py in geometry to have chi5 angle calculated correctly


marscher:
you need to add this also in dihedral.py `"__all__"` to make it work.

marscher:
Ah I see, this was added in #1322

PR #1325, opened 2018-01-15
fixed typo in Kabsch-Sander paper

gph82:
Somebody has to do the heavy lifting here...;)


msultan:
Needs more test coverage i think. :)

PR #1331, opened 2018-02-17
[formats/xtc] use byte offsets between frames for stride if avail

marscher:
This should avoid a lot of IO for high stride arguments,
because unwanted frames are just seeked over.

This feature relies on the byte offsets, which have to be provided
or explicitly computed by accessing the offsets property once.

If no bytes offset is available, the current behaviour is untouched.

Added tests.


mpharrigan:
@cxhernandez can you port the doc building fixes to fix the ci?

marscher:
Thanks!

PR #1332, opened 2018-02-28
[TRR] added efficient stride handling

marscher:
This also changes how many frames will be returned when using the stride. Eg. if one requests 100 frames with stride 2, one will also get 100 frames offset by 2 steps. This is at least what the docstring said, but was not true until this change.


mpharrigan:
Great, thanks!

(test failure is still the doc thing)

PR #1334, opened 2018-03-05
Added support for prm7 topology file extension.

lohedges:
The .prm7 topology file extension appears to be widely used as an alternative to .parm7 within the AMBER community. I've added support for this.


marscher:
thanks

PR #1349, opened 2018-04-30
Compile geometry extension with OpenMP.

### LINKED ISSUE #1335: Shrake-Rupley Parallelization?
When I run the `shrake_rupley` algorithm, I don't see any benefit to using multiple cores, although it looks from `sasa()` in `sasa.cpp` that OpenMP should be being used to compute the SASA of frames in parallel.

I am measuring this by comparing walltime to used cpu time. For example, with `rmsd`:

```python
%time _ = md.rmsd(t, t)
CPU times: user 4.12 s, sys: 3.36 s, total: 7.48 s
Wall time: 243 ms
```

Here, CPU several times higher than the walltime, implying the use of many cores.

When I do the same with `shrake_rupley`:

```python
%time md.shrake_rupley(t[0:20], probe_radius=0.28, n_sphere_points=960*4)
CPU times: user 3min 11s, sys: 121 ms, total: 3min 12s
Wall time: 3min 12s 
```

Here, walltime and CPU time match, suggesting no parallelization.

Does anyone have any thoughts for how to track this down further?

justinrporter:
Adding OpenMP compiler flags to the geometry extension, fixes #1335.

Should this also be done for `format_extensions`? If you grep many of the `.c` files built by cython, there is an `#ifdef _OPENMP`/`#endif` block, but I don't see where parallelism is actually used... maybe someone more familiar with that code could comment? (@Lnaden @mpharrigan @marscher)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -214,9 +214,12 @@ def rmsd_extensions():
 
 def geometry_extensions():
     compiler.initialize()
-    compiler_args = (compiler.compiler_args_sse2 + compiler.compiler_args_sse3 +
-                     compiler.compiler_args_opt + compiler.compiler_args_warn)
+    compiler_args = (
+        compiler.compiler_args_openmp +
+        compiler.compiler_args_sse2 + compiler.compiler_args_sse3 +
+        compiler.compiler_args_opt + compiler.compiler_args_warn)
     define_macros = None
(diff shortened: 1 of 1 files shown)


mpharrigan:
Thanks!

PR #1350, opened 2018-04-30
Fix the Pickle in Element Virtual Site

### LINKED ISSUE #1346: Multiprocessing support?
It would be nice if there would be a way to add multiprocessor support to the `md.baker_hubbard()` function, as on very large structures it can be quite slow.

I tried to replicate the functionality by breaking the trajectory into smaller chunks, and then using Python's `multiprocessing` module along with the `Pool()` class. However, is that it uses the `pickle` module to save the `traj` object, but that object cannot be saved to disk by `pickle`.

```
PicklingError: Can't pickle (0, 'virtual site', 'VS', 0.0, 0.0): attribute lookup virtual site on mdtraj.core.element failed
```

Lnaden:
The simplest of solutions to the problem in #1346 where Pickle cannot handle the space in the Virtual Site's `Element.name` field of `"virtual site"`.

Fixes #1346

### DIFF
--- a/mdtraj/core/element.py
+++ b/mdtraj/core/element.py
@@ -239,7 +239,7 @@ def get_by_symbol(symbol):
 # The radii for Ions (Na, K, Cl, Ca, Mg, and Cs are based on the CHARMM27
 # Rmin/2 parameters for (SOD, POT, CLA, CAL, MG, CES) by default.
 
-virtual =        Element(  0,"virtual site","VS", 0.0, 0.0)
+virtual =        Element(  0,"virtual_site","VS", 0.0, 0.0)
 hydrogen =       Element(  1,"hydrogen","H", 1.007947, 0.12)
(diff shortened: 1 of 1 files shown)


justinrporter:
Awesome thanks for doing this, Levi!

I wonder, could we also add a check for this to the pickling test in [this test](https://github.com/mdtraj/mdtraj/blob/master/tests/test_topology.py#L144)?

Lnaden:
Possibly, but it would be hard to test every combination of objects which could be in the `Topology` object. It might be better to test each object for pickling through other tests individually and then some more functional tests for the compound objects like the one you posted.

I'll add a pickle test for the `Element` object though

Lnaden:
The windows build failure here appears to be the `conda` path getting mangled, but not because the tests themselves are failing

justinrporter:
@rmcgibbo do you think this is merge-ready?

PR #1352, opened 2018-05-09
[formats/dcd] handle stride like numpy

marscher:
This yields almost the same frames a it used to,
but at the end of the trajectory there were some
frames missing in the current impl.
In PyEMMA we rely upon mdtraj is handling the
stride like numpy.

Added hints on howto improve the reader further
(efficient seeking by byte offset)

PR #1357, opened 2018-05-17
Revert "[formats/dcd] handle stride like numpy"

marscher:
Reverts mdtraj/mdtraj#1352


marscher:
This was not really doing what I intended. Sorry.

PR #1371, opened 2018-07-21
Fix bug in TINKER arc reader

leeping:
In the TINKER .arc reader, there was a bug that when the box information is read, the next line is read but it is not split into words, which leads to the box information being fed into atomic coordinates.  This PR should fix the issue.


leeping:
Any blockers to merging this?

rmcgibbo:
No, sorry for the delay.

leeping:
Thanks a lot!

PR #1378, opened 2018-09-10
Replace strip() with split() in `mol2.py`

rmatsum836:
I attempted to load a mol2 file containing Ti atoms using MDTraj, I would get the error: `KeyError: 'T'`.  After doing some digging in `mol2.py`, I found that the reason for this is because of the following line: `atoms_mdtraj["element"] = atoms.atype.apply(lambda x: x.strip(".")[0])`.  Because strip() returns a string, this line returns the first letter of a two letter element symbol.

As a result, I think split() is a more appropriate method in this case since it returns a list.  For example, `'Ti'.split('.')` will return `['Ti', '']` and `'Ti'.split('.')[0]` will return `'Ti'` which is the desired result.

In addition to this change, I added a unit test to `test_mol2.py`.  By loading in `cl.mol2`, this test ensures that the correct element name is loaded in.

### DIFF
--- a/mdtraj/formats/mol2.py
+++ b/mdtraj/formats/mol2.py
@@ -99,7 +99,7 @@ def load_mol2(filename):
     # If this is a sybyl mol2, there should be NAN (null) values
     if atoms_mdtraj.element.isnull().any():
         # If this is a sybyl mol2, I think this works generally.
-        atoms_mdtraj["element"] = atoms.atype.apply(lambda x: x.strip(".")[0])
+        atoms_mdtraj["element"] = atoms.atype.apply(lambda x: x.split(".")[0])
 
     atoms_mdtraj["resSeq"] = np.ones(len(atoms), 'int')
     atoms_mdtraj["chainID"] = np.ones(len(atoms), 'int')
--- a/tests/data/cl.mol2
+++ b/tests/data/cl.mol2
@@ -1,0 +1,9 @@
+@<TRIPOS>MOLECULE
+*****
+ 1 0 0 0 0
+SMALL
+GASTEIGER
+
+@<TRIPOS>ATOM
+      1 CL         -1.2092    0.7047    0.0000 Cl      1  UNL1        0.0000
+@<TRIPOS>BOND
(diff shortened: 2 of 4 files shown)


mattwthompson:
Looks like it's failing while attempting to infer elements from a [mol2](https://github.com/mdtraj/mdtraj/blob/e41a92299990fa6586a8f35739653a1801e33c31/tests/data/li.mol2#L9) file I added a while ago. The column for `atom_type`, which normally expects a SYBYL atom type ([these](http://www.cmbi.ru.nl/samsam/tutorials/studs/mopac/force_field3.html)?) like `H`, `C.ar`, etc. from which an element can be inferred, is a string that doesn't fit that definition. Maybe this makes it too improper to be considered a mol2 file, in which case this is a headache that foyer should deal with.

P.S. not sure how that unit test was passing before, maybe it inferred `opls_404` as oxygen?

rmatsum836:
I think you're right, the unit test as passing previously because it as interpreting `opls_404` as oxygen, since `'opls_404'.strip('.')[0]` returns 'o'.  

Since Foyer writes the atom type names to the SYBYL atom type column, this seems like an issue to deal with in Foyer to ensure compatibility with mdtraj's mol2 reader

mattwthompson:
I think we should just change the `li.mol2` to have its atom type simply be `Li` (https://github.com/mdtraj/mdtraj/commit/1ead4ffca7df900dfc0ffe293306c06cdb6ea7b7). In foyer can probably think about writing mol2 files with a `element.atom_type` sort of formatting.

rmatsum836:
In `li.mol2` I just changed the OPLS atom type to `Li`, in reference to the last comment.

mattwthompson:
Looks like an unrelated test is failing on Python 2.7 only.

mattwthompson:
I'm not sure who bug about this with a tag - @rmcgibbo @mpharrigan @cxhernandez ?

rmcgibbo:
I just merged another PR and it looks like there's a merge conflict now?

rmatsum836:
Merge conflict is resolved.

rmcgibbo:
Great. Sorry for the delay in looking at these.

PR #1381, opened 2018-09-21
fix typo in docs

smsaladi:
In the docs for `compute_contacts`, the first return value is referred to as `distances`, but then in the description of `residue_pairs`, the text refers to `distance`. If I understand correctly, this should actually be `distances`


marscher:
thanks! Those commits should contain the tag [ci skip] in their messages, because CI checking typo fixes is a waste of public resources.

Issue #1379, opened 2018-09-11
Issue with bond order information in mol2 reader


rmatsum836:
In reference to #1308, I'm getting a `KeyError: '1'` when trying to load in a mol2 file that contains no amide or aromatic bonds.  The reason for this being that `bond_type_map` is expecting strings to be passed when it's being passed a `np.int64`.  I believe this issue may have gone unnoticed since if the mol2 file contains an amide or aromatic bond, the single, double and triple bonds are passed into `bond_type_map` as strings.  

I think this issue could be solved simply by setting the `bond_value` as a string, such as:
`bond_augment[:, 0] = [float(bond_type_map[str(bond_value)]) for bond    _value in bonds["bond_type"].values]`.

If this method is acceptable, I will submit a PR with this fix and a unit test to catch this error.

mattwthompson:
This bug is easy to reproduce and can be done with any mol2 file in which the first bond is order `1`.

```(master *)$ python
Python 3.6.6 | packaged by conda-forge | (default, Jul 26 2018, 09:55:02) 
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import mdtraj as md
>>> md.load('/Users/mwt/software/mdtraj/tests/data/adp.mol2')
/anaconda3/envs/mosdef/lib/python3.6/site-packages/pandas/io/parsers.py:2230: FutureWarning: split() requires a non-empty pattern match.
  yield pat.split(line.strip())
/anaconda3/envs/mosdef/lib/python3.6/site-packages/pandas/io/parsers.py:2232: FutureWarning: split() requires a non-empty pattern match.
  yield pat.split(line.strip())
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mwt/software/mdtraj/mdtraj/core/trajectory.py", line 430, in load
    value = loader(filename, **kwargs)
  File "/Users/mwt/software/mdtraj/mdtraj/formats/mol2.py", line 122, in load_mol2
    bond_augment[:, 0] = [float(bond_type_map[bond_value]) for bond_value in bonds["bond_type"].values]
  File "/Users/mwt/software/mdtraj/mdtraj/formats/mol2.py", line 122, in <listcomp>
    bond_augment[:, 0] = [float(bond_type_map[bond_value]) for bond_value in bonds["bond_type"].values]
KeyError: 1
```

[...]

PR #1382, opened 2018-09-28
Ensure 'bond_value' is a string

rmatsum836:
In reference to #1379, `bond_value` is converted to a string which allows all mol2 files to be loaded in properly.

`test_load_all_mol2` unit test was added to ensure that every mol2 file in the `data` directory is properly loaded.


rmcgibbo:
LGTM.

Issue #1387, opened 2018-10-01
mol2 test failure on py27


marscher:
================================== FAILURES ===================================
```
________________________ test_load_all_mol2[adp.mol2] _________________________
mol2_file = 'adp.mol2', get_fn = <function _get_fn at 0x089200B0>
    @pytest.mark.parametrize('mol2_file', [('li.mol2'),
    ('lysozyme-ligand-tripos.mol2'), ('imatinib.mol2'),
    ('status-bits.mol2'),('adp.mol2')])
    def test_load_all_mol2(mol2_file, get_fn):
>       trj = md.load_mol2(get_fn(mol2_file))
tests\test_mol2.py:156: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
..\_test_env\lib\site-packages\mdtraj\formats\mol2.py:130: in load_mol2
    top = Topology.from_dataframe(atoms_mdtraj, bonds_mdtraj)
..\_test_env\lib\site-packages\mdtraj\core\topology.py:510: in from_dataframe
    a = Atom(atom['name'], elem.get_by_symbol(atom['element']),
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
symbol = 'O3'
    def get_by_symbol(symbol):
        s = symbol.strip().upper()
>       return Element._elements_by_symbol[s]
E       KeyError: 'O3'
..\_test_env\lib\site-packages\mdtraj\core\element.py:230: KeyError
=========================== short test summary info ===========================
```

marscher:
@rmatsum836 Your introduced test uses 'O3' as element name, which suprisingly only raises this error on Python 2.7. Can you please have a look again?

mattwthompson:
I think `O3` is not a valid SYBYL atom type, so changing it to `O.3` would make it more proper. But I'm not entirely sure because I don't work with SYBYL and I'm not sure which google result is the correct documentation.

marscher:
On 10/1/18 5:00 PM, Matt Thompson wrote:
> I think |O3| is not a valid SYBYL atom type, so changing it to |O.3| 
> would make it more proper. But I'm not entirely sure because I don't 
> work with SYBYL and I'm not sure which google result is the correct 
> documentation.

[...]

PR #1388, opened 2018-10-01
Prep py37, some bugfixes

marscher:


marscher:
the remaining test failures are related to the example notebooks and the mol2 test file see #1387

PR #1392, opened 2018-10-09
allow RMSD calls when ref_atom_indices and atom_indices are used

### LINKED ISSUE #1304: Overly restrictive RMSD sanity checks?
In doing a number of comparisons of trajectory data with structures from the PDB I bumped into a reoccurring problem with `md.rmsd`. I am getting ValueError exceptions if I provide selections (in both ref and target) which are not consistent with the other topology: fx. as selection A in topology X should be valid also for a topology Y and vice versa with a selection B in topology Y.
I've identified the follow checks in `rmsd/_rmsd.pyx` which I find may be overly restrictive (all in `rmsd(...)`):
` (atom_indices < reference.xyz.shape[1])`
i.e. all selected atom (target) indices should be within the range of the number of atoms in the reference (this is only the case if either selection is not given as argument)
` (ref_atom_indices < target.xyz.shape[1]) *`
i.e. all selected atom (reference) indices should be within the range of the number of atoms in the target (same as above)
` if not (target.xyz.shape[1] == reference.xyz.shape[1]):`
i.e. target and reference must have same number of atoms (this check should be done following the atom selection)

I think it is fairly straight-forward to resolve this, but before I proceed to change this and do a PR I'd like to consult with you to see if I'm overlooking something here.

sunhwan:
fixes #1304

### DIFF
--- a/mdtraj/rmsd/_rmsd.pyx
+++ b/mdtraj/rmsd/_rmsd.pyx
@@ -128,8 +128,6 @@ def rmsd(target, reference, int frame=0, atom_indices=None,
     # import time
     cdef bool atom_indices_is_none = False
 
-    # Error checks
-
     if atom_indices is None:
         atom_indices_is_none = True
         atom_indices = slice(None)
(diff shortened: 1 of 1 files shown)


marscher:
@psolsson can you please confirm that this fixes your problem?

you could install this patch like this:
```
pip install https://github.com/sunhwan/mdtraj/archive/760a943491dc9fdaf58109dcbe80afe6a6dc4baf.zip
```

@sunhwan Actually it would be best to have another test case for it, but it is kind of a minor change so we can skip this, when @psolsson confirms that it works as well.

PR #1401, opened 2018-11-14
Notebook tests: `from __future__` must come first

dwhswenson:
A few tests are failing because recent versions of Jupyter strictly enfore the rule that `from __future__` imports must be the first line in a cell (see, for example, the Py-3.6 test [here](https://travis-ci.org/mdtraj/mdtraj/jobs/454545185#L3469)). These notebooks had cells like:

```python
%matplotlib inline
from __future__ import print_function
```

This PR just switches the order of those first two.

### DIFF
--- a/examples/principal-components.ipynb
+++ b/examples/principal-components.ipynb
@@ -15,8 +15,8 @@
    },
    "outputs": [],
    "source": [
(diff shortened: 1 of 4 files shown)


marscher:
LGTM! Thanks!

Issue #1340, opened 2018-04-09
nosetests may have some bug 


autodataming:
if the function have get_fn parameter, the test was failed.  nose may cannot recognize conftest.py file. can you give me some advice?

**nosetestsanapy2.exe  mdtraj -v**
ERROR: mdtraj.tests.test_zipped.test_write_bz2
----------------------------------------------------------------------
Traceback (most recent call last):
  File "D:\anaconda2py2764\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
TypeError: test_write_bz2() takes exactly 1 argument (0 given)

----------------------------------------------------------------------
Ran 484 tests in 26.330s

FAILED (errors=317)

**pytestanpy2.exe  -v  .\tests\**
-- Docs: http://doc.pytest.org/en/latest/warnings.html
 9 failed, 925 passed, 106 skipped, 63 xfailed, 31 warnings in 792.96 seconds =

**nosetestsanapy2.exe  -v .\tests\**
======================================================================
ERROR: mdtraj.tests.test_zipped.test_write_bz2
----------------------------------------------------------------------
Traceback (most recent call last):
  File "D:\anaconda2py2764\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
TypeError: test_write_bz2() takes exactly 1 argument (0 given)

----------------------------------------------------------------------
Ran 484 tests in 15.880s

marscher:
the documentation is outdated, one should now use pytest to run the test suite.

PR #1403, opened 2018-11-14
Add XTCReporter for OpenMM 

cescgina:
Add an XTCReporter for OpenMM with support for atom subsets and append mode.

### DIFF
--- a/mdtraj/formats/xtc/include/xdr_seek.h
+++ b/mdtraj/formats/xtc/include/xdr_seek.h
@@ -12,5 +12,6 @@
 
 int64_t xdr_tell(XDRFILE *xd);
 int xdr_seek(XDRFILE *xd, int64_t pos, int whence);
+int xdr_flush(XDRFILE* xd);
 
 #endif
--- a/mdtraj/formats/xtc/src/xdr_seek.c
+++ b/mdtraj/formats/xtc/src/xdr_seek.c
@@ -48,3 +48,8 @@ int xdr_seek(XDRFILE *xd, int64_t pos, int whence)
 
 	return exdrOK;
 }
+
+int xdr_flush(XDRFILE* xdr)
+{
+    return fflush(xdr->fp);
+}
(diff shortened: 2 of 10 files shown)


marscher:
Thanks for your contribution - this looks pretty useful. I think the newly provided test will not be executed, because we do not install openmm in the test-environment (maybe we should change that?).

marscher:
openmm is present in most testing environments. So we can rely upon CI here.

cescgina:
Great! I have a question about the tests, in the documentation it says that they should be run using nosetest, however I could not get them to work, but I could run them through pytest, is that how it's supposed to work and the docs are outdated or there is some problem in my installation? I had an error similar to issue #1340.

marscher:
thanks for pointing this out. I created another PR to fix the documentation pointing to pytest.

marscher:
@cescgina can I push to your branch?

cescgina:
Yes, no problem. Should I add you as a collaborator on my fork?

marscher:
you don't need to, just enable "enable edits from maintainers", which should be the default option, afaik.

cescgina:
Ok, found it, the option is active

marscher:
I do not understand why the test skipping does not work. It works locally (with negated condition :grin:) on pytest-4.0.0

marscher:
https://docs.pytest.org/en/latest/skipping.html#id1 it is even the documented semantics...

cescgina:
Thanks @marscher! Now it works perfectly.

marscher:
On 20.11.18 11:18, Joan Francesc Gilabert wrote:
> Thanks @marscher! Now it works perfectly.
> 
You're welcome. However I did not figure out why the test skipping on
windoze does not work (on the other platforms). The used semantics are
exactly the same as in the other testing units.

marscher:
For now I could only spot failures due to the notorious MOL2 thing and the notebook examples. Anybody else want to review this?

PR #1405, opened 2018-11-14
Add compute_center_of_geometry

mattwthompson:
Seems to be low-hanging fruit, sometimes we deal with non-element "sites," i.e. coarse-grained beads. In my case I happen to be using LAMMPS files for which reading masses is cumbersome and a center of geometry is good enough.


marscher:
should solvent be excluded by default when computing this?

mattwthompson:
In my opinion, anything like that should be handled by the user, given how nicely mdtraj selects and slices atoms out.

mattwthompson:
Looks like the test I added is passing, failed tests I think are only related to this mol2 mess

PR #1414, opened 2018-12-29
add RMSF analysis

### LINKED ISSUE #1210: RMSF function
I had to do some RMSF (Root-Mean-Square Fluctuation, definition: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2830444/#fd6) calculations recently, thought a function to do this might be useful. 

Related to #1188 because taking the expression for RMSD @jchodera gives there, RMSF is simply given by a different mean:

``` python
trajectory.superpose(reference, atom_indices=atom_indices1, ref_atom_indices=ref_atom_indices1)

rmsf = np.sqrt(3*np.mean((trajectory.xyz[:, atom_indices2, :] - reference.xyz[:, ref_atom_indices2, :])**2, axis=(0,2)))
```
(or maybe there's another way to do this faster?) 

Do people think this would be useful to have as a convenience function?

sunhwan:
resolve #1210

### DIFF
--- a/docs/analysis.rst
+++ b/docs/analysis.rst
@@ -19,6 +19,7 @@ Root-mean-square deviation (RMSD)
     :toctree: api/generated/
 
     rmsd
+    rmsf
     lprmsd
     Trajectory.superpose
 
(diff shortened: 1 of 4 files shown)


rmcgibbo:
This seems like a good idea.

PR #1416, opened 2019-01-10
Fix appveyor

mattwthompson:
I forgot exactly the github thread I found this fix from, but it worked for us in https://github.com/mosdef-hub/mbuild/pull/477 so it should work here

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,10 +10,10 @@ environment:
     - PYTHON: "C:\\Miniconda36-x64"
       CONDA_PY: "27"
       ARCH: "64"
-    - PYTHON: "C:\\Miniconda3-x64"
+    - PYTHON: "C:\\Miniconda36-x64"
       CONDA_PY: "36"
       ARCH: "64"
-    - PYTHON: "C:\\Miniconda3-x64"
+    - PYTHON: "C:\\Miniconda36-x64"
       CONDA_PY: "37"
       ARCH: "64"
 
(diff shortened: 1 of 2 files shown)


rmcgibbo:
Should I merge this? It looks like it's failing, but idk if that means its succeeding and just failing farther into the build?

mattwthompson:
Probably not good to merge. I have a few other things I can try.

> succeeding and just failing farther into the build?

Right, previously it was failing to build the package at all, now it's able to build it but fails some tests later on.

justinGilmer:
If I recall correctly, we were running into this issue as well. The `AttributeError` below is due to the plugin `pytest-ignore-flaky`. Some recent update causes it to fail.
image


We migrated our `flaky` tests to using the decorator `@pytest.mark.xfail(strict=False)`
This way, if the tests pass, its marked as `xpass` and the test suite continues, if they fail, they are marked as `xfail` and the tests continue as well.

Links to this feature in `pytest` [here](https://docs.pytest.org/en/latest/skipping.html#xfail-mark-test-functions-as-expected-to-fail)

mattwthompson:
Looks like the only failing test at the moment is this one, and only on Python 3.x

```tests/test_rmsd.py::test_trajectory_rmsf_aligned FAILED```

rmcgibbo:
@sunhwan -- that's a test you just committed, correct? could you take a look at it?

rmcgibbo:
@mattwthompson: i'll merge this, since it seems to fix the infra (thanks!) and hopefully we can revert or fix the rmsf in a separate PR.

PR #1423, opened 2019-01-17
Add distance calculations between different times

mattwthompson:
I'm not sure if this would be useful to any other MDTraj users, but I'm computing the Van Hove function for some systems and am partway through adding it here. It's basically a transient RDF, [ref](http://advances.sciencemag.org/lookup/doi/10.1126/sciadv.1603079)

image

and as far as I can tell, no existing functions allow for simple distance calculations within a trajectory across different times. I only need to compare to time 0 but this should be adaptable for arbitrary time by looking elsewhere in the C array.

This is messy and ugly right now, but I have it working for some simple systems. I thought I'd open it now in case anybody finds this notably useful or objectionable.


mattwthompson:
Not sure what all goes into reviewing a feature of this size, but it's been working for us in production for many months now. Most of this code is copied from existing RDF-related functions. 

We are also working on 2-D RDFs, which may seem silly for some biophysics but are useful in studying some confined systems involving 2-D materials.

rmcgibbo:
I'm happy to let you merge this in at your leisure. Maybe some of the information your first comment on this PR could go into the main docstring, but I think it looks fine and I trust you :)

Issue #1422, opened 2019-01-17
resSeq attribute not set by mdtraj.Topology.from_openmm()


JBGreisman:
Hi -- 

For system prep I have been using Modeller in OpenMM, and I have been using `mdtraj.Topology.from_openmm()` occasionally to use atom selections. However, it seems that the `Residue.resSeq` attribute is not set when building the Topology with `mdtraj.Topology.from_openmm()`. 

Is it possible to change this line in `from_openmm()` in mdtraj/core/topology.py:
```
...
for residue in chain.residues():
    r = out.add_residue(residue.name, c)
...
```
to the following
```
...
for residue in chain.residues():
    r = out.add_residue(residue.name, c, residue.id)
...
```
?

The `Residue.resSeq` attribute seems to be set when I load a pdb directly in mdtraj using `load_pdb()`, so it doesn't seem like this change is likely to have side effects. Please let me know if you think there's anything I'm missing here.

rmcgibbo:
If you proposed a pull request with that change, I'd merge it.

JBGreisman:
Closing issue: resolved in pull request: #1424

PR #1424, opened 2019-01-18
Use and set resSeq attribute in Topology.to_openmm() and from_openmm()

JBGreisman:
This is a small change to `mdtraj.Topology` in order to preserve residue numbering during Topology conversions with `to_openmm()` and `from_openmm()`. 

`to_openmm()` is modified to pass along the `resSeq` info to be set as the `id` for Openmm Residues, and vice versa for `from_openmm()`. 

This is a follow-up to Issue #1422

### DIFF
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -389,7 +389,10 @@ class Topology(object):
         for chain in value.chains():
             c = out.add_chain()
             for residue in chain.residues():
-                r = out.add_residue(residue.name, c, resSeq=residue.id)
+                try:
+                    r = out.add_residue(residue.name, c, resSeq=int(residue.id))
+                except ValueError:
+                    r = out.add_residue(residue.name, c)
                 for atom in residue.atoms():
                     if atom.element is None:
                         element = elem.virtual


JBGreisman:
Small follow-up note. There is a type issue lurking here -- `mdtraj.Topology.Residue` expects `resSeq` to be an int. However, `openmm.Topology` sets `id` as a string. As it stands, my pull request sets the `resSeq` to whatever the value of `id` is regardless of type. However, "residue" or "resSeq" atom selections require `resSeq` to be an int. 

One solution would be to always coerce `resSeq` to be an int, but that would fail if the `id` attribute is set to something non-numeric (I'm not sure when this would be useful). 

Alternatively, I could add an optional flag to `from_openmm()` such as `keep_id=False`. If this flag is set to True, I could raise a TypeError if the value cannot be converted to an int, and otherwise convert all the `id` values to ints. Please let me know if this seems like a worthwhile approach as it involves a minor change to the call signature of `from_openmm()`, although it doesn't change the default behavior.

rmcgibbo:
Which way of setting the types makes this feature most useful to you? It seems like casting it to an `int` might be more useful, so that can actually use the atom selections.

rmcgibbo:
Also, Hi Jack! I didn't recognize your username and put two and two together when responding to #1422.

JBGreisman:
Hey Robert -- I hope all's going well! 

I agree that casting to `int` is the most useful option in `from_openmm()` to enable atom selections. In my code I am currently casting them all before my call to `from_openmm()` but it would be cleaner to make it so that one doesn't need to think about types. Similarly, I think one should cast to `str` in `to_openmm()`. I've found at least one place in OpenMM that breaks if its `Residue.id` attribute is an `int` (`PDBFile.writeFile()` with `keepIds=True`).

My only concern is that wrapping my change in `int()` in `from_openmm()` will raise a ValueError if someone is using a non-numeric `Residue.id` value. Seeing as this "preserving intrinsic residue numbering" is already an opt-in feature in `OpenMM's PDBFile.writeFile()`, I figured it may make sense here as well.

rmcgibbo:
Either way is fine with me. Whichever you prefer. Just ping me when you want me to merge.

rmcgibbo:
Also, as you might imagine, I don't have a ton of time to contribute to open source these days, so if you're using the library heavily and would like to be added to the "team" and get a "commit bit", just let me know!

peastman:
In OpenMM, a Topology can be created from various sources: a PDB file, a PDBx/mmCIF file, an Amber prmtop file, etc.  The meaning of the `id` field varies between formats.  PDB files require the residue ID to be a number.  So if you created your Topology by loading a valid PDB file, it will always be a number.  If you have a Topology where the residue IDs are *not* numbers, and if you use PDBFile to write it to a file and specify `keepIds=True`, the result will not be a valid PDB file.

On the other hand, PDBx/mmCIF files do not require it to be a number (see http://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_atom_site.auth_seq_id.html).  So if you create a Topology by loading one, `id` could be an arbitrary string.

Here's what I suggest.  `from_openmm()` should try to cast the ID to an integer.  If it works, record it in the `resSeq` field.  If it throws an exception, just ignore it and don't try to set `resSeq`.  `to_openmm()` should always convert the value to a string, since that's what OpenMM expects.

JBGreisman:
Thanks for the suggestions -- I pushed changes that always convert the resSeq value to a string in `to_openmm()`. I attempt to cast the `id` to an integer to set `resSeq` and disregard it if it throws a ValueError in `from_openmm()`.

JBGreisman:
Hey @rmcgibbo, I think this is all ready to go now. The changes are minor, but they ensure consistency between OpenMM's Residue.id and MDTraj's Residue.resSeq where it's possible. 

Thanks!

PR #1407, opened 2018-11-27
[formats/mol2] add more checks to element parsing

marscher:
In DataFrame.atype.apply we get a list for Python2, but a
str for Python3 - so we need to distinguish this.

Moreover we need to check for orbital hybridization ($E.sp_x)
and 'weird' element names, which are not specified in SYBL nor GAFF,
e.g. CT, OS, CK etc.
The latter is ensured via trying to parse the element first with its
two letter code (Li etc.), if that fails, fall back to a one letter
code. This succeeds at least for the provided test examples.

Fixes #1387

PR #1426, opened 2019-01-22
Avoid infinite recursion error in mol2 parser

mattwthompson:
In #1407 an little function was added to try and more intelligently parse the various atomtypes that can be included in a mol2 file. However, it can hit an infinite loop if the string and its first element each do not constitute an element. Currently trying to parse a bad element, i.e. `'L'` will cause a recursion error from repeatedly calling itself and hitting the key error, whereas this change will cause it to simply fail out with a key error without hitting the recursion limit. This shouldn't affect any existing behavior since the other checks in the function already happened.

cc @marscher


marscher:
Good catch! Thanks you.

PR #1437, opened 2019-03-09
[dcd] read physical time

### LINKED ISSUE #1436: Time in .DCD files
The documentation of `mdtraj.load_dcd()` states, that .DCD files contain time information, but the `DCDTrajectoryFile.read()` method does only return `xyz`, `cell_lengths` and `cell_angles`.
Is there way to extract the time information via the `DCDTrajectoryFile` class?

marscher:
Fixes #1436


marscher:
the write impl in dcdplugin.c calculates the current time step statically...

PR #1434, opened 2019-02-13
adopted to pep518 dependency handling

marscher:
fixes #1433

PR #1441, opened 2019-04-25
updates setup/build requirements for pip install

salotz:
Previously the handling of build/setup requirements was foregone
because setuptools couldn't handle this correctly. This seems to now
be cleared up and the inclusion of PEP 517 and implementation in pip
allows for specification of build requirements in the pyproject.toml
file which is included.

I can report that this works for me. Let me know if there is something else you would want from me.

The motivation for this is that I have mdtraj as a dependency (optional albeit) for another project, and want it to be installable through pip and I had trouble with mdtraj raising the errors in the setup.py even though I was declaring it's dependencies for it in my project.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,27 +19,32 @@ sys.path.insert(0, '.')
 from basesetup import (write_version_py, build_ext,
                        StaticLibrary, CompilerDetection)
 
-try:
-    import numpy
-    import Cython
-    if Cython.__version__ < '0.28':
-        raise ImportError
-    from Cython.Build import cythonize
-except ImportError:
-    print('-'*80, file=sys.stderr)
-    print('''Error: building mdtraj requires numpy and cython>=0.19
+import numpy as np
+from Cython.Build import cythonize
 
-Try running the command ``pip install numpy cython`` or
-``conda install numpy cython``.
 
-or see http://docs.scipy.org/doc/numpy/user/install.html and
-http://cython.org/ for more information.
 
-If you're feeling lost, we recommend downloading the (free) Anaconda python
-distribution https://www.continuum.io/downloads, because it comes with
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -19,27 +19,32 @@
...
-    import numpy
-    import Cython
-    if Cython.__version__ < '0.28':
-        raise ImportError
-    from Cython.Build import cythonize
-except ImportError:
-    print('-'*80, file=sys.stderr)
-    print('''Error: building mdtraj requires numpy and cython>=0.19
-
-Try running the command ``pip install numpy cython`` or
```
rmcgibbo: do we need to take these messages out? i imagine some people might still be using an older pip that doesn't support pep517, and maybe these messages would help them (maybe we should add a note here about upgrading pip instead of `conda install`ing numpy and cython.

salotz: Some sort of message could be helpful, but perhaps it belongs in the installation docs. I would reserve messages in the actual installation scripts to assume you know what you want and help you fix them rather than telling you to do it another way. This is what led to my confusion in the first place, since I was providing Cython in my requirements and the error was still being raised. Until I saw the way the error was implemented I thought it was a problem with my `setup.py`. 
However, the conditional logic for the raising of these messages needs to be changed, because this will always fail when you try to `setup.py install` with cython in the build dependencies. Maybe there is a flag for pep517 or a pip version we can check for.

marscher:
I think it would be sane to require at least the first pip version, 
which implements pep517.

marscher:
superseded by #1434

salotz:
As I mentioned in #1434 [here](https://github.com/mdtraj/mdtraj/pull/1434#issuecomment-489704339) ; this PR isn't fully superceded by it .

PR #1431, opened 2019-02-10
Reporters no longer override user request to write unit cell information

jchodera:
Addresses #1429 : `Reporter` classes no longer override user request to write unit cell information if `Topology` is missing unit cell information.

PR #1443, opened 2019-05-04
Fix a test failure in test_reporters.py caused by merge of #1431

rmcgibbo:
cc @jchodera, #1431.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,3 +1,7 @@
 [pytest]
 testpaths = tests examples
 addopts = -vra
+filterwarnings =
+    ignore:.*Function restrict_atoms is deprecated;.*:DeprecationWarning
+    ignore:.*Using or importing the ABCs from 'collections' instead of.*:DeprecationWarning
+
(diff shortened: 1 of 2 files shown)


jchodera:
Thanks for catching this! 

If I understand correctly, the test needed to be fixed because the default behavior has now changed, and unit cells are no longer dropped in some cases.

Sorry about that---I had totally missed the travis test failure in #1431, though it looked like travis caught that failure in that PR:
https://travis-ci.org/mdtraj/mdtraj/jobs/491367893#L3731

rmcgibbo:
In this test, the OpenMM simulation is created with CutoffNonPeriodic. What _values_ end up in the cell_lengths / cell_angles, and are they actually meaningful?

jchodera:
The default periodic box values when creating a `System` would be stored, and no they aren't meaningful. 

Perhaps we should have two tests? One that uses a periodic system and one that uses a non-periodic system?

rmcgibbo:
Got it. This makes sense then.

rmcgibbo:
Yeah, more tests = usually good.

PR #1448, opened 2019-05-10
Fix egg info

### LINKED ISSUE #1447: setup.py egg_info requires numpy
This command is used by pip to determine the projects metadata like version. So it should not require Numpy at this stage, because dependent projects may not yet have created the build environment containing NumPy. This effectively blocks depending on mdtraj via a pep517 pyproject.toml file.

marscher:
Fixes #1447

### DIFF
--- a/basesetup.py
+++ b/basesetup.py
@@ -5,6 +5,8 @@ import json
 import shutil
 import subprocess
 import tempfile
+import warnings
+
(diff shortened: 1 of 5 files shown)


rmcgibbo:
@marscher: I tried to fix the travis failures here with https://github.com/marscher/mdtraj/pull/1

PR #1452, opened 2019-05-16
Remove bundled dependencies astor and pyparsing

marscher:
* Replaced bundled dependencies astor and pyparsing with external deps.
* fixed size of parsing cache to fit all elements of mdtrajs selection language (speed improvement).


rmcgibbo:
what is the benefit of not vendoring those packages? they add neglegably to the download size. with conda it's fine, but i'm still wary of the ergonomics of forcing users to install dependencies if they're coming through pip/setup.py

marscher:
But you started trusting pep517, didn't you? It works quite well. I do 
not like bundled things, because they never receive any bug fixes and 
bloat the repository. Close it if you don't like, but we should at least 
include the cache size fix.

rmcgibbo:
no, you're probably right. i just don't have a lot of experience with pep517 yet.

PR #1466, opened 2019-08-08
add missing type conversion in image_molecules before calling C code

fabian-paul:
Fixes "ValueError: Buffer dtype mismatch, expected 'int32_t'" but got 'long' in mdtraj.geometry._geometry.image_frame.

### DIFF
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -1965,7 +1965,7 @@ class Trajectory(object):
 
         if make_whole and sorted_bonds is None:
             sorted_bonds = sorted(self._topology.bonds, key=lambda bond: bond[0].index)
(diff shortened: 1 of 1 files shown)


poonampandey05:
Still getting the same error with mdtraj 1.9.3

marscher:
could you please share a piece of code triggering the error?

poonampandey05:
The code is as follow:
from __future__ import print_function, division
import mdtraj as md

t = md.load('d1le3.new_1le3.1.dcd', top='../step2_drude.psf')
protein = t.topology.guess_anchor_molecules()
newtraj = t.image_molecules(anchor_molecules=protein)

fabian-paul:
This bugfix isn't in the 1.9.3 release. I think you have to install from github.

PR #1467, opened 2019-08-08
Add gyration tensor calculation

KirillShmilovich:
Easy access to the gyration tensor has a number of useful properties related to [calculating a number of shape metrics of a system](https://isg.nist.gov/deepzoomweb/measurement3Ddata_help). With access to the gyration tensor calculation of the principal moments, axis, asphericity, acylindricity, and relative shape anisotropy can all be computed relatively easily.

This pull request adds `compute_gyration_tensor` to geometry/order.py with associated unit tests in `test_order.py` in much of the same style as `compute_inertial_tensor` is currently implemented. Calculations of the shape descriptions listed above can also be implemented, if you would like.

1. [x] Added `compute_gyration_tensor` and `_compute_gyration_tensor_slow` to `geometry/order.py`
2. [x] Added unit test to `test_order.py`

### DIFF
--- a/mdtraj/geometry/order.py
+++ b/mdtraj/geometry/order.py
@@ -361,3 +361,25 @@ def _compute_inertia_tensor_slow(traj):
         I_ab[n, 2, 0] = I_ab[n, 0, 2]
         I_ab[n, 2, 1] = I_ab[n, 1, 2]
     return I_ab
+
+def _compute_gyration_tensor_slow(traj):
+    """Compute the gyration tensor of a trajectory. """
+    xyz = traj.xyz
+    center_of_geom = np.expand_dims(xyz.mean(axis=0), axis=1)
+    centered_xyz = xyz - center_of_geom
+
+    S_nm = np.zeros(shape=(traj.n_frames, 3, 3), dtype=np.float64)
+    for n, xyz in enumerate(centered_xyz):
+        N = xyz.shape[0]
+        for r in xyz:
+            S_nm[n, 0, 0] += r[0] * r[0]
+            S_nm[n, 1, 1] += r[1] * r[1]
+            S_nm[n, 2, 2] += r[2] * r[2]
+            S_nm[n, 0, 1] += r[0] * r[1]
+            S_nm[n, 0, 2] += r[0] * r[2]
+            S_nm[n, 1, 2] += r[1] * r[2]
+        S_nm[n, 1, 0] = S_nm[n, 0, 1]
+        S_nm[n, 2, 0] = S_nm[n, 0, 2]
+        S_nm[n, 2, 1] = S_nm[n, 1, 2]
+        S_nm[n, :, :] /= N
+    return S_nm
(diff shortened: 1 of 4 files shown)


mdtraj/geometry/order.py
```diff
@@ -189,6 +189,30 @@ def compute_inertia_tensor(traj):
     return A * eyes - B
 
 
+def compute_gyration_tensor(traj):
+    """Compute the gyration tensor of a trajectory.
+
+    For each frame,
+        S_{nm} = sum_{i_atoms} r^{i}_m r^{i}_n
```
rmcgibbo: I think if you format this a bit differently, you could get it to show up in proper latex on the mdtraj.org website. e.g.: https://github.com/mdtraj/mdtraj/blob/master/mdtraj/geometry/internal.py#L373 or https://github.com/mdtraj/mdtraj/blob/e41a92299990fa6586a8f35739653a1801e33c31/mdtraj/geometry/hbond.py#L151

mdtraj/geometry/order.py
```diff
@@ -189,6 +189,30 @@ def compute_inertia_tensor(traj):
...
+
+    Parameters
+    ----------
+    traj : Trajectory
+        Trajectory to compute gyration tensor of.
+
+    Returns
+    -------
+    S_xy:  np.ndarray, shape=(traj.n_frames, 3, 3), dtype=float64
```
rmcgibbo: this uses `xy` but above you used `m` and `n`.

mdtraj/geometry/order.py
```diff
@@ -189,6 +189,30 @@ def compute_inertia_tensor(traj):
...
+        Gyration tensors for each frame.
+
+    """
+    xyz = traj.xyz
+    center_of_geom = np.expand_dims(compute_center_of_geometry(traj), axis=1)
+    xyz -= center_of_geom
```
rmcgibbo: i think this may change the coordinates in the trajectory object itself, which might be unexpected for users who wouldn't expect `compute_gyration_tensor` to modify the trajectory in place.

mdtraj/geometry/order.py
rmcgibbo: what about also adding a link to the url of the nist.gov page you mentioned in the PR? that has a nice, concise, description.

KirillShmilovich: Done. I added a reference with the link in the doc string.

mdtraj/geometry/order.py
KirillShmilovich: Changed all to use `x` `y`.

mdtraj/geometry/order.py
KirillShmilovich: Yup, I think this would would perhaps unwantingly modify the coordinates. Changed to more closely reflect how `compute_inertial_tensor` is currently written.

rmcgibbo:
lgtm. thanks!

PR #1482, opened 2019-09-24
Fix `read_trr_natoms` leaving malformed file open

### LINKED ISSUE #1481: Malformed files left open (TRR, but possibly elsewhere)
If you open a malformed TRR file, MDtraj leaves the file open (at least, with macOS 10.14.6). This is a problem: we're trying to use MDTraj's TRR reading to read a file while it is being written. Occasionally we try to read something that is malformed (in the middle of being written). We catch the `OSError` and handle it in our code (wait and try again). But this can happen often enough that we hit the OS-imposed limit on open files, since MDTraj leaves the files open.

I suspect this may apply to other file types as well (at least, probably XTC).

### Example

Create a (very) malformed "TRR file":
```bash
$ echo "foo" > malformed.trr
```

Run the following little script:

```python
#!/usr/bin/env python
import mdtraj
import psutil

print("Before: " + str(psutil.Process().open_files()))

try:
    mdtraj.formats.TRRTrajectoryFile("malformed.trr")
except OSError as e:
    print("OSError: " + str(e))

print("After: " + str(psutil.Process().open_files()))
```

Output:

```
Before: []
OSError: Malformed TRR file. Number of atoms <= 0. Are you sure this is a valid GROMACS TRR file?
After: [popenfile(path='/Users/dwhs/tmp/mdtraj_malformed_trr/malformed.trr', fd=4)]
```

### Workaround/Fix (Mac/Unix)

One workaround would be for the user to change the max number of files with `ulimit -n`, but that's not very user-friendly.

Here's an outline of the workaround I've implemented for us:

```python

open_files = psutil.Process().open_files()  # get open files owned by current process

def filter_criterion(popenfile):
    ... # client code for identifying the files that are relevant, return bool

files_to_close = [p for p in open_files if filter_criterion(p)]  # filter
for p in files_to_close:
    open(p.fd).close()  # I think Windows needs something else here
```

My understanding is that Windows works differently with file descriptors (namely, that there's only one per file), so maybe `open(p.path).close()` would work? Or maybe this isn't a problem in Windows? I don't have a Windows box to try stuff on.

If you wanted to add this to MDTraj, the `filter_criterion` would just check whether the path `popenfile.path` matches the input file. The downside is that you'd need to add psutil as a requirement. Otherwise you leave it to users to fix the problem on their end if it causes them problems -- I hope this issue will help them!

dwhswenson:
Fixes #1481 (at least, once it is done I hope it will).

I'll also check a few other file formats (at least XTC) to see if this repeats elsewhere before calling it ready.

@rmcgibbo : For the moment, I added `psutil` to the testing requirements in the conda recipe (test requires it), but also allowed it to skip with a `pytest.importerskip` if `psutil` isn't available. Other options would be to not include `psutil` in the test reqs and skip it, or to not allow it to be skipped. Let me know if you'd prefer a different approach.

### DIFF
--- a/tests/test_trr.py
+++ b/tests/test_trr.py
@@ -219,16 +219,17 @@ def test_ragged_2():
             f.write(xyz)
 
 
-def test_malformed_trr():
+def test_malformed():
     with open(temp, 'w') as tmpf:
-        tmpf.write("foo")  # verty badly malformed TRR
+        tmpf.write("foo")  # very badly malformed TRR
 
-    with pytest.raises(OSError):
+    with pytest.raises(IOError):
         TRRTrajectoryFile(temp)
 
     psutil = pytest.importorskip("psutil")
     open_files = psutil.Process().open_files()
-    assert open_files == []
+    paths = [os.path.realpath(f.path) for f in open_files]
+    assert os.path.realpath(temp) not in paths
 
 
 def test_tell(get_fn):
--- a/tests/test_trr.py
+++ b/tests/test_trr.py
@@ -219,6 +219,18 @@ def test_ragged_2():
             f.write(xyz)
 
 
+def test_malformed_trr():
+    with open(temp, 'w') as tmpf:
+        tmpf.write("foo")  # verty badly malformed TRR
+
+    with pytest.raises(OSError):
+        TRRTrajectoryFile(temp)
+
+    psutil = pytest.importorskip("psutil")
+    open_files = psutil.Process().open_files()
+    assert open_files == []
+
+
 def test_tell(get_fn):
     with TRRTrajectoryFile(get_fn('frame0.trr')) as f:
         eq(f.tell(), 0)
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -53,6 +53,7 @@ test:
     - pytest-ignore-flaky =1
     - pandas
     - pytables
+    - psutil
 
   source_files:
     - examples/*
--- a/mdtraj/formats/xtc/src/xdrfile_trr.c
+++ b/mdtraj/formats/xtc/src/xdrfile_trr.c
@@ -480,8 +480,10 @@ int read_trr_natoms(char *fn,int *natoms)
 	xd = xdrfile_open(fn,"r");
 	if (NULL == xd)
 		return exdrFILENOTFOUND;
-	if ((result = do_trnheader(xd,1,&sh)) != exdrOK)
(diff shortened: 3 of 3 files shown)


tests/test_trr.py
```diff
@@ -219,6 +219,18 @@ def test_ragged_2():
...
+def test_malformed_trr():
+    with open(temp, 'w') as tmpf:
+        tmpf.write("foo")  # verty badly malformed TRR
+
+    with pytest.raises(OSError):
+        TRRTrajectoryFile(temp)
+
+    psutil = pytest.importorskip("psutil")
+    open_files = psutil.Process().open_files()
+    assert open_files == []
```
marscher: What if pytest is writing to a XML results file while running this assertion? I think this has to be more specific.

marscher:
@dwhswenson Thanks for the provided fix. I thought Robert was talking about changing the position of setting of 'is_open' flag, so the handle will be closed properly. I would not start changing the c code - as long as you do not have a good reason for it, e.g. something is clearly a bug.

marscher:
Sorry I missed that part of the discussion. You're right, the function 
should clean its own mess :)

dwhswenson:
This now does what the PR title claims. It's ready for further review/merge.

A few related issues that might be worth addressing at some point, but perhaps separately:

1. XTC files won't have this problem, but that's because `read_xtc_natoms` also doesn't have the checks for malformed files. Someday, that should probably be added.

2. `xtc/src/*.c` has inconsistent indentation formatting. I assume this is because these are edited versions of the files from the Gromacs team, and looks like they use tabs and you use 4 spaces.

3. At some point, it might be good to switch the reported errors to the more specific ones available in Python 3 (e.g., `FileNotFoundError` where appropriate). Since these are subclasses of `OSError` (and `OSError is IOError` in Py3), it won't break backward compatibility (just need to set `FileNotFoundError = IOError` in Py2 compatibility). It would probably be best to do this for all formats at once, though.

4. The only testing error I'm getting now appears to be some issue with macOS/NumPy 1.7. I wonder if https://github.com/conda-forge/numpy-feedstock/issues/57#issuecomment-344650319 is to blame. Weirdly, there have been relatively recent Travis builds of MDTraj that claim to be macOS/NumPy 1.7 but clearly aren't -- in tests, you get deprecation messages issued by NumPy >=1.16. (e.g., https://travis-ci.org/mdtraj/mdtraj/jobs/541309840#L3087).

I'd rather not try to address all of these in this PR (keep the PR straightforward and single-topic), but it's worth documenting future thoughts for improvements.

rmcgibbo:
lgtm

PR #1484, opened 2019-10-04
Initial attempt at Azure pipelines

rmcgibbo:


rmcgibbo:
@marscher: What do you think about dropping Travis-CI and Appveyor and switching over to Azure Pipelines? I think I got it running this afternoon, and it has Linux, OS-X and Windows all in one service, and it seems like the Windows thing actually works.

rmcgibbo:
I propose the following:

1. Remove Travis-CI and Appveyor, and replace them with Azure pipelines. The reason is that travis-ci got bought, most of their staff got laid off, and the new owners do not have a good track record of supporting free software. See e.g. https://hynek.me/articles/simple-python-azure-pipelines/ Appveyor has always been horrible. Unifying Linux, OS-X and Windows CI under one roof should help.
2. Stop testing python 2.7. This is a volunteer project and I don't have a lot of time to contribute. I don't want to waste my time on python 2.7.

marscher:
Thank you very much for lifting the CI into the future! I totally agree with you to drop support for Python 2.7. Let the old snake rest in peace.
Do you mind, if I add a pip installation job to the build matrix, e.g. for only Linux?

rmcgibbo:
Adding pip install sounds good.

Issue #1485, opened 2019-10-17
Error loading in MOL2 files with various residues


rmatsum836:
I'm attempting to load a mol2 file into a trajectory with multiple residue names.  I am getting the following error:
```ValueError: All of the atoms with residue index 1 do not share the same residue name```

Tracing this error, I believe this is due to the way that `resSeq` and `chainID` are all set to one in lines 119-120 in `formats/mol2.py`. 

In `from_dataframe` in `topology.py`, there is a for loop starting at line 501 attempting to loop through the unique `resSeq`.  However since each atom has `resSeq` of 1, we end up checking if all of the residue names in the system are equal to a particular residue name, which will return False every time.   

The suggested fix here is to change how `resSeq` is defined in the atoms data frame so that each residue name has a unique `resSeq`.

rmcgibbo:
Nice find! Thanks so much. Would you be able to make that change (and add a test case) as a Pull Request?

rmatsum836:
Yes, I'll take a look this week.

PR #1490, opened 2019-10-28
Fix residue sequence in mol2 reader

rmatsum836:
This is a fix to #1485, where a ValueError was being raised when attempting to load in a mol2 file containing multiple residues.  Rather than setting `resSeq` in the atoms DataFrame to an array of 1s, it is set to the value of `code`.

I also had to check that `element` was valid in `mol2.py` by calling `elem.get_by_symbol`.  Now in `load_mol2`, the order is to first check if `element` from the atomtype names is valid.  If a KeyError is given, the next step is to try and get `element` from the atom names.

I added a system of water and acetonitrile for testing purposes, as well as an additional test to check that the correct amount of residues are in the trajectory.


rmcgibbo:
Wonderful.

Issue #1301, opened 2017-09-29
Error in wrapping serial numbers in write_gro()


tverho:
https://github.com/mdtraj/mdtraj/blob/8e85405586e549109f8865046867e1fda7f580d6/mdtraj/formats/gro.py#L407

In `formats/gro.py`, function `write_gro()`, the code says:

```
if serial >= 100000:
    serial -= 100000
```

when it should say

```
if serial >= 100000:
    serial %= 100000
```

As a result, gro files with more than 200000 atoms will become invalid.

tverho:
The bot labeled this as wontfix but this really is a trivial bug and only requires a change of one character to fix.

rmcgibbo:
Could you file a pull request with the change?

PR #1491, opened 2019-10-29
Fix issue #1301

tverho:
https://github.com/mdtraj/mdtraj/issues/1301

### DIFF
--- a/mdtraj/formats/gro.py
+++ b/mdtraj/formats/gro.py
@@ -405,7 +405,7 @@ class GroTrajectoryFile(object):
             if serial is None:
                 serial = atom.index
             if serial >= 100000:
-                serial -= 100000
+                serial %= 100000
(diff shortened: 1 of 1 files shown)


rmcgibbo:
Could you add a test case too?

tverho:
I don't have a development environment to run the tests for latest master. Cython 0.29 is required, I have older.

rmcgibbo:
Could you maybe attach and example file that fails with the current master branch but would work with the fix, if you have one? I can do the mechanics of testing it

tverho:
Here's a .gro file with 1M particles. I don't have a topology because I deal with united atom polymers (the particles are all the same type).

IIRC the problem shows so that the serial field becomes wider than in the specification after 200k particles leading to an invalid file.

[polymer_melt.gro.zip](https://github.com/mdtraj/mdtraj/files/3787458/polymer_melt.gro.zip)

PR #1494, opened 2019-11-06
Gsd reader/writer

ahy3nz:
This PR aims to introduce compatibility with the [HOOMD GSD file](https://gsd.readthedocs.io/en/stable/index.html). This requires an extra python package `gsd` (available on conda-forge) that provides a file layer and context manager for parsing trajectory, frame, and topological information from `gsd` files. 

"Residue" and "chain" information are not stored in GSD files, so the resultant topologies parsed from GSD files all belong to the same residue and chain, but bonds, atomic naming, and unitcell information are maintained and translated.

There is also functionality to write `mdtraj.Trajectory` into a `gsd` file, only translating particle names + positions, bonds, and unitcell information. 

Unit tests for I/O are included

My git commit history is weird but I think all the correct files-to-be-changed were changed and no extraneous files were unnecessarily modified


rmcgibbo:
People often install mdtraj via pip or from source rather than conda-forge, and generally the runtime requirements for other packages are pretty minimal and cover what most people have in a scientific python installation (numpy, scipy, etc).

Could you make the dependency on the external 'gsd' package optional?

ahy3nz:
Alright I moved any `import gsd` into the function routines. I think `gsd` is only in the `meta.yaml` test:requires, and I can uninstall `gsd` but still `import mdtraj` just fine

rmcgibbo:
Are there features that GSD supports (maybe GHMC simulations with a varying number of atoms) that MDTraj can't support? If so, what kind of error message would this throw, if any, if it's trying to read a file with one of those features?

ahy3nz:
Oh yes GSD can support variable topologies via HOOMD's compatibility with some MC methods. The `mdtraj.Topology` is built from the first frame, and any frame with a different topology from that first frame will raise `IOError`. For speed purposes, I was only comparing n_atoms and n_bonds within each frame (`gsd.hoomd.snapshot`) without going through all the work of creating a new `mdtraj.Topology` for each frame

Issue #1497, opened 2019-11-08
Inconsistency of 'NetCDFTrajectoryFile.read()' 'stride' parameter


padix-key:
To my knowledge, the `stride` parameter in almost all of the `read()` methods of *MDTraj* trajectory file classes does not decrease the number of outputted frames:

For example if I have a trajectory with 1000 frames and call
```python
file.read(n_frames=100, stride=2)
```
I get 100 frames (0, 2, ..., 196, 198).

But the `NetCDFTrajectoryFile` class is an exception:
If I call
```python
file.read(n_frames=100, stride=2)
```
I would get 50 frames (0, 2, ..., 96, 98)

marscher:
You're right, this should return as many frames as requested!

padix-key:
The problem is, that a slice is used. The slice selects only the *k*-th frame (*k* is the stride) of the *n* given  frames. Multiplying `n_frames` with `stride` should solve this. I prepared a PR for fixing this issue.

padix-key:
With the merged PR, this issue should be resolved.

PR #1501, opened 2019-11-14
Fixing 'stride' parameter for '.netcdf' files

padix-key:
Fixing #1497


rmcgibbo:
Could you add a test too, if you have any extra cycles?

padix-key:
I will have some spare time to add the tests probably next week

PR #1508, opened 2019-12-06
Thread parallelism for DRID

justinrporter:
As written, DRID calculations were not thread parallelized. Most of the work had already been done to do this, however, since each atom's DRID can be calculated in parallel, and dispatch to the `drid_moments` function is independent by atom.

This PR just adds openmp parallelization of the inner loop of the DRID calculation.

### DIFF
--- a/mdtraj/geometry/drid.pyx
+++ b/mdtraj/geometry/drid.pyx
@@ -28,7 +28,9 @@
 from __future__ import print_function, division, absolute_import
 import numpy as np
 from mdtraj.utils import ensure_type
+
 cimport cython
+from cython.parallel import prange
 
 __all__ = ['compute_drid']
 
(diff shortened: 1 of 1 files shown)


rmcgibbo:
Did you benchmark parallelizing over the outer loop vs over the inner loop?

rmcgibbo:
(I would have assumed going over the frames in parallel would be faster)

justinrporter:
Yes, I did--and I was surprised as well.

On my workstation and the system I chose (admittedly a rather large one), it was ever so slightly faster over the inner loop. Runtimes also seemed to be more stable run-to-run. Maybe it reduces cache misses or something? The effect was super minor though, on the order 1-2%.

rmcgibbo:
Sounds good then.

PR #1520, opened 2020-01-06
Update azure-pipelines-linux to python 3.8

rmcgibbo:
We generally support the 2 most recently released minor versions of python, so this drops testing of 3.6 and adds 3.8. If it works I'll expand the PR to the other OSes.


devtools/azure-pipelines-linux.yml
```diff
@@ -7,10 +7,10 @@ jobs:
   strategy:
     matrix:
       Python36:
```
sroet: small nitpick, but if you're changings the name on line 12, you might as well change this to `Python37:`
```suggestion
       Python37:
```

Issue #1509, opened 2019-12-17
python 3.8 incompatibility in selection


Olllom:
Python 3.8 requires an argument `posonlyargs` in `ast.arguments`.

The error message is:
``` 
 File "lib/python3.8/site-packages/mdtraj/core/topology.py", line 1048, in select
    filter_func = parse_selection(selection_string).expr 
  File "lib/python3.8/site-packages/mdtraj/core/selection.py", line 400, in __call__    
    compile(ast.fix_missing_locations(func), '<string>', mode='eval'),        
TypeError: required field "posonlyargs" missing from arguments
```

https://github.com/mdtraj/mdtraj/blob/f80d885d374a357a7badeefafc0d331de617bd52/mdtraj/core/selection.py#L394

Adding `posonlyargs=[]` here should fix the problem, see for example https://github.com/xonsh/xonsh/pull/3304/files .

(I have not checked if this change is backwards compatible with python 3.5-3.7 or if there are other usages of `ast.arguments` in mdtraj.)

rmcgibbo:
Sounds good. Would you be able to file a pull request with that change?

acnash:
Is there a brief example of the solution? I've got the same issue. I'm very new to Python and I expect I've probably changed the core code incorrectly. 

`            signature = ast.arguments(args=args, vararg=None, kwarg=None,
                                      posonlyargs=[],
                                      kwonlyargs=[], defaults=[],
                                      kw_defaults=[])`

mattwthompson:
I'm not new to Python (but far from an expert) and that change looks to be correct. Have you been able to see if it works for you locally?

acnash:
Any assistance is welcome. From within PyCharm:
`str(md)`
**"<module 'mdtraj' from '/Users/anthonynash/PycharmProjects/AGE/venv/lib/python3.8/site-packages/mdtraj/__init__.py'>"**

And the location of the file I've changed:
/Users/anthonynash/PycharmProjects/AGE/venv/lib/python3.8/site-packages/mdtraj/core

After making the change I imported MDTraj and reran the respective python commands but with no success. Any handy means of dumping the selection.py source to the command line? Thanks.

acnash:
I was running it through the PyCharm IDE and it was getting all a bit muddled. Issues resolved.

rahabet:
so how did you solved it? is there a step I should follow? I am new for python and I am getting that error and I couldn't find anything that helps me on how to fix it

[...]

PR #1523, opened 2020-01-24
Add posonlyargs argument for Python 3.8 compatibility

mattwthompson:
Just seeing if the proposal in https://github.com/mdtraj/mdtraj/issues/1509 passes CI. This passes locally this locally in a 3.7 environment, which is encouraging.

### DIFF
--- a/mdtraj/core/selection.py
+++ b/mdtraj/core/selection.py
@@ -392,8 +392,8 @@ class parse_selection(object):
         else:
             args = [ast.arg(arg='atom', annotation=None)]
             signature = ast.arguments(args=args, vararg=None, kwarg=None,
-                                      kwonlyargs=[], defaults=[],
-                                      kw_defaults=[])
+                                      posonlyargs=[], kwonlyargs=[],
+                                      defaults=[], kw_defaults=[])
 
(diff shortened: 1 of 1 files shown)


rmcgibbo:
Seems reasonable to merge, right?

mattwthompson:
This is parts of the codebase (and standard library) I'm not familiar with, so I'm not sure if there's some behavior not covered by tests. But CI is passing and it seems like this is a straightforward enough patch, so I think so.

rmcgibbo:
We’re you able to test it on 3.8 and see that it fixed the observed issues?

mattwthompson:
Well, that's embarrassing - one test was failing locally and I thought I'd try CI anyway. Of course I forgot it's not running on 3.8.

This change fixes that for me locally - at least `test_selection.py`, which passes in my 3.7 and 3.8 environments. Python 3.8 snuck in a `kind` field to `ast.Constant`. I thought about just having the tests compare `.value` but just passing `kind=None` to the constructor seemed like a better idea.

There are other failing tests in my 3.8 environment but they're all related to glue. Something for another day when people want to bump the version.

PR #1536, opened 2020-03-10
Try to get green check mark on AZP

mattwthompson:
Fixing one upstream issue while working on #1423, not sure if this will fix everything but I'm starting with one failing unit test

### DIFF
--- a/devtools/azure-pipelines-osx.yml
+++ b/devtools/azure-pipelines-osx.yml
@@ -2,7 +2,7 @@ jobs:
 - job:
   displayName: "OS X"
   pool:
-    vmImage: 'macOS-10.13'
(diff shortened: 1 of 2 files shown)


mattwthompson:
Okay, the macOS and Linux builds are passing, but the Windows build is failing due to a dependency issue: https://dev.azure.com/rmcgibbo/b71495a1-63d0-4110-bb94-45ba2ca11626/_apis/build/builds/182/logs/37

I doesn't look like it can't find the package on Anaconda Cloud, but it looks like something fails when it's trying to install it.

I checked to be sure and the Windows VM is still supported, so I doubt it's that.

rmcgibbo:
Nice work with the pandas bug.

Issue #1535, opened 2020-03-04
[tests] DataFrame.ix method removed in pandas-1.0.0


marscher:
We need to adopt tests using the long deprecated DataFrame.ix method.

- [ ] test_pdb.test_3nch_serial_resSeq

marscher:
set_index seems to remap the indicies to a consecutive sequence, thereby dropping the "holes" in the pdb file. I guess there is a Pandas like solution to this, but I'm not an expert.

mattwthompson:
Seems to be fixed in #1536, although it was purely a pandas fix and had nothing to do with PDB stuff

Issue #1545, opened 2020-05-05
Iterload function: "chunk" parameter is "chunk" multiplied by "stride"


alberginia:
I am using MDTraj with large trajectory files (dcd with 10001 frames) and I was trying to use "iterload" to carry out my calculations. 

Fixing "stride = 1", the chunks seem to be OK and of the specified size. 

However, anything other than that, yields the most weird behaviour, with strange number of chunks not corresponding to the specified size.

After pulling my hair out trying to understand how it works, I've found out looking at the code that instead o using "chunk" for the chunk size it is actually MULTIPLYING both parameters:   n_frames=chunk*stride  (see lines 520 and 522 of trajectory.py file).

Being a large trajectory, I'd like to take one of every n-th frames of the trajectory and then handle the calculations of those selected frames one by one to take it easy on computer resources (hence, I want to use "chunk = 1" or at least a small number). And right now it is impossible to do so, chunk is always a large number because stride is, which is the opposite of what you would normally need when you are using iterload.

I cannot imagine for what reason could this be the intended behaviour, but it seems strange that it was an accident. Do you have an explanation for that or could maybe fix it?

Thanks!

rmcgibbo:
Great. All fixed now. Thanks for the contribution.

PR #1546, opened 2020-05-06
Fix issue #1535, simply removes "stride" factor from "chunk" in "iterload"

alberginia:
I think this would solve issue #1545

### DIFF
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -517,9 +517,9 @@ def iterload(filename, chunk=100, **kwargs):
                 f.seek(skip)
             while True:
                 if extension not in _TOPOLOGY_EXTS:
-                    traj = f.read_as_traj(topology, n_frames=chunk*stride, stride=stride, atom_indices=atom_indices, **kwargs)
+                    traj = f.read_as_traj(topology, n_frames=chunk, stride=stride, atom_indices=atom_indices, **kwargs)
                 else:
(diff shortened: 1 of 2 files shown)


rmcgibbo:
Can you add a test, and show that the test fails for the current code and passes for the new code?

alberginia:
Hi Robert! I don't have experience with this but I tried to add one, maybe you can check if it makes sense. Greetings!

rmcgibbo:
```

def test_iterload_chunk_dcd(get_fn):
    # Makes sure that the actual chunk size yielded by iterload corresponds to the number of
    # frames specified when calling it (for dcd files).
    file = get_fn("alanine-dipeptide-explicit.dcd")
    top = get_fn("alanine-dipeptide-explicit.pdb")

    skip_frames = 3
    frames_chunk = 2

    chunks = []
    for traj_chunk in md.iterload(file, top=top, stride=skip_frames, chunk=frames_chunk):
        eq(len(traj_chunk), frames_chunk)
        chunks.append(traj_chunk)
```

Here's a fixed version of the test that doesn't crash when loading the file.

rmcgibbo:
However the test still does not appear to pass.

rmcgibbo:
Thanks for fixing this. I cleaned up the test in #1549. If that passes I'll merge that and close this PR.

rmcgibbo:
Fixed now.

PR #1551, opened 2020-05-16
Replace astor with astunparse for selection ast manipulations

rmcgibbo:
Astor doesn't appear to work with python38, and astuparse is a superior/maintained version of the same thing.
Required for #1520


rmcgibbo:
Note, this removes `astor` as a dependency in conda's meta.yaml and setup.py, and adds `astunparse` as a dependency.

PR #1552, opened 2020-05-18
fix collections.abc deprecation warnings

sroet:
I am getting some `DeprecationWarning`s in some package that I manage, that bubble up from mdtraj.
Like:
```
~/miniconda3/envs/test/lib/python3.8/site-packages/mdtraj/core/trajectory.py:32: 

DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
    from collections import Iterable
```

Apparently the [Abstract Base Classes](https://docs.python.org/3.8/library/collections.abc.html#collections-abstract-base-classes) have been moved to `collections.abc` since python `3.3`. mdtraj uses the ABCs `Iterable` and `Sequence`.

This PR fixes that deprecation warning.

### DIFF
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -29,7 +29,7 @@ from __future__ import print_function, division
 import os
 import warnings
 from copy import deepcopy
-from collections import Iterable
+from collections.abc import Iterable
 import numpy as np
 import functools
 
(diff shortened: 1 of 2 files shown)


rmcgibbo:
Great

Issue #1553, opened 2020-05-29
Possible signature modification to mdtraj.geometry.hbond.baker_hubbard()


mcs95:
Hi,
 
In the context of the baker_hubbard hydrogen bond function, I was wondering whether you could implement the suggestion on  [this line of the code](https://github.com/mdtraj/mdtraj/blob/10a0b08f53d4a7c3be48ebe301b5a91222c78e1a/mdtraj/geometry/hbond.py#L221) to expose the distance_cutoff and angle_cutoff criteria as function arguments.

Thank you in advance,
Maria Chiara

rmcgibbo:
Yeah, that seems very reasonable. Are you interested in filing a pull request to add that?

mcs95:
Great thank you! I will file a pull request in the next few days.

mcs95:
I've made a PR [here](https://github.com/mdtraj/mdtraj/pull/1556)

mcs95:
I've updated the PR to change angle_cutoff to degrees.

rmcgibbo:
👍

PR #1556, opened 2020-06-02
Signature modification to geometry.hbond.baker_hubbard() Issue #1553 

mcs95:
Changed spacing in signature

Adding a unit test

Fixed mistake in reading file

Double to single quotation mark


rmcgibbo:
Do you think it would be more helpful to pass the angle cutoff argument in degrees rather than radians?

mcs95:
Sure, passing the angle_cutoff in degrees would make the signature tidier and more intuitive. I'll modify the code and make a new pull request.

rmcgibbo:
Looks great

PR #1578, opened 2020-09-03
bugfix _topology_from_subset correct residue indexing

### LINKED ISSUE #1577: atom_slice: residue index
Hi, this is connected #1343. 

What do you think should be the behavior of `Residue.index` after atom_slice? It's initialized here:
https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/topology.py#L1391

but not updated after atom_slice. This leads to a behavior like this:
```
jgeom = igeom.atom_slice(igeom.top.select('resid 0 2'))
jgeom.top.residue(1).index
# 2
```

which seems (IMO) like something to avoid (and also contradicts the doc)
```
    index : int
        The index of the Residue within its Topology
```
https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/topology.py#L1377

I can work around it by simply monkey-patching the `index` attribute after every   `atom_slice` call, but it seems like something that could be implemented.

Would you be willing to accept a PR?

Thanks!

gph82:
The old `_topology_from_subset` was 

- first creating a new residue anyways (increasing residue count)
- then checking whether it would be added or not to the new topology.

This meant the residue count was increased regardless, leading to inconsistent residue indexing.

Now the residue is only created if the atom is going to be added 

This fixes #1577

### DIFF
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -91,14 +91,16 @@ def _topology_from_subset(topology, atom_indices):
         newChain = newTopology.add_chain()
         for residue in chain._residues:
             resSeq = getattr(residue, 'resSeq', None) or residue.index
-            newResidue = newTopology.add_residue(residue.name, newChain,
-                                                 resSeq, residue.segment_id)
+            newResidue = None
             for atom in residue._atoms:
                 if atom.index in atom_indices:
                     try:  # OpenMM Topology objects don't have serial attributes, so we have to check first.
                         serial = atom.serial
                     except AttributeError:
                         serial = None
+                    if newResidue is None:
+                        newResidue = newTopology.add_residue(residue.name, newChain,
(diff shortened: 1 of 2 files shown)


gph82:
BTW, locally with pytest:
`4 failed, 962 passed, 106 skipped, 63 xfailed, 233 warnings in 218.84s (0:03:38)`
Failed were:
```
FAILED test_hoomdxml.py::test_load - ImportError: 
FAILED test_hoomdxml.py::test_load_no_ions - ImportError: 
FAILED test_hoomdxml.py::test_correct_ordering - ImportError: 
FAILED test_nmr.py::test_spartaplus - AttributeError: 'DataFrame' object has no attribute 'ix'
```

rmcgibbo:
Ugh, looks like more conda stuff in the Azure tests.

Issue #1583, opened 2020-10-13
traj.save writes internal atom ID and not the topology serial number


gitkol:
Hi, I am trying to write a simple script that reads a trajectory, restricts it to a subset of the frames and, finally, writes a PDB file that includes only a subset of the atoms.
```
import mdtraj as md
import sys
trj_name = sys.argv[1]
top_name = sys.argv[2]
select_frames = sys.argv[3]
frame_indx = [266,448,90,30,50,46,444,115,39,84,230,350,476,478,344,346,28,9,70,196,155,10,19,75,76,24,185]
traj = md.join(md.load_frame(trj_name, index=frame_i, top=top_name) for frame_i in frame_indx)
print(traj.top)
table, bonds = traj.top.to_dataframe()
print(table.head())

keep_atoms = traj.top.select("resid 39 to 74 and name CA")
print(keep_atoms)
traj.atom_slice(keep_atoms, inplace=True)
print (traj)
print(traj.top)
table, bonds = traj.top.to_dataframe()
print(table.head())

traj.save(select_frames)

```
Here is the screen output and the top of the saved PDB file.
```
<mdtraj.Topology with 13382 chains, 13655 residues, 44606 atoms, 31234 bonds>
   serial name element  resSeq resName  chainID segmentID
0       1    N       N       1     VAL        0          
1       2    H       H       1     VAL        0          
2       3   H2       H       1     VAL        0          
3       4   H3       H       1     VAL        0          
4       5   CA       C       1     VAL        0          
[ 644  659  673  688  704  723  742  764  780  799  811  833  843  860
  884  898  919  930  945  956  976  996 1011 1021 1031 1042 1059 1076
 1087 1109 1128 1139 1156 1178 1195 1214]
<mdtraj.Trajectory with 27 frames, 36 atoms, 36 residues, and unitcells>
<mdtraj.Topology with 1 chains, 36 residues, 36 atoms, 0 bonds>
   serial name element  resSeq resName  chainID segmentID
0     645   CA       C      40     GLU        0          
1     660   CA       C      41     THR        0          
2     674   CA       C      42     GLU        0          
3     689   CA       C      43     VAL        0          
4     705   CA       C      44     LEU        0          


REMARK   1 CREATED WITH MDTraj 1.9.3, 2020-10-13
CRYST1   72.954   84.873   72.598  90.00  90.00  90.00 P 1           1 
MODEL        0
[...]

PR #1584, opened 2020-10-14
write out serial instead of index in pdbs

sroet:
This should solve #1583, 
it should not break `mdtraj` loading as that sets the number to the `Atom.serial` anyway. 

The one thing that I don't know if I am breaking a `pdb` standard here, feel free to close if this is the case.


sroet:
test failures are asure loading python 3.8, even when building python 3.7.  [see this failing line](https://dev.azure.com/rmcgibbo/mdtraj/_build/results?buildId=267&view=logs&j=0a2947bc-4ed7-5966-e5d8-17037b1704c1&t=06226abc-ecd5-556d-84df-c19948d3d6c9&l=41)

Issue #1585, opened 2020-10-14
residue.index does not make sense for some sliced trajectories that don't contain residue 0 


sroet:
Main issue:
```python
#Load a trajectory
full = md.load("data/4obe.pdb")
# Slice to any number of residues that don't start with index 0
selection = full.topology.select("resid 1 to 10")
sliced = full.atom_slice(selection)

 # get the index of the last residue
idx = [res.index for res in sliced.topology.residues][-1]
print(idx)
#try to get that index this fails
res = sliced.topology.residue(idx)
```
this returns
```
10

---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-3-2b59da77cad5> in <module>
      3 print(idx)
      4 #try to get that index this fails
----> 5 res = sliced.topology.residue(idx)

~/miniconda3/envs/contact_map/lib/python3.8/site-packages/mdtraj/core/topology.py in residue(self, index)
    808             The `index`-th residue in the topology.
    809         """
--> 810         return self._residues[index]
    811 
    812     @property

IndexError: list index out of range
```

I have tracked down the bug, and should be able to provide a PR with a fix soon

sroet:
cc @dwhswenson this came up while testing out `atom_slice`d trajectories for `contact_map` (most of our residue.index assumptions break then)

sroet:
Closed by #1586

PR #1586, opened 2020-10-14
Fix residue idx sliced traj

sroet:
solves #1585 
Added that example as a test

### DIFF
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -121,7 +121,6 @@ def _topology_from_subset(topology, atom_indices):
             # we only put bonds into the new topology if both of their partners
             # were indexed and thus HAVE a new atom
 
-
     # Delete empty residues
     newTopology._residues = [r for r in newTopology._residues if len(r._atoms) > 0]
     for chain in newTopology._chains:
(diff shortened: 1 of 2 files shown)


dwhswenson:
Looks like the pipeline is pinned to conda 4.7.12 (both Linux and Mac): https://github.com/mdtraj/mdtraj/blob/3897b18b0ea30663f4141382693d363410b923c0/devtools/azure-pipelines-linux.yml#L30

and it from the [pipeline output](https://dev.azure.com/rmcgibbo/mdtraj/_build/results?buildId=268&view=logs&j=1caebfbd-87a2-523c-35d6-6ec792022cdf&t=e8f1b440-ac3c-5cee-17c5-5bd0d8f33822&l=253), it looks like there's a problem in the python version requirements for that. (It seems to forbid 3.7 and 3.8, but for some reason allows 3.9?!) Similarly, conda-build forbids 3.8.

Suggest bumping up to `conda=4.8.5` and `conda-build=3.20.4`, which are the current versions. From `git blame`, it looks like the current pins were just the versions when MDTraj switched to Azure; @rmcgibbo, any reason not to bump these?

Issue #1186, opened 2016-10-19
Atom slicing issue


jchodera:
Using the current `master`, I'm having trouble slicing a trajectory that was written by the same version of `mdtraj`:

``` python
import mdtraj as md
import os

h5_filename = 'traj.h5'
t = md.load(h5_filename)
solute = t.topology.select('not water')
tnew = t.atom_slice(solute, inplace=False)
tnew[0].save('test.pdb')
```

Example trajectory attached.

[traj.h5.zip](https://github.com/mdtraj/mdtraj/files/537965/traj.h5.zip)

jchodera:
Oddly enough, I tested a few trajectories in `mdtraj/testing/reference/` but couldn't reproduce this issue on those trajectories.

jchodera:
Output:
```
[chodera@gpu-2-4 analyze]$ python mdtraj-test.py
Traceback (most recent call last):
  File "mdtraj-test.py", line 8, in <module>
    tnew[0].save('test.pdb')
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/mdtraj/core/trajectory.py", line 1304, in save
    return saver(filename, **kwargs)
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/mdtraj/core/trajectory.py", line 1397, in save_pdb
    bfactors=bfactors[i])
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/mdtraj/formats/pdb/pdbfile.py", line 672, in __exit__
    self.close()
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/mdtraj/formats/pdb/pdbfile.py", line 501, in close
    self._write_footer()
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/mdtraj/formats/pdb/pdbfile.py", line 429, in _write_footer
    index1 = atomIndex[atom1]
KeyError: MOL1-P1
Exception KeyError: (MOL1-P1,) in <bound method PDBTrajectoryFile.__del__ of <mdtraj.formats.pdb.pdbfile.PDBTrajectoryFile object at 0x7f5303888550>> ignored
```

JoaoRodrigues:
Bumping this because I am running into the same issue, although with a different use case:

```
import mdtraj as md
trj = md.load('whatever.nc', top='whatever.psf')
[...]

PR #1605, opened 2020-11-14
Bug #1186 atom slicing then saving to pdb

badisa:


mdtraj/formats/pdb/pdbfile.py
```diff
@@ -406,11 +406,12 @@ def _write_footer(self):
                             'A', 'G', 'C', 'U', 'I', 'DA', 'DG', 'DC', 'DT', 'DI', 'HOH']
         conectBonds = []
         if self._last_topology is not None:
+            atoms = list(self._last_topology.atoms)
```
badisa, line 409: There is likely a more sophisticated way of solving this. It seems that the atoms stored in `topology.bonds` differ by the chain index (mdtraj/core/topology.py::1569) from the atoms retrieved by iterating through chains. Unclear as what is causing that behavior based on a initial read of the code.

tests/test_selection.py
```diff
@@ -283,6 +284,14 @@ def test_top(ala):
     eq(np.asarray(wat), np.arange(22, 2269))
 
 
+def test_atom_slice(get_fn):
```
badisa, line 287: This test should likely be moved somewhere more 'pdb' centric as it is really a slice+pdb write that is the issue being addressed.

badisa:
Had been working off an older master, looks like this is no longer an issue.

PR #1618, opened 2021-02-01
increase numpy min version and tables for HDF5reporter.

### LINKED ISSUE #1617: PIP package numpy incompatibility
We are having trouble the binary incompatibility with mdtraj as pulled with pip.
I am reporting the details from a Ubuntu20.04 docker container as this is probably the cleanest environment.
The python version is 3.8 and the mdtraj version pulled is 1.9.5

Installing `mdtraj` with pip and using it to write trajectories with the HDF5 Reporter results in this import error"
```
  File "bulk_npt.py", line 10, in <module>
    from mdtraj.reporters import HDF5Reporter  # pylint: disable=import-error
  File "/home/ludwigschneider/.local/lib/python3.7/site-packages/mdtraj-1.9.5-py3.7-linux-x86_64.egg/mdtraj/__init__.py", line 30, in <module>
    from .formats.registry import FormatRegistry
  File "/home/ludwigschneider/.local/lib/python3.7/site-packages/mdtraj-1.9.5-py3.7-linux-x86_64.egg/mdtraj/formats/__init__.py", line 4, in <module>
    from .dcd import DCDTrajectoryFile
  File "mdtraj/formats/dcd/dcd.pyx", line 1, in init mdtraj.formats.dcd
ValueError: numpy.ufunc size changed, may indicate binary incompatibility. Expected 216 from C header, got 192 from PyObject
```
Manually requesting the numpy version `1.16.5` and requesting manually `pytables` seems to fix these problems.

Is there a possibility to get a pip install that works with newer versions of numpy?

InnocentBug:
Resolves #1617 
Increase the numpy minimum version.
I am reasonably sure, that I have the issues discussed in #1617 with numpy version 17.X and 18.X
I also needed to include `tables` for the newer python vserion (3.9) for HDF5Reporter to work.

Maybe you have no issues with increasing the minimum version in setup.py.
In that case your next PyPi build should work fine for every one.


rmcgibbo:
I am skeptical that this is what we want to do.

MDTraj has always worked fine with older versions of numpy because it does not use any of the latest and greatest features. numpy is generally forward compatible in the sense that extensions built with older version of numpy work with newer versions of numpy but not backward compatible (extensions built with newer versions of numpy to _not_ work with older version of numpy). Thus, increasing the required numpy version would force all users to upgrade to the latest numpy, even if they're happy with the version they're currently on.

InnocentBug:
It is the best way, I could come up with to solve the issue that I reported.
But maybe you have a better idea.
the numpy version is freely available so users can pull the new version.

I fixed the issue locally for me, I just wanted to give something back.

rmcgibbo:
Of course. Thanks. I'm glad you've fixed the issue locally. I'm going to close this PR and the associated issue, but we'll revisit it if more reports come in of the same error.

Issue #1613, opened 2021-01-08
ustunparse and py38


gph82:
I'm not sure but I think the re-introduction of astunparse (https://github.com/mdtraj/mdtraj/commit/a8b2b9c57b7af876aaddcb54ac31ea8ab27a54c5) might be causing trouble. This is off a fresh py38 environment, with newest mdtraj 1.9.5. 

Probably related: https://github.com/simonpercivall/astunparse/issues/43#issuecomment-743255764
```
Python 3.8.0 (default, Nov  6 2019, 21:49:08) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.19.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import mdtraj as md
In [2]: geom = md.load("3SN6.pdb.gz")
In [3]: geom.top.select("name O")

AttributeError                            Traceback (most recent call last)
<ipython-input-3-7de49f93a83f> in <module>
----> 1 geom.top.select("name O")

~/miniconda3/envs/new_mdtraj/lib/python3.8/site-packages/mdtraj/core/topology.py in select(self, selection_string)
   1054         """
   1055 
-> 1056         filter_func = parse_selection(selection_string).expr
   1057         indices = np.array([a.index for a in self.atoms if filter_func(a)])
   1058         return indices

~/miniconda3/envs/new_mdtraj/lib/python3.8/site-packages/mdtraj/core/selection.py in __call__(self, selection)
    397 
    398         func = ast.Expression(body=ast.Lambda(signature, astnode))
--> 399         source = astunparse.unparse(astnode)
    400 
    401         expr = eval(

~/miniconda3/envs/new_mdtraj/lib/python3.8/site-packages/astunparse/__init__.py in unparse(tree)
     11 def unparse(tree):
     12     v = cStringIO()
---> 13     Unparser(tree, file=v)
     14     return v.getvalue()
     15 

~/miniconda3/envs/new_mdtraj/lib/python3.8/site-packages/astunparse/unparser.py in __init__(self, tree, file)
     36         self.future_imports = []
     37         self._indent = 0
---> 38         self.dispatch(tree)
     39         print("", file=self.f)
     40         self.f.flush()

~/miniconda3/envs/new_mdtraj/lib/python3.8/site-packages/astunparse/unparser.py in dispatch(self, tree)
     64             return
     65         meth = getattr(self, "_"+tree.__class__.__name__)
---> 66         meth(tree)
     67 
     68 

[...]

PR #1619, opened 2021-02-01
update numpy build pins

sroet:
This implements the solution that was discussed on #1617, found by @rmcgibbo in [this comment](https://github.com/mdtraj/mdtraj/issues/1617#issuecomment-771070431) . It should prevent further issues until python `3.10`

I have only implemented it for python 3.6-3.10 and tested `import mdtraj` for the following installs ( I assume the intermediate numpy installations should also work)

|python| min_numpy| max_numpy| comment
|---|---|---|---|
|3.6|1.14.5|1.19.5| min is bound by scipy, max by pypi
|3.7|1.16.5|1.20.0| min is bound by scipy, import works with scipy warning for numpy==1.14.5
|3.8|1.17.5|1.20.0|
|3.9|1.19.3|1.20.0| 1.19.5 was used for the build but 1.19.3 (absolute minimum numpy for python 3.9) does not complain

Please let me know if you want more python versions supported than this

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,6 +2,10 @@
 requires = [
     "setuptools",
     "wheel",
-    "numpy",
+    "numpy==1.11.3; python_version=='3.6'",
+    "numpy==1.14.6; python_version=='3.7'",
+    "numpy==1.17.5; python_version=='3.8'",
+    "numpy==1.19.5; python_version=='3.9'",
+    "numpy; python_version>='3.10'",
     "Cython>0.28",
 ]


sroet:
linux/OSX/win python 3.7 failures is a missing sparta+ dependencies for the examples (plus a warning that  pytest.mark.flaky  is not recognized)
linux python `3.8` failures seem related to `astunparse` as mentioned in #1613

Most deprecation warnings are about numpy `int` now being a wrapper around python `int`, which can be solved in another PR

sroet:
Nooo, you were just to soon, I also figured out a way to fix `astunparse`. I will open up a new PR for that :stuck_out_tongue:

Issue #1620, opened 2021-02-02
enable enforcePeriodicBox option to store unwrapped coordinates


franklhy:
Currently, `mdtraj.reporters.HDF5Reporter` only saves wrapped coordinates but discards image information, which could be important for the analysis of long term diffusion. Is it possible to enable the enforcePeriodicBox option to store unwrapped coordinates? (Similar to the enforcePeriodicBox option provided by the [PDBReporter](http://docs.openmm.org/development/api-python/generated/simtk.openmm.app.pdbreporter.PDBReporter.html) from openmm.)

I have a working modified version to include this option. Should I start a PR to include this feature?

rmcgibbo:
Sure, I'd be happy to merge that.

I would be careful with the numerics for your analysis here. If everything is stored unwrapped in single precision floating point, it's very easy to lose precision that might be important. For NVT, you might have better results storing the "number of unwrappings" as an integer and the the position within the box in float32, if you're looking over long timescales.

franklhy:
Thank you for the suggestions.

> For NVT, you might have better results storing the "number of unwrappings" as an integer and the position within the box in float32, if you're looking over long timescales.

This would be the ideal way of dealing with coordinates of long simulations, just like "dump x y z ix iy iz" in LAMMPS. I would love to see such a feature. However, my current modification is rather simple and does not calculate the integer "number of unwrappings".

PR #1622, opened 2021-02-02
Add enforcePeriodicBox option for HDF5Reporter

franklhy:
To solve issue #1620. Allow HDF5Reporter to save unwrapped coordinates.

### DIFF
--- a/mdtraj/reporters/basereporter.py
+++ b/mdtraj/reporters/basereporter.py
@@ -62,7 +62,8 @@ class _BaseReporter(object):
 
     def __init__(self, file, reportInterval, coordinates=True, time=True,
                  cell=True, potentialEnergy=True, kineticEnergy=True,
-                 temperature=True, velocities=False, atomSubset=None):
+                 temperature=True, velocities=False, atomSubset=None,
+                 enforcePeriodicBox=None):
         """Create an OpenMM reporter
 
         Parameters
@@ -122,6 +123,7 @@ class _BaseReporter(object):
         self._needEnergy = potentialEnergy or kineticEnergy or temperature
         self._atomSubset = atomSubset
         self._atomSlice = None
+        self._enforcePeriodicBox = enforcePeriodicBox
 
         if not OPENMM_IMPORTED:
             raise ImportError('OpenMM not found.')
@@ -190,7 +192,7 @@ class _BaseReporter(object):
             energies respectively.
         """
         steps = self._reportInterval - simulation.currentStep % self._reportInterval
-        return (steps, self._coordinates, self._velocities, False, self._needEnergy)
+        return (steps, self._coordinates, self._velocities, False, self._needEnergy, self._enforcePeriodicBox)
 
     def report(self, simulation, state):
         """Generate a report.
--- a/mdtraj/reporters/hdf5reporter.py
+++ b/mdtraj/reporters/hdf5reporter.py
@@ -70,7 +70,7 @@ class HDF5Reporter(_BaseReporter):
     atomSubset : array_like, default=None
         Only write a subset of the atoms, with these (zero based) indices
         to the file. If None, *all* of the atoms will be written to disk.
-    enforcePeriodicBox: bool
+    enforcePeriodicBox: bool or None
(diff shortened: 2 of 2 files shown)


mdtraj/reporters/basereporter.py
```diff
@@ -62,7 +62,8 @@ def backend(self):
 
     def __init__(self, file, reportInterval, coordinates=True, time=True,
                  cell=True, potentialEnergy=True, kineticEnergy=True,
-                 temperature=True, velocities=False, atomSubset=None):
+                 temperature=True, velocities=False, atomSubset=None,
+                 enforcePeriodicBox=None):
```
rmcgibbo, line 66: Should the default value be `False` rather than `bool`?

mdtraj/reporters/hdf5reporter.py
```diff
@@ -94,9 +100,11 @@ def backend(self):
 
     def __init__(self, file, reportInterval, coordinates=True, time=True,
                  cell=True, potentialEnergy=True, kineticEnergy=True,
-                 temperature=True, velocities=False, atomSubset=None):
+                 temperature=True, velocities=False, atomSubset=None,
+                 enforcePeriodicBox=None):
```
rmcgibbo, line 104: Same comment as above about the default value.

rmcgibbo (changes requested):

mdtraj/reporters/basereporter.py
franklhy: I "copied" this from `__init__` of  class PDBReporter in openmm's [pdbreporter.py](https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/pdbreporter.py)
In the beginning, I also feel that the default should better be a `bool` (either `False` or `True`) instead of `None`. However, after reading openmm's source code, I found that openmm deal with this parameter [in this for loop](https://github.com/openmm/openmm/blob/e53bdc5eab600b4137abd80e544cb100d1acac68/wrappers/python/openmm/app/simulation.py#L214). It seems that the reporter is designed in such a way by openmm so it is better to use `None` as a default value. Therefor, I decided to follow the way openmm does and set the default to `None`.

peastman: If the argument is `None` that means to apply default wrapping rules: do it for periodic systems, but not for non-periodic systems.  That's the default value for other reporters.

franklhy: According to the openmm code, the enforcePeriodicBox can take one of the three values: `None`, `False`, and `True`. When using `None`, the coordinates will be wrapped if the simulation uses the periodic boundary condition and will be unwrapped if the simulation uses the nonperiodic boundary condition. This is the behavior of the previous HDF5Reporter (Please correct me if it is not true).
So the advantage of setting `None` as the default is that it will be backward compatible. Users who download the new MDTraj don't need to change their original code.

rmcgibbo: Got it. Could you make this more clear in the mdtraj docstring then?

mdtraj/reporters/hdf5reporter.py
```diff
@@ -70,6 +70,12 @@ class HDF5Reporter(_BaseReporter):
     atomSubset : array_like, default=None
         Only write a subset of the atoms, with these (zero based) indices
         to the file. If None, *all* of the atoms will be written to disk.
+    enforcePeriodicBox: bool
```
rmcgibbo, line 73: change to "bool or None"

rmcgibbo (changes requested):

mdtraj/reporters/hdf5reporter.py
franklhy: Good idea.

franklhy: Done.

mdtraj/reporters/basereporter.py
franklhy: Done. Thanks for pointing that out.

peastman (approved):

rmcgibbo:
Nice

PR #1629, opened 2021-03-12
fix a numpy 1.20 depr warning

sroet:
A downstream project of mine directly interfaced with this code and our test surface a:
```
DeprecationWarning: `np.int` is a deprecated alias for the builtin `int`. To silence this warning, use `int` by itself. Doing this will not modify any behavior and is safe. When replacing `np.int`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

This fixes that instance. ~And I did not see more warnings surface when running `py.test` on this repo, but I might be missing some.~
EDIT: did a `grep -r -n "np.int" --include="*.py*"` and will go to the other ones now

### DIFF
--- a/mdtraj/formats/amberrst.py
+++ b/mdtraj/formats/amberrst.py
@@ -301,7 +301,7 @@ class AmberRestartFile(object):
         coordinates, time, cell_lengths, cell_angles = self._parse(lines)
 
         if atom_indices is not None:
-            atom_slice = ensure_type(atom_indices, dtype=np.int, ndim=1,
+            atom_slice = ensure_type(atom_indices, dtype=int, ndim=1,
                                      name='atom_indices', warn_on_cast=False)
             if not np.all(atom_slice) >= 0:
                 raise ValueError('Entries in atom_slice must be >= 0')
@@ -589,7 +589,7 @@ class AmberNetCDFRestartFile(object):
             raise ValueError('NetCDF restart has ConventionVersion %s. Only '
                              'Version 1.0 is supported.' % convention_version)
         if atom_indices is not None:
-            atom_slice = ensure_type(atom_indices, dtype=np.int, ndim=1,
+            atom_slice = ensure_type(atom_indices, dtype=int, ndim=1,
                                      name='atom_indices', warn_on_cast=False)
             if not np.all(atom_slice) >= 0:
                 raise ValueError('Entries in atom_slice must be >= 0')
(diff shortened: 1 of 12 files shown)


rmcgibbo (commented):
This seems reasonable to me. Is there any relevant difference between doing this in the .pyx and the .py files?

sroet:
So there is some difference in the cytonized c code (you can do that your self by running `cythonize` on a file in this branch and compared to running the same command master).

Now, as far as I can see this might actually increase performance as it does not do an attribute access through python but instead uses python `int`

One of the changes for the cythonized `c` file "mdtraj/geometry/neighbors.pyx"
master (`pyx_t_2` is `np.int`):
```c
      __Pyx_GetModuleGlobalName(__pyx_t_1, __pyx_n_s_np); if (unlikely(!__pyx_t_1)) __PYX_ERR(0, 125, __pyx_L1_error)
      __Pyx_GOTREF(__pyx_t_1);
      __pyx_t_2 = __Pyx_PyObject_GetAttrStr(__pyx_t_1, __pyx_n_s_int); if (unlikely(!__pyx_t_2)) __PYX_ERR(0, 125, __pyx_L1_error)
      __Pyx_GOTREF(__pyx_t_2);
      __Pyx_DECREF(__pyx_t_1); __pyx_t_1 = 0;
      if (PyDict_SetItem(__pyx_t_5, __pyx_n_s_dtype, __pyx_t_2) < 0) __PYX_ERR(0, 125, __pyx_L1_error)
      __Pyx_DECREF(__pyx_t_2); __pyx_t_2 = 0;
```
this PR (`(PyObject *)(&PyInt_Type)` is `python`s `int`:
```c
 if (PyDict_SetItem(__pyx_t_5, __pyx_n_s_dtype, ((PyObject *)(&PyInt_Type))) < 0) __PYX_ERR(0, 125, __pyx_L1_error)
```

The following is a quote from [this SO answer](https://stackoverflow.com/a/46416257/15100302)
> My advise regarding np.int: Avoid it whenever possible. In Python code it's equivalent to int and in Cython code it's also equivalent to Pythons int but if used as type-identifier it will probably confuse you and everyone who reads the code! It certainly confused me...

So it seems to be better practice to use `int` over `np.int` in `pyx` files as well.

rmcgibbo:
Awesome. Thanks!

PR #1633, opened 2021-04-02
Add time-dependent distance and RDF functions

rmatsum836:
This PR adds functions to calculate the distance between two particles and the RDF as a function of time.  @mattwthompson began this work, adding functions for when `periodic=True` and `opt=True`.  I added the implementations for when `periodic=False`, and `opt=False`, as well as for triclinic systems.  Unit tests have also been added.

Best attempts were made to adhere to PEP8 standards, I could additionally run black or flake8 to enforce code style.


rmatsum836:
The build checks passed but the deploy check failed.  I'm going to go ahead and mark this as ready for review in the meantime.

rmcgibbo:
Would it be possible for you to squash the commits so that they're organized into more clear logical structure (e.g. one commit per new API-level function, or something of that sort?)

peastman:
Click on the arrow next to "Merge pull request".  There's an option for "Squash and merge".

rmcgibbo:
I was hoping that the PR authors could consolidate the commits first, because it might be best grouped as more than 1 commit.

mattwthompson:
I agree that the best outcome would be a few (~3-5) commits, but I don't think I have the git foo to safely do that without pulling my hair out (@rmatsum836 - feel free to try!). I'm fine with this behemoth being sqash-merged if that's the easiest route forward

mattwthompson:
I can also spend some time reviewing this, but maybe not for a few days

rmatsum836:
I made an attempt to group commits through squashing but made a mess of things.  I can try again, or am fine with the squash-merge route if that's okay.

rmatsum836:
Since the master branch has been merged into this branch multiple times, I believe I would have to re-resolve merge conflicts if I were to squash certain commits together.  I'm not sure if I have the permissions to squash and merge.  @mattwthompson could you do this once you have free time to review this PR?

mdtraj/geometry/distance.py
```diff
@@ -86,6 +85,61 @@ def compute_distances(traj, atom_pairs, periodic=True, opt=True):
...
+def compute_distances_t(traj, atom_pairs, time_pairs, periodic=True, opt=True):
+    """Compute the distances between pairs of atoms from times 0 to t.
+
+    Parameters
+    ----------
+    traj : Trajectory
+        An mtraj trajectory.
+    atom_pairs : np.ndarray, shape=(num_pairs, 2), dtype=int
+        Each row gives the indices of two atoms involved in the interaction.
+    time_pairs : nd.array
```
rmcgibbo, line 97:
```suggestion
    time_pairs : nd.array, shape=(num_times, 2), dtype=int
```

mdtraj/geometry/distance.py
```diff
@@ -86,6 +85,61 @@ def compute_distances(traj, atom_pairs, periodic=True, opt=True):
...
 
 
+def compute_distances_t(traj, atom_pairs, time_pairs, periodic=True, opt=True):
+    """Compute the distances between pairs of atoms from times 0 to t.
+
+    Parameters
+    ----------
+    traj : Trajectory
+        An mtraj trajectory.
+    atom_pairs : np.ndarray, shape=(num_pairs, 2), dtype=int
```
rmcgibbo, line 95:
```suggestion
    atom_pairs : np.ndarray, shape=(num_atom_pairs, 2), dtype=int
```

mdtraj/geometry/distance.py
```diff
@@ -86,6 +85,61 @@ def compute_distances(traj, atom_pairs, periodic=True, opt=True):
...
+        information, we will compute distances under the minimum image
+        convention.
+    opt : bool, default=True
+        Use an optimized native library to calculate distances. Our optimized
+        SSE minimum image convention calculation implementation is over 1000x
+        faster than the naive numpy implementation.
+
+    Returns
+    -------
+    distances : np.ndarray, shape=(n_frames, num_pairs), dtype=float
```
rmcgibbo, line 110: Reading this docstring, without first having read all of the code, I'm not sure what the shape of the return value is. If there are multiple time pairs (which I'm inferring from the fact that the argument is pluralized // time "pairs" rather than time "pair"), when why isn't the return value 3-dimensional (num_frames x num_atom_pairs x num_time_pairs)?

rmcgibbo, line 110:
```suggestion
    distances : np.ndarray, shape=(n_frames, num_atom_pairs), dtype=float
```
?

rmcgibbo: Oh no, it's (num_time_pairs x num_atom_pairs). My misunderstanding. Anyways the docstrings should be clarified.

mdtraj/geometry/distance.py
```diff
@@ -86,6 +85,61 @@ def compute_distances(traj, atom_pairs, periodic=True, opt=True):
...
+        The distance between each pair of atoms at t=0 and t=t.
+    """
+    xyz = ensure_type(traj.xyz, dtype=np.float32, ndim=3, name='traj.xyz', shape=(None, None, 3), warn_on_cast=False)
+    pairs = ensure_type(atom_pairs, dtype=np.int32, ndim=2, name='atom_pairs', shape=(None, 2), warn_on_cast=False)
+    times = ensure_type(time_pairs, dtype=np.int32, ndim=2, name='time_pairs', shape=(None, 2), warn_on_cast=False)
+    if not np.all(np.logical_and(pairs < traj.n_atoms, pairs >= 0)):
```
rmcgibbo, line 116: This is a sanity check that the atom_pairs are logical indices. Do you also want to add a sanity check that all of the time_pairs indices are >= 0 and < len(traj)?

mdtraj/geometry/distance.py
```diff
@@ -228,6 +282,18 @@ def _distance(xyz, pairs):
     return (delta ** 2.).sum(-1) ** 0.5
 
 
+def _distance_t(xyz, pairs, times):
+    "Distance between pairs of points in specified frames"
+    # TODO: Speed this up
```
rmcgibbo, line 287: Maybe `np.subtract.outer` is the function you want here to speed this up.

mdtraj/geometry/rdf.py
```diff
@@ -91,3 +91,85 @@ def compute_rdf(traj, pairs, r_range=None, bin_width=0.005, n_bins=None,
...
+
+def compute_rdf_t(traj, pairs, times, period_length=None, r_range=None,
+                  bin_width=0.005, n_bins=None, self_correlation=True,
+                  periodic=True, opt=True):
+    """Compute time-dependent radial distribution functions.
+
+    Parameters
+    ----------
+    traj : Trajectory
+        Trajectory to compute time-dependent radial distribution function in.
```
rmcgibbo, line 104:
```suggestion
        Trajectory to compute time-dependent radial distribution function on.
```

mdtraj/geometry/rdf.py
```diff
@@ -91,3 +91,85 @@ def compute_rdf(traj, pairs, r_range=None, bin_width=0.005, n_bins=None,
     norm = len(pairs) * np.sum(1.0 / traj.unitcell_volumes) * V
     g_r = g_r.astype(np.float64) / norm  # From int64.
     return r, g_r
+
+
+def compute_rdf_t(traj, pairs, times, period_length=None, r_range=None,
+                  bin_width=0.005, n_bins=None, self_correlation=True,
+                  periodic=True, opt=True):
+    """Compute time-dependent radial distribution functions.
```
rmcgibbo, line 99: It would be useful to add a little bit more color here for people who might be familiar with a radial distribution function, but not necessarily a time-dependent radial distribution function.

Like, does that just mean we compute the radial distribution function as it was at various times? Or is it related to the change in the radial distribution between pairs of time points?

mdtraj/geometry/src/geometry.cpp
```diff
@@ -132,6 +132,82 @@ void dist_mic_triclinic(const float* xyz, const int* pairs, const float* box_mat
...
+                        const float* box_matrix, float* distance_out, 
+                        float* displacement_out, const int n_frames,
+                        const int n_atoms, const int n_pairs) {
+    bool store_displacement = (displacement_out != NULL);
+    bool store_distance = (distance_out != NULL);
+    for (int i = 0; i < n_frames; i++) {
+        // Load the periodic box vectors and make sure they're in reduced form.
+        // Get box from frame of first index of time pair
+        int box_offset = times[2*i + 0] * 9;
+        box_matrix += box_offset;
```
rmcgibbo, line 145: I don't understand the purpose of this addition and the corresponding subtraction at the end of the function.

tests/test_rdf.py
```diff
@@ -86,3 +86,64 @@ def compare_gromacs_xvg(ref_filename, r, g_r):
...
+
+    r_t, g_r_t = md.compute_rdf_t(traj, pairs, times, self_correlation=False, periodic=periodic, opt=opt)
+    mean_g_r_t = np.mean(g_r_t, axis=0)
+    r, g_r = md.compute_rdf(traj[:frames], pairs, periodic=periodic, opt=periodic)
+
+    assert eq(r_t, r)
+    assert eq(mean_g_r_t, g_r, decimal=1)
+
+@pytest.mark.skip('Binning does not match up with gromacs currently.')
+def test_compare_gromacs_rdf_t(get_fn):
```
rmcgibbo, line 140: Which if these tests cases are triclinic? That seems like an important case to test thoroughly.

mdtraj/geometry/distance.py
rmatsum836: Good thought, I added this sanity check.  I also added a unit test to check that these exceptions are correctly raised.

tests/test_rdf.py
rmatsum836: I added a triclinic system for testing as a start.

mdtraj/geometry/src/geometry.cpp
rmatsum836: The goal here is get the box vector information at the first index of `time`. 

The purpose of `box_matrix += box_offset` is to get the box information at the first index of time.  Then, the subtraction at the end of the function is to reset the box information to the first frame.  

 For example, if the distance between atom pairs at `t=0` and `t=2`, the goal is to grab the box information at `t=0`. The `box_offset` allows us to grab the box information at this time, and then will reset to the first frame.   If the next time pair is at `t=5` and `t=7`, `box_offset` will find the box information at `t=5`.

Instead of subtracting the `box_offset` at the end of the function, it may be better to get the correct indices of `box_matrix` by finding the difference between the previous and current time frames of interest.

It's also probably worth noting in the docstrings the the time-dependent calculations are best performed in an  ensemble with constant volume.  In this case, the box information will be constant throughout the course of the trajectory.

mdtraj/geometry/distance.py
```diff
@@ -107,15 +107,18 @@ def compute_distances_t(traj, atom_pairs, time_pairs, periodic=True, opt=True):
 
     Returns
     -------
-    distances : np.ndarray, shape=(n_frames, num_pairs), dtype=float
-        The distance between each pair of atoms at t=0 and t=t.
+    distances : np.ndarray, shape=(n_frames, num_atom_pairs), dtype=float
```
rmcgibbo, line 110: What is 'n_frames'?

mdtraj/geometry/distance.py
```diff
@@ -107,7 +107,7 @@ def compute_distances_t(traj, atom_pairs, time_pairs, periodic=True, opt=True):
 
     Returns
     -------
-    distances : np.ndarray, shape=(n_frames, num_atom_pairs), dtype=float
+    distances : np.ndarray, shape=(len(time_pairs), num_atom_pairs), dtype=float
```
rmcgibbo, line 110:
```suggestion
    distances : np.ndarray, shape=(num_times, num_atom_pairs), dtype=float
```
it was given a name above, right?

rmatsum836: Oops, my bad, fixed it so they are now consistent.

mdtraj/geometry/src/geometry.cpp
rmcgibbo: Reading through it a second time, I understand.

Rather than actually modifying box_offset, you maybe could just index box_matrix by `box_offset + 0` through `box_offset + 8` in the lines below. It might be more clear. But I see.

mdtraj/geometry/rdf.py
```diff
@@ -91,3 +91,88 @@ def compute_rdf(traj, pairs, r_range=None, bin_width=0.005, n_bins=None,
...
+
+    # Returns shape (len(times), len(pairs))
+    frame_distances = compute_distances_t(traj, pairs, times, periodic=periodic, opt=opt)
+
+    for n, distances in enumerate(frame_distances):
+        tmp, edges = np.histogram(distances, range=r_range, bins=n_bins)
+        g_r_t[n, :] += tmp
+    r = 0.5 * (edges[1:] + edges[:-1])
+
+    # Normalize by volume of the spherical shell (see above)
```
mattwthompson, line 172: I don't know what "see above" refers to - the similar lines in `compute_rdf`? There's a decent chance I wrote this comment, so if I can't remember that should be added

tests/test_distance.py
```diff
@@ -190,3 +192,43 @@ def test_closest_contact_nan_pos():
     xyz = xyz[:-1]
     traj = md.Trajectory(xyz=xyz, topology=None)
     _verify_closest_contact(traj)
+
+
+def test_distance_t_inputs():
+    incorrect_pairs = np.array((0, ptraj.n_atoms+1))
+    with pytest.raises(ValueError):
```
mattwthompson, line 199: Just to make sure the right `ValueError` is thrown
```suggestion
    with pytest.raises(ValueError, match='atom_pairs'):
```

tests/test_distance.py
```diff
@@ -190,3 +192,43 @@ def test_closest_contact_nan_pos():
...
+        compute_distances_t(ptraj, incorrect_pairs, times)
+
+    incorrect_times = np.array((0, ptraj.n_frames+1))
+    with pytest.raises(ValueError):
```
mattwthompson, line 203: Ditto
```suggestion
    with pytest.raises(ValueError, match='time pairs'):
```

tests/test_rdf.py
```diff
@@ -86,3 +86,78 @@ def compare_gromacs_xvg(ref_filename, r, g_r):
...
+    traj = md.load(get_fn('tip3p_300K_1ATM.xtc'), top=get_fn('tip3p_300K_1ATM.pdb'))
+    pairs = traj.top.select_pairs('name O', 'name O')
+    times = list()
+    for j in range(2):
+        for i in range(4):
+            times.append([j*4+i, j*4])
+
+    r, g_r_t = md.compute_rdf_t(traj, pairs, times, r_range=(0, 1), periodic=periodic, opt=opt)
+
+    assert g_r_t.shape == (8, 200)
```
mattwthompson, line 101: Any other sanity checks that can be added here? Maybe i.e. assert the location of a peak

tests/test_rdf.py
```diff
@@ -86,3 +86,78 @@ def compare_gromacs_xvg(ref_filename, r, g_r):
...
+                        7.4472, 22.4253, 19.8283, 20.6935]) / 10
+    traj = md.load(get_fn('test_good.nc'), top=get_fn('test.parm7'))
+    pairs = [[0, 9999]]
+    time_pairs = [[0, 2]]
+    r_t, g_r_t = md.compute_rdf_t(traj, pairs, time_pairs, 
+            self_correlation=False, periodic=periodic, opt=opt)
+
+
+@pytest.mark.skip('Binning does not match up with gromacs currently.')
+def test_compare_gromacs_rdf_t(get_fn):
```
mattwthompson, line 154: This test is always skipped. If it weren't, would there even be GROMACS data to refer to? There is in the case of vanilla RDF, but if there is no time-dependent RDF data from GROMACS to compare to, I'm not sure what this test could ever capture. So I think it should be removed.

If I am the author (or original copy+paster) of this test - whoops!

mattwthompson (approved):
I found a few details that could be improved but nothing stuck out too much (especially seeing as science has been done with this code for ~2 years now).

Thanks for picking up this work from the messy state I left it.

mdtraj/geometry/rdf.py
rmatsum836: I believe this refers to lines 84-89 which point to a discussion regarding the normalization through the volume of a spherical shell.

tests/test_rdf.py
rmatsum836: This was definitely my doing.  My intention of this test was to check that calculating the time-dependent RDF with `time_pairs=[t, t]` would be equal to the calculation of the time-independent RDF.  However, this is similarly done in `test_compare_rdf_t_at_0` so I'm fine with removing this test.

As for the gromacs tests being skipped, I think this due to the MDTraj functions returning the center of the histogram bins, whereas gromacs returns the left of the bin.  Should be an easy enough fix.

tests/test_rdf.py
rmatsum836: Yep, I need to think a little more on this.

PR #1634, opened 2021-04-02
Fixes typos in shape.py

### LINKED ISSUE #1632: Docs and function name for 'relative_shape_anisotropy' seem incorrect
Hello. Loving the tools - Why is `mdtraj.relative_shape_antisotropy` not called `mdtraj.relative_shape_anisotropy`? I think the latter is the more accepted term (I don't know what 'antisotropy' is). This may be a spelling error. In addition, the documentation for the function return seems to be incorrect, suggesting that the function returns the acylindricity, not the relative shape anisotropy.

KirillShmilovich:
fixes #1632

### DIFF
--- a/mdtraj/geometry/shape.py
+++ b/mdtraj/geometry/shape.py
@@ -127,7 +127,7 @@ def acylindricity(traj):
     return c
 
 def relative_shape_anisotropy(traj):
-    """Compute the acylindricity of a trajectory.
+    """Compute the relative shape anisotropy of a trajectory.
 
     For each frame compute the principal moments then,
 
--- a/mdtraj/geometry/shape.py
+++ b/mdtraj/geometry/shape.py
@@ -26,7 +26,7 @@ import numpy as np
 
 from mdtraj.geometry.distance import compute_center_of_geometry
 
-__all__ = ['compute_gyration_tensor', 'principal_moments', 'asphericity', 'acylindricity', 'relative_shape_antisotropy']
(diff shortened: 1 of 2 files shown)


nec4 (approved):

nec4 (approved):

mdtraj/geometry/shape.py
```diff
@@ -126,7 +126,7 @@ def acylindricity(traj):
     c = pm[:,1] - pm[:,0]
     return c
 
-def relative_shape_antisotropy(traj):
+def relative_shape_anisotropy(traj):
     """Compute the acylindricity of a trajectory.
```
nec4, line 130: All the changes look good, but the docs here still imply that the function computes the acylindricity instead of the relative shape anisotropy.

nec4 (changes requested):
All the changes look great but I think that the docstring still implies that the acylindricity is being computed instead of the relative shape anisotropy.

nec4 (approved):

nec4:
Thanks for making the changes! Everything looks great - just waiting for the tests to finish.

nec4:
Alright - everything looks good to me!

SamCox822:
@nec4 @KirillShmilovich 
I think these issues still exist in the documentation [here](https://mdtraj.org/1.9.4/api/generated/mdtraj.relative_shape_antisotropy.html#mdtraj.relative_shape_antisotropy) FYI.

mattwthompson:
That's an old version (1.9.4) but the bigger problem is the possibility that we haven't actually re-built and published docs in years ... (#1810?)

PR #1635, opened 2021-04-04
[ci] Remove conda, add nix, switch from azure-pipelines to github-actions

rmcgibbo:
The attempt here is to rework the CI and doc deployment into something much simpler, faster, and more reproducible.

I got tired of wading through the azure pipelines documentation which is unreadable and the conda "solving environment" times, which are glacial.


setup.py
```diff
@@ -259,8 +259,7 @@ def geometry_extensions():
       cmdclass={'build_ext': build_ext},
       install_requires=['numpy>=1.6',
                         'scipy',
-                        "astunparse; python_version!='3.8'",
-                        "astunparse<1.6.3; python_version=='3.8'",
```
sroet, line 263: I am pretty sure this is still required for pip installs on python `3.8`? I will test and open a PR to revert this change if required.

sroet: Nvm, 204ed06 seems to have fixed that bug :smile:. Opened a reverting PR on the conda-forge feedstock as well (to be merged whenever a new release is done with these changes), to completely undo the pin.

PR #1639, opened 2021-04-14
Topology.join updates resSeq of new residues as default now

MauriceKarrenbrock:
When I first used Trajectory.stack I noticed that the resSeq of the added part restarted from 1, this can be quite annoying especially when you are adding one single residue on the top of a file (pdb, gro, ...) , as example if you add one ligand on the top pf a water box, because you would have 2 residues 1.

So I changed Topology.join in order to renumber the new residues according to the last residue of the other topology as default

both Topology.join and Trajectory.stack now accept a new argument keepIds that if set to True will give the old behavior

### DIFF
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -257,13 +257,17 @@ class Topology(object):
 
         return hash_value
 
-    def join(self, other):
+    def join(self, other, keepIds=False):
         """Join two topologies together
 
         Parameters
         ----------
         other : Topology
             Another topology object
+        keepIds : bool, optional, default=False
+            if False (default) the residue numbers (resSeq) of the
+            topology that is joined are updated in order to
+            continue from the last resSeq of this topology
 
         Returns
         -------
@@ -275,11 +279,21 @@ class Topology(object):
             raise ValueError('other must be an instance of Topology to join')
         out = self.copy()
 
+        #I need this in order to have the resSeq of the
+        #new residues to continue from the one of the
+        #las residue of this topology
+        if not keepIds:
+            out_resSeq = out.atom(-1).residue.resSeq
+
         atom_mapping = {}
         for chain in other.chains:
             c = out.add_chain()
             for residue in chain.residues:
-                r = out.add_residue(residue.name, c, residue.resSeq, residue.segment_id)
+                if keepIds:
+                    out_resSeq = residue.resSeq
+                else:
+                    out_resSeq += 1
+                r = out.add_residue(residue.name, c, out_resSeq, residue.segment_id)
                 for atom in residue.atoms:
                     a = out.add_atom(atom.name, atom.element, r,
                                      serial=atom.serial)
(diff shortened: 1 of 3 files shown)


mdtraj/core/topology.py
```diff
@@ -257,13 +257,17 @@ def __hash__(self):
 
         return hash_value
 
-    def join(self, other):
+    def join(self, other, keepIds=False):
```
MauriceKarrenbrock, line 260: if you set keepIds=True you will get the old behavior

mdtraj/core/trajectory.py
```diff
@@ -1059,7 +1059,7 @@ def join(self, other, check_topology=True, discard_overlapping_frames=False):
         return self.__class__(xyz, deepcopy(self._topology), time=time,
             unitcell_lengths=lengths, unitcell_angles=angles)
 
-    def stack(self, other):
+    def stack(self, other, keepIds=False):
```
MauriceKarrenbrock, line 1062: if you set keepIds=True you will get the old behavior

rmcgibbo:
Would it be straightforward to add a test?

MauriceKarrenbrock:
Here you are, I added a more general test for Topology.join (couldn't find one) and a specific one for the 2 variants of join keepIds

ps how long does it usually take to see this changes inside the normal mdtraj version you install via conda and pip?

pps it's the first time I add a commit to a pull request, so if you can't see any difference from before tell me, because it means I did something wrong

rmcgibbo:
> ps how long does it usually take to see this changes inside the normal mdtraj version you install via conda and pip?

~1-6 months -- it depends on how frequently I do the release and whether or not someone bugs me and emphasizes that a release is important to them. It only take a few minutes, but I generally only do it when prompted by users.

> pps it's the first time I add a commit to a pull request, so if you can't see any difference from before tell me, because it means I did something wrong

I see the commit ("create tests for previous commit"). It looks perfect.

gph82:
I might be getting this wrong, but IMO this shouldn't be the default behavior, although I see a use case of why you'd want it. It should be an option, but not the default [that is, if I'm getting this right]

My point: it's absolutely normal for molecular topologies to have repeated resSeq, in no way is it expected NOT have repeated resSeqs. Check the PDB-DB for any molecular complex (GPCRs and Gproteins, e.g.), you have many sub-units, resSeqs are repeated as many times as needed.

The resSeq carries a meaning which you normally want to conserve after joining. Structural biologists (in absence of consensus nomenclature) use these descriptors quite often (ARG131 or whatever), and there's a biophysical meaning to them, usable across different DBs.

Please excuse if you already considered all this, but it seems a bit strong an API change. 

So now I join two topologies and I the residues of the second one are getting new indices... by default?

Again, please feel free to correct me, but I don't really think this is a good idea for default behavior

MauriceKarrenbrock:
Hello, I must admit I didn't think too much about it, I simply set it in the way I considered fit in that moment.
In any case to change the default there is only to change the default value of keepIds in both Trajectory.stack and Topology.join and one line in the new test I have done for the change.

For me personally it's the same so feel free to change it if necessary

rmcgibbo:
@gph82: Sounds good.

Can we change the default back to `keepIds=True`?

Issue #1598, opened 2020-11-09
Compute center of mass for a given selection of atoms


prateekbansal97:
Hello,

I need to find the center of mass of only given selected atoms for every frame in a trajectory. The atoms are selected by the trajectory.topology.select() function. 

The current method I am using is to extract coordinates for those atoms and compute it manually, by taking a weighted average of the coordinates. Is there a better way to do this? 

Any help would be appreciated. Thanks in advance!

toontun:
Hello, 

I don't know if it is a better way but you can do it with mdtraj.

**First, create your string selection:**
`string_selection = "resid 456 to 661 or resid 1118 to 1323"`

**Second, select your atoms in your trajectory:**
`atoms_of_interest = traj.top.select(string_selection)`

Here I selected all atoms in residues 456 to 661 and 1118 to 1323. 

**Then, compute centers of mass along your trajectory:**
`centers_of_mass =  mdtraj.compute_center_of_mass(traj.atom_slice(atoms_of_interest))`

mattwthompson:
The above is the preferred way of doing an analysis of a portion of a trajectory. I'd only emphasize that `traj.atom_slice` _returns_ a new `mdtraj.Topology` object, which can be carried along and run whatever other analysis on. So, if one wanted to run analysis on that selection multiple times,

```python3
com_traj = traj.atom_slice(atoms_of_interest)
centers_of_mass = mdtraj.compute_center_of_mass(com_traj)
#  Some other analysis on `com_traj`
...
```

prateekbansal97:
Hello,

Thank you for the solutions. I was able to successfully implement it using the traj.atom_slice() method.

MauriceKarrenbrock:
Hello,

I think it would be nice to create a wrapper function that makes this process easier and less error prone.
If you like the idea I will do a pull request.

I was thinking about something as easy as

```
def compute_center_of_mass_with_selection(trajectory, selection_string='all'):

    atoms_of_interest = trajectory.topology.select(selection_string)

    return  compute_center_of_mass(trajectory.atom_slice(atoms_of_interest))
```

rmcgibbo:
I think the most convenient thing would be to add a `selection` keyword argument to `compute_center_of_mass`

MauriceKarrenbrock:
Here you are, let me know what you think about it

PR #1640, opened 2021-04-14
added select option to compute_center_of_mass

MauriceKarrenbrock:
added select option to compute_center_of_mass to make thing easier and less error prone, the default will act exactly as the older version

I also added it to the  unit test

have also substituted a np.zeros with a np.empty in the same function because it should be faster and doesn't change the result

We were talking about it in the closed issue #1598

### DIFF
--- a/mdtraj/geometry/distance.py
+++ b/mdtraj/geometry/distance.py
@@ -163,7 +163,7 @@ def compute_center_of_mass(traj, select=None):
     else:
         atoms_of_interest = traj.topology.select(select)
 
-        masses = np.array([traj.top.atom[i].element.mass for i in atoms_of_interest])
+        masses = np.array([traj.top.atom(i).element.mass for i in atoms_of_interest])
         masses /= masses.sum()
 
         xyz = traj.xyz[:, atoms_of_interest]
--- a/mdtraj/geometry/distance.py
+++ b/mdtraj/geometry/distance.py
@@ -163,10 +163,10 @@ def compute_center_of_mass(traj, select=None):
     else:
         atoms_of_interest = traj.topology.select(select)
 
-        masses = np.array([a.element.mass for a in traj.top.atoms if a.index in atoms_of_interest])
+        masses = np.array([traj.top.atom[i].element.mass for i in atoms_of_interest])
         masses /= masses.sum()
 
-        xyz = np.array(traj.xyz[:, atoms_of_interest], order='C')
+        xyz = traj.xyz[:, atoms_of_interest]
 
     for i, x in enumerate(xyz):
         com[i, :] = x.astype('float64').T.dot(masses)
--- a/mdtraj/geometry/distance.py
+++ b/mdtraj/geometry/distance.py
@@ -135,13 +135,17 @@ def compute_displacements(traj, atom_pairs, periodic=True, opt=True):
     return _displacement(xyz, pairs)
 
 
-def compute_center_of_mass(traj):
+def compute_center_of_mass(traj, select='all'):
     """Compute the center of mass for each frame.
 
     Parameters
     ----------
     traj : Trajectory
         Trajectory to compute center of mass for
+    select : str, optional, default=all
+        a mdtraj.Topology selection string that
+        defines the set of atoms of which to calculate
+        the center of mass, the default is all atoms
 
     Returns
     -------
@@ -149,7 +153,10 @@ def compute_center_of_mass(traj):
          Coordinates of the center of mass for each frame
     """
 
-    com = np.zeros((traj.n_frames, 3))
+    atoms_of_interest = traj.topology.select(select)
+    traj = traj.atom_slice(atoms_of_interest)
+
+    com = np.empty((traj.n_frames, 3))
     masses = np.array([a.element.mass for a in traj.top.atoms])
     masses /= masses.sum()
 
--- a/mdtraj/geometry/distance.py
+++ b/mdtraj/geometry/distance.py
@@ -135,7 +135,7 @@ def compute_displacements(traj, atom_pairs, periodic=True, opt=True):
     return _displacement(xyz, pairs)
 
 
-def compute_center_of_mass(traj, select='all'):
+def compute_center_of_mass(traj, select=None):
     """Compute the center of mass for each frame.
 
     Parameters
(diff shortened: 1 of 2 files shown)


mdtraj/geometry/distance.py
```diff
@@ -135,21 +135,28 @@ def compute_displacements(traj, atom_pairs, periodic=True, opt=True):
...
 
     Returns
     -------
     com : np.ndarray, shape=(n_frames, 3)
          Coordinates of the center of mass for each frame
     """
 
-    com = np.zeros((traj.n_frames, 3))
+    atoms_of_interest = traj.topology.select(select)
+    traj = traj.atom_slice(atoms_of_interest)
```
rmcgibbo, line 157: It might be more efficient not to actually slice the trajectory, but to instead just select out the right atoms using list comprehensions and numpy indexing?

MauriceKarrenbrock: I see

by using timeit I figured out that this should be the fastest solution specially when nothing is selected (when something is a quite big overhead, as expected)

```
#selection
python -m timeit -s 'import mdtraj; traj=mdtraj.load("tests/data/2EQQ.pdb")' -n 100 'mdtraj.geometry.distance.compute_center_of_mass(traj, "residue 1")'
100 loops, best of 5: 23 msec per loop

#no selection
python -m timeit -s 'import mdtraj; traj=mdtraj.load("tests/data/2EQQ.pdb")' -n 100 'mdtraj.geometry.distance.compute_center_of_mass(traj)'
100 loops, best of 5: 1.04 msec per loop
```

The only problem is that it's much less readable than before

```
def compute_center_of_mass(traj, select=None):
    com = np.empty((traj.n_frames, 3))

    if select is None:
        masses = np.array([a.element.mass for a in traj.top.atoms])
        masses /= masses.sum()

        xyz = traj.xyz

    else:
        atoms_of_interest = traj.topology.select(select)

        masses = np.array([a.element.mass for a in traj.top.atoms if a.index in atoms_of_interest])
        masses /= masses.sum()

        xyz = np.array(traj.xyz[:, atoms_of_interest], order='C')


    for i, x in enumerate(xyz):
        com[i, :] = x.astype('float64').T.dot(masses)
    return com
```

what do you think about it?

rmcgibbo: I think that's pretty readable. Looks good.

MauriceKarrenbrock:
Here you are

mdtraj/geometry/distance.py
```diff
@@ -135,25 +135,40 @@ def compute_displacements(traj, atom_pairs, periodic=True, opt=True):
...
+        masses = np.array([a.element.mass for a in traj.top.atoms])
+        masses /= masses.sum()
 
-    for i, x in enumerate(traj.xyz):
+        xyz = traj.xyz
+
+    else:
+        atoms_of_interest = traj.topology.select(select)
+
+        masses = np.array([a.element.mass for a in traj.top.atoms if a.index in atoms_of_interest])
```
sroet, line 166: small possible optimization (loop over just the subset instead of the full topology):
 ```suggestion
         masses = np.array([traj.top.atom[i].element.mass for i in atoms_of_interest])
```
(this should be identical for all topologies, but does not have to be (`top.atom[a.index]` should always return `a` but a user might have overriden `a.index`)

File: mdtraj/geometry/distance.py [hunk 3]
```diff
@@ -135,25 +135,40 @@ def compute_displacements(traj, atom_pairs, periodic=True, opt=True):
...
+        masses /= masses.sum()
+
+        xyz = np.array(traj.xyz[:, atoms_of_interest], order='C')
```
sroet, line 169: What is the reason for this extra copy? traj.xyz should already be c-ordered, right? (otherwise the copy should also be done on line `161`)
```suggestion
        xyz = traj.xyz[:, atoms_of_interest]
```

sroet (commented):
just some small optimisation comments/questions, feel free to ignore these

mdtraj/geometry/distance.py
MauriceKarrenbrock: I see, and it also looks cleaner to read,  let's see what @rmcgibbo thinks about it
+1 for me

mdtraj/geometry/distance.py
MauriceKarrenbrock: I must admit that I copied this line from the code of the topology atom slicing method without thinking too much about it.
Actually the code might be a bit faster with a fortran ordering as later we are doing a matrix multiplicatio of the transpose?

one should see what is faster between having an array contiguous in memory but having to do this copy or having a not contiguous one but no copy needed

MauriceKarrenbrock: Ok, I played around with timeit and your solution is the faster one

mdtraj/geometry/distance.py
rmcgibbo: 👍

mdtraj/geometry/distance.py
rmcgibbo: The test failure looks legitimate, I think.

mdtraj/geometry/distance.py
```diff
@@ -135,25 +135,40 @@ def compute_displacements(traj, atom_pairs, periodic=True, opt=True):
...
+        masses = np.array([a.element.mass for a in traj.top.atoms])
+        masses /= masses.sum()
 
-    for i, x in enumerate(traj.xyz):
+        xyz = traj.xyz
+
+    else:
+        atoms_of_interest = traj.topology.select(select)
+
+        masses = np.array([traj.top.atom[i].element.mass for i in atoms_of_interest])
```
sroet, line 166: I made a mistake on this line, it should be 
```suggestion
        masses = np.array([traj.top.atom(i).element.mass for i in atoms_of_interest])
```

mdtraj/geometry/distance.py
sroet: Error looks like it came from the other changed line:
```python
>   masses = np.array([traj.top.atom[i].element.mass for i in atoms_of_interest])
E   TypeError: 'method' object is not subscriptable
```
should be fixed by [this suggestion](https://github.com/mdtraj/mdtraj/pull/1640#discussion_r614274707) (I always forget that `atom` is a function and not a `list` )

PR #1641, opened 2021-04-16
Changed default value of keepIds for Topology.join & Trajectory.stack to True

MauriceKarrenbrock:
In this way it is retro compatible as requested in pull request #1639


rmcgibbo:
cc @gph82 to review

MauriceKarrenbrock:
@rmcgibbo Would it please be possible to make a new release once this pull request is gone through? Sorry if I have to bug you but I will have to use the newly implemented things in the near future

ps @gph82 should be ok with this new version because (if I haven't messed up) now the default is the old behavior

gph82:
Hi, I'm sorry but I only have intermittent access to the web RN and might be a bit late, why not ``keep_resSeqs``? "ID" is not really defined in the mdtraj, right?

MauriceKarrenbrock:
I see what you mean, I simply wanted to use the same syntax of the pdb writer in openMM (even though I don't remember if in the case of openMM the ids that are kept are the atom or the residue numbers). But if you think that  `keep_resSeqs` would be more clear or make more sense feel free to change it

PR #1647, opened 2021-05-11
now Topology.from_dataframe methon can manage residues with same resSeq

### LINKED ISSUE #1646: topology.py from_dataframe method
Hi, the problem is a bit similar to #1485 but not exactly the same.

Is there a reason why the from_dataframe method on topology.py raises an exception when there are 2 residues with the same resSeq, as this is pretty common? (line 529)

would it be a problem to make something like

```
for ri in np.unique(chain_atoms['resSeq']):
                residue_atoms = chain_atoms[chain_atoms['resSeq'] == ri]
                rnames = residue_atoms['resName']
                for name in rnames:
                    tmp_residue_atoms = residue_atoms[residue_atoms['resName'] == name]
                    segids = tmp_residue_atoms['segmentID']
                    segment_id = np.array(segids)[0]
                
                    r = out.add_residue(name, c, ri,segment_id)

                    for atom_index, atom in tmp_residue_atoms.iterrows():
                        atom_index = int(atom_index)  # Fixes bizarre hashing issue on Py3K.  See #545
                        a = Atom(atom['name'], elem.get_by_symbol(atom['element']),
                                  atom_index, r, serial=atom['serial'])
                        out._atoms[atom_index] = a
                        r._atoms.append(a)
```
(this is simply a draft)

In this way you will get 2 different residues with the same resSeq

this problem arises because openmmtools.utils.deserialize uses this mdtraj method and I happen to have structures with repeated resSeq pretty often

MauriceKarrenbrock:
closes #1646

Now the Topology.from_dataframe methon in topology.py is able to deal with a structure that has many residues with the same resSeq respecting their input order, the function in fact now is much similar to the pdb parser (maybe give a quick check if there is something I haven't considered on how dataframes need to be handled)

I have also added a test

ps if the PR is accepted could it be possible to include it in the 1.9.6 release or is it too late?

### DIFF
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -515,30 +515,30 @@ class Topology(object):
                              'starting from zero.')
         out._atoms = [None for i in range(len(atoms))]
 
-        for ci in np.unique(atoms['chainID']):
-            chain_atoms = atoms[atoms['chainID'] == ci]
-            c = out.add_chain()
+        c = None
+        r = None
+        previous_chainID = None
+        previous_resName = None
+        previous_resSeq = None
+        for atom_index, atom in atoms.iterrows():
 
-            for ri in np.unique(chain_atoms['resSeq']):
-                residue_atoms = chain_atoms[chain_atoms['resSeq'] == ri]
+            int(atom_index) # Fixes bizarre hashing issue on Py3K.  See #545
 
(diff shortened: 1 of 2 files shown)


MauriceKarrenbrock:
Hi, I noticed that the pull request has remained stuck, is there something that you don't like in the code, some bugs or similar?
If it's simply because you are busy don't worry

Have a nice day

PR #1648, opened 2021-05-14
Faster load function and more omogeneus file parser interface

MauriceKarrenbrock:
Hi, I found myself in the need to load thousands of pdb or gro files all in once and I noticed that the mdtraj.load function can be sped up without losing its high level functionalities

And I also noted that if using pdb or gro files (or in general files in the _TOPOLOGY_EXTS category) the topology gets re-parsed again for each single file.
Therefore I made the PDB and the GRO file parsers able to accept the top argument and therefore re-use the already parsed topology (I would suggest to slowly make the interface more omogeneus by implementing it in all the file parsers in _TOPOLOGY_EXTS, see #1649)

I made the load function able to deal more flexibly with the fact that some parsers may not accept top so if in the future someone implements the top keyword in other parsers it won't be necessary to change load

And changed the if order in _parse_topology in order for it to be faster if a Topology or a Trajecotry is given

The speed up is pretty good for lists of files (12% faster):
```
Before

python -m timeit -n 10 -s 'import mdtraj' -- 'mdtraj.load("tests/data/1am7_protein.pdb", top="tests/data/1am7_protein.pdb")'

10 loops, best of 5: 517 msec per loop



python -m timeit -n 2 -s 'import mdtraj; list=["tests/data/1am7_protein.pdb"]*10' -- 'mdtraj.load(list, top="tests/data/1am7_protein.pdb")'

2 loops, best of 5: 3.17 sec per loop


After

python -m timeit -n 10 -s 'import mdtraj' -- 'mdtraj.load("tests/data/1am7_protein.pdb", top="tests/data/1am7_protein.pdb")'

10 loops, best of 5: 460 msec per loop


python -m timeit -n 2 -s 'import mdtraj; list=["tests/data/1am7_protein.pdb"]*10' -- 'mdtraj.load(list, top="tests/data/1am7_protein.pdb")'

2 loops, best of 5: 2.63 sec per loop
```

ps if the PR is accepted could it be possible to include it in the 1.9.6 release or is it too late?

### DIFF
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -391,6 +391,8 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
     topkwargs.pop("top", None)
     topkwargs.pop("atom_indices", None)
     topkwargs.pop("frame", None)
+    topkwargs.pop("stride", None)
+    topkwargs.pop("start", None)
     kwargs["top"] = _parse_topology(kwargs.get("top", filename_or_filenames[0]), **topkwargs)
 
     #get the right loader
@@ -409,13 +411,9 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
         _assert_files_or_dirs_exist(filename_or_filenames)
 
 
-    #if some atom indices are requested make a subset topology
-    if kwargs.get("atom_indices", None) is not None:
-        subset_topology = kwargs["top"].subset(kwargs["atom_indices"])
-
-    else:
-        subset_topology = kwargs["top"]
-
+    if extension not in _TOPOLOGY_EXTS:
+        # standard_names is a valid keyword argument only for files containing topologies
+        kwargs.pop('standard_names', None)
 
     trajectories = []
     tmp_file = filename_or_filenames.pop(0)
@@ -430,15 +428,22 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
         #this part and have a more consistent interface and a faster load function
         t = loader(tmp_file, **kwargs)
         
-    except TypeError:
+    except TypeError as e:
+
+        #Don't want to intercept legit
+        #TypeErrors
+        if "got an unexpected keyword argument 'top'" not in str(e):
+            raise
+
         kwargs.pop('top', None)
 
         t = loader(tmp_file, **kwargs)
 
-    finally:
-
-        t.topology = subset_topology
-        trajectories.append(t)
+    #in case only a part of the atoms where selected
+    #I get the right topology to laret share with the
+    #other frames
+    subset_topology = t.topology
+    trajectories.append(t)
 
 
     # we know the topology is equal because we sent the same topology
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -439,9 +439,9 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
 
         t = loader(tmp_file, **kwargs)
 
-    #in case only a part of the atoms where selected
-    #I get the right topology to laret share with the
-    #other frames
+    # In case only a part of the atoms were selected
+    # I get the right topology to later share with the
+    # other frames
     subset_topology = t.topology
     trajectories.append(t)
 
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -435,6 +435,8 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
         if "got an unexpected keyword argument 'top'" not in str(e):
             raise
 
+        warnings.warn('top= kwargs ignored since this file parser does not support it')
+
         kwargs.pop('top', None)
 
         t = loader(tmp_file, **kwargs)
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -418,6 +418,7 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
 
 
     trajectories = []
+    tmp_file = filename_or_filenames.pop(0)
     try:
         # this is a little hack that makes calling load() more predictable. since
         # most of the loaders take a kwargs "top" except for load_hdf5, (since
@@ -427,12 +428,12 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
         # it.
         #TODO make all the loaders accept a pre parsed topology (top) in order to avoid
         #this part and have a more consistent interface and a faster load function
-        t = loader(filename_or_filenames.pop(0), **kwargs)
+        t = loader(tmp_file, **kwargs)
         
     except TypeError:
         kwargs.pop('top', None)
 
-        t = loader(filename_or_filenames.pop(0), **kwargs)
+        t = loader(tmp_file, **kwargs)
 
     finally:
 
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -454,7 +454,8 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
 
     trajectories.append(t)
 
-    # Only do this extra copy and monkey patching if needed
+    # Only do this monkey patching if needed in order not to
+    # modify the output topology
     if ('top' in kwargs) and (
         kwargs.get('atom_indices', None) is not None) and (
         len(filename_or_filenames) > 1):
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -166,7 +166,11 @@ def _parse_topology(top, **kwargs):
     else:
         ext = None  # might not be a string
 
-    if isinstance(top, string_types) and (ext in ['.pdb', '.pdb.gz', '.h5','.lh5']):
+    if isinstance(top, Topology):
+        topology = top
+    elif isinstance(top, Trajectory):
+        topology = top.topology
+    elif isinstance(top, string_types) and (ext in ['.pdb', '.pdb.gz', '.h5','.lh5']):
         _traj = load_frame(top, 0, **kwargs)
         topology = _traj.topology
     elif isinstance(top, string_types) and (ext in ['.prmtop', '.parm7', '.prm7']):
@@ -183,10 +187,6 @@ def _parse_topology(top, **kwargs):
         topology = load_hoomdxml(top, **kwargs).topology
     elif isinstance(top, string_types) and (ext in ['.gsd']):
         topology = load_gsd_topology(top, **kwargs)
-    elif isinstance(top, Trajectory):
-        topology = top.topology
-    elif isinstance(top, Topology):
-        topology = top
     elif isinstance(top, string_types):
         raise IOError('The topology is loaded by filename extension, and the '
                       'detected "%s" format is not supported. Supported topology '
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -359,7 +359,7 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
     >>> print traj2
     <mdtraj.Trajectory with 250 frames, 423 atoms at 0x11136e410>
 
-    >>> traj3 = md.load_hdf5('output.xtc', atom_indices=[0,1] top='topology.pdb')
+    >>> traj3 = md.load_hdf5('output.xtc', atom_indices=[0,1], top='topology.pdb')
     >>> print traj3
     <mdtraj.Trajectory with 500 frames, 2 atoms at 0x18236e4a0>
 
@@ -438,6 +438,17 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
         kwargs.pop('top', None)
 
         t = loader(tmp_file, **kwargs)
+
+    except ValueError as e:
+
+        if 'xyz must be shape' in str(e):
+
+            raise ValueError('The topology and the trajectory files might not contain the same atoms\n'
+            'The input topology must contain all atoms even if '
+            'you want to select a subset of them with atom_indices'
+            ) from e
+
+        raise
 
     # In case only a part of the atoms were selected
     # I get the right topology to later share with the
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -369,69 +369,96 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
         The resulting trajectory, as an md.Trajectory object.
     """
 
-    if "top" in kwargs:  # If applicable, pre-loads the topology from PDB for major performance boost.
-        topkwargs = kwargs.copy()
-        topkwargs.pop("top", None)
-        topkwargs.pop("atom_indices", None)
-        topkwargs.pop("frame", None)
-        kwargs["top"] = _parse_topology(kwargs["top"], **topkwargs)
+    # If a single filename make a list out of it in
+    #order to have an easier function later on
+    if isinstance(filename_or_filenames, string_types):
+        filename_or_filenames = [filename_or_filenames]
+    
+    
+    extensions = [_get_extension(f) for f in filename_or_filenames]
+    extension = extensions[0]
+    #Make the needed checks
+    if len(set(extensions)) == 0:
+        raise ValueError('No trajectories specified. '
+                            'filename_or_filenames was an empty list')
+    elif len(set(extensions)) > 1:
+        raise TypeError("Each filename must have the same extension. "
+                        "Received: %s" % ', '.join(set(extensions)))
 
-    # grab the extension of the filename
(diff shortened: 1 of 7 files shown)


MauriceKarrenbrock:
I have tried this new stuff in a real life situation and the speed up is actually much greater than 12%
One process that took me more than one hour is now taking me more or less 10 minutes

mdtraj/formats/gro.py
```diff
@@ -85,10 +85,13 @@ def load_gro(filename, stride=None, atom_indices=None, frame=None):
         Use this option to load only a single frame from a trajectory on disk.
         If frame is None, the default, the entire trajectory will be loaded.
         If supplied, ``stride`` will be ignored.
+    top : mdtraj.core.Topology, default=None
+        if you give a topology as input the topology won't be parsed from the pdb file
```
sroet, line 89: this should be `gro`
```suggestion
        if you give a topology as input the topology won't be parsed from the gro file
```

mdtraj/formats/gro.py
```diff
@@ -112,6 +115,9 @@ class GroTrajectoryFile(object):
     force_overwrite : bool
         If opened in write mode, and a file by the name of `filename` already
         exists on disk, should we overwrite it?
+    top : mdtraj.core.Topology, default=None
+        if you give a topology as input the topology won't be parsed from the pdb file
```
sroet, line 119:
```suggestion
        if you give a topology as input the topology won't be parsed from the gro file
```

mdtraj/formats/pdb/pdbfile.py
```diff
@@ -559,23 +542,54 @@ def _read_models(self):
...
-        if len(connectBonds) > 0:
-            # Only add bonds that don't already exist.
-            existingBonds = set(self._topology.bonds)
-            for bond in connectBonds:
-                if bond not in existingBonds and (bond[1], bond[0]) not in existingBonds:
-                    self._topology.add_bond(bond[0], bond[1])
-                    existingBonds.add(bond)
+
+
+        #load the topology if none is given
```
sroet, line 547:
```suggestion
        # Load the topology if None is given
```

mdtraj/core/trajectory.py
```diff
@@ -369,69 +369,102 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
...
+        if "got an unexpected keyword argument 'top'" not in str(e):
+            raise
+
+        kwargs.pop('top', None)
+
+        t = loader(tmp_file, **kwargs)
+
+    #in case only a part of the atoms where selected
+    #I get the right topology to laret share with the
+    #other frames
```
sroet, line 444: Missing spaces after `#` and a typo
```suggestion
    # In case only a part of the atoms were selected
    # I get the right topology to later share with the
    # other frames
```

mdtraj/formats/pdb/pdbfile.py
```diff
@@ -161,6 +164,8 @@ def load_pdb(filename, stride=None, atom_indices=None, frame=None,
 
         topology = f.topology
         if atom_indices is not None:
+            #to avoid modifying the topology if given as input
+            topology = deepcopy(topology)
             topology = topology.subset(atom_indices)
```
sroet, line 169: This does very undefined stuff if you give it a pre-subsetted topology:
```python
>>> import mdtraj as md
>>> top = md.Topology()
>>> c = top.add_chain()
>>> r = top.add_residue('test', c, 0, 0)
>>> top.add_atom('a0', 'H', r, 0)
test0-a0
>>> top.add_atom('a1', 'H', r, 1)
test0-a1
>>> top
<mdtraj.Topology with 1 chains, 1 residues, 2 atoms, 0 bonds at 0x7fa5a7447090>
>>> tops = top.subset([1])
>>> tops
<mdtraj.Topology with 1 chains, 1 residues, 1 atoms, 0 bonds at 0x7fa5445f3490>
>>> tops2 = tops.subset([1]) # Redo the same subset
>>> tops2
<mdtraj.Topology with 0 chains, 0 residues, 0 atoms, 0 bonds at 0x7fa5445f35d0>
```
Maybe error out if `topology.n_atoms` is not equal to `len(atom_indices)` after this `subset`?

tests/test_trajectory.py
```diff
@@ -627,7 +628,13 @@ def test_length(get_fn):
 
     for file in files:
         opened = md.open(get_fn(file))
-        loaded = md.load(get_fn(file), top=get_fn('native.pdb'))
+
+        if '.' + file.split('.', 1)[-1] in mdtraj.core.trajectory._TOPOLOGY_EXTS:
```
sroet, line 632: best practice to get the last extension, feel free to ignore 
```suggestion
        if '.' + file.rsplit('.', 1)[-1] in mdtraj.core.trajectory._TOPOLOGY_EXTS:
```

sroet (changes requested):
Mostly minor issues (typo's/ `pep8`). One bigger point: I don't think it is safe to subset a given topology (even with the deepcopy), as the atom_indices are not preserved between subsets, so maybe error out if `len(atom_indices) != top.n_atoms` after the subsetting.

MauriceKarrenbrock:
I am sorry I think I removed your comment about the deepcopy on the code as an error.

In any case what happens to you is normal, atom indexes are always zero indexed and select creates a new topology. When you where selecting atom 1 on a 2 atoms topology you were looking for a non existent second atom and therefore you got an empty output topology

In any case I don't know of the deepcopy is actually necessary in order not to modify the input topology, any opinions? (I am always unsure on how python deals with pointers to objects in functions)

sroet:
> In any case what happens to you is normal, atom indexes are always zero indexed and select creates a new topology. When you were selecting atom 1 on a 1 atom topology you were looking for a non existent second atom and therefore you got an empty output topology

Yes, I understand that that is what is supposed to happen, I am just very worried that a user might do:
```python
# Say I am only interested even indices in a pdb with 8 atoms
frame0 = mdtraj.load("test1.pdb", atom_indices=[0,2,4,6])
# frame0.topology is now a subsetted topology 
# Now use the loading trick as you describe (don't build a new top)
frame1 = mdtraj.load("test2.pdb", atom_indices=[0,2,4,6], top=frame0.topology)
```
This leads would lead to `frame1` having a broken topology with only information for atom `0` and `6` (the even atoms in the subsetted topology) while the `xyz` contains the info for atoms `0`,`2`,`4`,`6`. Personally, I would expect that a given top would never be altered (even a copy inside an instance). 

> In any case I don't know of the deepcopy is actually necessary in order not to modify the input topology, any opinions? (I am always unsure on how python deals with pointers to objects in functions)

Python is mostly pass by reference, so I would keep the deepcopy to be safe

MauriceKarrenbrock:
I tried to do exactly as you said and I got an exception
```
  File "/media/maurice/DATI/Documenti/maurice/Phd_unige/mdtraj/mdtraj/core/trajectory.py", line 429, in load
    t = loader(tmp_file, **kwargs)
  File "/media/maurice/DATI/Documenti/maurice/Phd_unige/mdtraj/mdtraj/formats/pdb/pdbfile.py", line 189, in load_pdb
    unitcell_angles=unitcell_angles)
  File "/media/maurice/DATI/Documenti/maurice/Phd_unige/mdtraj/mdtraj/core/trajectory.py", line 1202, in __init__
    self.xyz = xyz
  File "/media/maurice/DATI/Documenti/maurice/Phd_unige/mdtraj/mdtraj/core/trajectory.py", line 916, in xyz
    warn_on_cast=False, add_newaxis_on_deficient_ndim=True)
  File "/media/maurice/DATI/Documenti/maurice/Phd_unige/mdtraj/mdtraj/utils/validation.py", line 148, in ensure_type
    raise error
ValueError: xyz must be shape (Any, 4, 3). You supplied  (1, 22, 3)
```

So I guess that we don't need to worry (I used native.pdb from the tests/data directory)

sroet:
Alright, I had some time to properly run through your code.

There is some feature regression with this:
master:
```python
>>> import mdtraj as md
>>> frame0 = md.load(["native.pdb"], atom_indices=[0,2,4,6])
>>> frame1 = md.load(["native.pdb"], atom_indices=[0,2,4,6], top=frame0.topology)
/home/sander/miniconda3/envs/python3/lib/python3.7/site-packages/mdtraj/core/trajectory.py:422: UserWarning: top= kwarg ignored since file contains topology information
  warnings.warn('top= kwarg ignored since file contains topology information')
```
this PR:
```python
>>> import mdtraj as md
>>> md.load(["native.pdb", "native.pdb"], atom_indices=[0,2,4,6])
<mdtraj.Trajectory with 2 frames, 4 atoms, 2 residues, without unitcells at 0x7eff8e6c59d0>
>>> frame0 = md.load("native.pdb", atom_indices=[0,2,4,6])
>>> frame0
<mdtraj.Trajectory with 1 frames, 4 atoms, 2 residues, without unitcells at 0x7eff8e6aefa0>
>>> frame1 = md.load("native.pdb", atom_indices=[0,2,4,6], top=frame0.topology)
  File "/home/sander/github_files/mdtraj/mdtraj/core/trajectory.py", line 429, in load
    t = loader(tmp_file, **kwargs)
  File "/home/sander/github_files/mdtraj/mdtraj/formats/pdb/pdbfile.py", line 187, in load_pdb
    traj = Trajectory(xyz=coords, time=time, topology=topology,
  File "/home/sander/github_files/mdtraj/mdtraj/core/trajectory.py", line 1202, in __init__
    self.xyz = xyz
  File "/home/sander/github_files/mdtraj/mdtraj/core/trajectory.py", line 915, in xyz
    value = ensure_type(value, np.float32, 3, 'xyz', shape=shape,
  File "/home/sander/github_files/mdtraj/mdtraj/utils/validation.py", line 148, in ensure_type
    raise error
ValueError: xyz must be shape (Any, 2, 3). You supplied  (1, 4, 3)
```
Which might or might not be acceptable. 
In my opinion the error now is very strange/wrong, as I gave a topology with 4 atoms, but now the xyz only expects 2.

However, this might be solvable with proper documentation warning the user that the topology has to be correct for the whole loaded pdb, not just the atom_indices

@MauriceKarrenbrock  Another speedup that you might want to look into is to prevent `subset` or `copy` of topologies. [In a project of mine](https://github.com/dwhswenson/contact_map/pull/35#issuecomment-389322129), I discovered that subsetting is reasonably slow (in the order of 0.4 s for a 2700 atom system), and it scales both with the number of atoms in the full topology and the number of atoms after slicing. 

If you want to prevent this overhead and you feel that this is in the scope of this PR, then you can add:

```python
    # Little monkey-patches to prevent further subsetting Topologies                    
    top = subset_topology.copy()                                                
    # Override copy to just return self after this  (this is called by deepcopy)                                  
    top.copy = type(top.copy)(lambda self: self, top)                           
    # Override subset to just return self                                             
    top.subset = type(top.subset)(lambda self, atom_indices: self, top)         
    kwargs['top'] = top                                                             
```
 before the `for f in filename_or_filenames` loop in `trajectory.py`

mdtraj/formats/pdb/pdbfile.py
```diff
@@ -161,6 +164,8 @@ def load_pdb(filename, stride=None, atom_indices=None, frame=None,
 
         topology = f.topology
         if atom_indices is not None:
+            #to avoid modifying the topology if given as input
+            topology = deepcopy(topology)
```
sroet, line 168: I looked into the code and `topology.subset` already makes a copy, so this is not needed
```suggestion
```

MauriceKarrenbrock:
I made the exception more informative in case one introduces a mismatching topology so there should be less problems of clarity and added a warning when the file parser doesn't support top in the kwargs

For the second thing I see the advantages but as the final trajectory will contain this modified topology there will be some hard to catch bugs on the user side (the user thinks he is copying or subsetting but in reality he is not) therefore I wouldn't insert it in the load function, but it might be added as a suggestion in the documentation under `Notes` so that a user can overwrite the subset method of the input topology if needed

sroet:
> For the second thing I see the advantages but as the final trajectory will contain this modified topology there will be some hard to catch bugs on the user side (the user thinks he is copying or subsetting but in reality he is not) therefore I wouldn't insert it in the load function, but it might be added as a suggestion in the documentation under Notes so that a user can overwrite the subset method of the input topology if needed.

Ah, but that broken topology is always overwritten with the correct `subset_topology` inside your load-loop, so it should never exit that function

MauriceKarrenbrock:
> Ah, but that broken topology is always overwritten with the correct subset_topology inside your load-loop, so it should never exit that function

Yes it's actually true so this problem doesn't exists

But the functions that do the actual parsing need the complete topology not the subsetted one so the patch to work would have to overwrite the subset function of the whole topology in order to return the subsetted one

On the other side I am not that sure that overwriting copy will help that much because subset doesn't use copy but does directly instantiate a new topology

sroet:
> On the other side I am not that sure that overwriting copy will help that much because subset doesn't use copy but does directly instantiate a new topology

That was added because there was a `deepcopy` inside the load, before we figured out that it was not needed.

> But the functions that do the actual parsing need the complete topology not the subsetted one so the patch to work would have to overwrite the subset function of the whole topology in order to return the subsetted one

Which functions are you referring to? The `subset_topology` should always be valid for loading the trajectories, otherwise that whole loop were you assign the `t.toplogy = subset_toplogy` is wrong?

MauriceKarrenbrock:
I hoped to make things more clear but maybe how it was before was easier: in the end the file parsing is made with the specific trajectory parser of the given format. And all this parsers want the full topology as input and will then deal autonomously with the subsetting stuff (if needed). But as there are some file formats that still don't accept a topology as an input they will all create a new instance of a topology and you will end up with hundreds if not thousands of perfectly equal topologies in memory (this is also true in the case of subsetting if you don't use your "trick").
You can deal with it in two ways
* In master as you can se the first frame of the trajectory points to a topology and the others are made to point to None in a loop
* I did it by making them all point to the same topology in a loop

The effect on the memory and on speed should be equivalent but I guess that it was more clear before, shall I make them all point to None except the first one as in master?

sroet:
> The effect on the memory and on speed should be equivalent but I guess that it was more clear before, shall I make them all point to None except the first one as in master?

As you said, it should not matter and I have no strong preference either way.

MauriceKarrenbrock:
So, I have added the monkey patch on topology.subset and I made all the topologies except the first one point to None (I think it might be clearer to someone new).
The monkey patch is only done if needed because otherwise the output topology may be affected.
In this way everything shall be fast and robust enough

mdtraj/core/trajectory.py
```diff
@@ -369,69 +369,126 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
...
+        len(filename_or_filenames) > 1):
+
+        # In case only a part of the atoms were selected
+        # I get the right topology that
+        # kwargs['top'].subset shall return
+        subset_topology = trajectories[0].topology
+
+        # Little monkey-patch to prevent further subsetting Topologies
+        # this modified version of the topology will never exit this function
+        kwargs['top'].subset = lambda self, atom_indices : subset_topology
```
sroet, line 470: This does not actually work (`self` is not passed when monkey patching in this way)
```suggestion
        kwargs['top'].subset = lambda atom_indices : subset_topology
```

mdtraj/core/trajectory.py
```diff
@@ -369,69 +369,126 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
...
+
+        raise
+
+    trajectories.append(t)
+
+    # Only do this monkey patching if needed in order not to
+    # modify the output topology
+    if ('top' in kwargs) and (
+        kwargs.get('atom_indices', None) is not None) and (
+        len(filename_or_filenames) > 1):
```
sroet, line 461: `top` is always in kwargs (set on line 396), and the `len` should check for `> 0` after the `pop` on `419`
```suggestion
    if kwargs.get('atom_indices', None) is not None) and (
        len(filename_or_filenames) > 0):
```

sroet (changes requested):
two small comments on the new piece

MauriceKarrenbrock:
I corrected everything except for the top in kwargs part because as you can see at the except TypeError line 431 top is sometimes removed for some parsers that don't support it

sroet:
> I corrected everything except for the top in kwargs part because as you can see at the except TypeError line 431 top is sometimes removed for some parsers that don't support it

:facepalm:  You are right, thanks for checking and adding the test

MauriceKarrenbrock:
No worries ;-)
Thank you for the revisions and the help!

MauriceKarrenbrock:
Hi, I noticed that the pull request has remained stuck, is there something that you don't like in the code, some bugs or similar?
If it's simply because you are busy don't worry

Have a nice day

sroet:
@MauriceKarrenbrock I assume your ping was for me (As I am the only other participant on this PR). I am just an active user of `mdtraj`, but don't have write access to accept PRs. Friendly pinging @mdtraj  might get someone with write access to have a look at this PR.

MauriceKarrenbrock:
Hi @sroet
yes, I thought you had write access. In any case Thank you very much for your kind answer!

PR #1661, opened 2021-07-09
Reduce memory usage of rdf_t

jacoblumpkins:
This will make the time dependent rdf work on a set number of atom-atom pairs at a time instead of the full list, then average the results at the end. It should reduce the maximum memory usage significantly and safely enable aggressive  multiprocessing.


rmcgibbo:
cc @rmatsum836

rmatsum836 (commented):
Looks good, thanks!  @mattwthompson Could you approve the CI builds to run?

I think the current tests should do a pretty thorough job of checking the accuracy of the function.

I would like to compare the result of the old implementation to the new but I need to think of how to do so.  Perhaps I can generate a txt file from `compute_rdf_t` with the master branch implementation and add a test that compares to the PR implementation?

Also not sure if `num_concurrent_pairs` is okay at `100000`.  Have you done any testing yet on performance by changing `num_concurrent_pairs`?

mattwthompson:
Ping me again when you think this is ready to go and I'm happy to review it!

mdtraj/geometry/rdf.py
```diff
@@ -158,21 +158,31 @@ def compute_rdf_t(traj, pairs, times, period_length=None, r_range=None,
...
-    V = (4 / 3) * np.pi * (np.power(edges[1:], 3) - np.power(edges[:-1], 3))
-    norm = len(pairs) / (period_length) * np.sum(1.0 / traj.unitcell_volumes) * V
-
-    g_r_t = g_r_t.astype(np.float64) / norm  # From int64.
-
-    return r, g_r_t
+    g_r_t = []
+    weights = []
+
+    # Splits pairs into smaller chunks so that frame_distances is not excesivly large
```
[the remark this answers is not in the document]
jacoblumpkins: I chose 100000 fairly arbitrarily based on a desired final size. I took 5 GB as the goal, then did the following logic:
8 bytes/element * num_concurrent_pairs * 200 times * 32 cpu processes = 5 GB. Solving gives num_concurrent_pairs ~= 100000. This number can likely be increased quite a bit from where it is now, I'll do some timing to see how sensitive the speed of the code is to it.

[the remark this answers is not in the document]
jacoblumpkins: After testing with `mprof`, 100000 seems to be the sweet spot for both memory usage and speed with 200 times analyzed and my test system of 2000 water molecules. It was actually faster than the original since my desktop did not need to buffer ram. I made it a flag so the user can change it if they wish.

rmatsum836:
@jacoblumpkins If you're happy with the state of this PR, could you mark it ready for review?

@mattwthompson I added a test that loads in rdf_t data generated from master and compares it the PR.  I'm not sure if this is necessary or if there is a better way to handle this test case, but I'd be happy with any suggestions.  I think we're at a point where this is ready for review.

rmatsum836:
@mattwthompson Could you approve the workflows?

mattwthompson:
Looks like the workflows ran?

rmatsum836:
@mattwthompson pinging for  review.  Thanks!

mdtraj/geometry/rdf.py
```diff
@@ -158,21 +160,30 @@ def compute_rdf_t(traj, pairs, times, period_length=None, r_range=None,
...
-
-    # Normalize by volume of the spherical shell (see above)
-    V = (4 / 3) * np.pi * (np.power(edges[1:], 3) - np.power(edges[:-1], 3))
-    norm = len(pairs) / (period_length) * np.sum(1.0 / traj.unitcell_volumes) * V
-
-    g_r_t = g_r_t.astype(np.float64) / norm  # From int64.
-
-    return r, g_r_t
+    g_r_t = []
+    weights = []
```
mattwthompson, line 164: I always thought initializing an empty array was better practice over starting with an empty list, but I never looked into why, or if that was even the case. It doesn't look like this is a major part of this patch, so it's more a point of curiosity as long as it runs well and does what you and @rmatsum836 need it to d.

mdtraj/geometry/rdf.py
```diff
@@ -158,21 +160,30 @@ def compute_rdf_t(traj, pairs, times, period_length=None, r_range=None,
...
+        r = 0.5 * (edges[1:] + edges[:-1])
+
+        # Normalize by volume of the spherical shell (see above)
+        V = (4 / 3) * np.pi * (np.power(edges[1:], 3) - np.power(edges[:-1], 3))
+        norm = len(pairs_set) / (period_length) * np.sum(1.0 / traj.unitcell_volumes) * V
+
+        temp_g_r_t = temp_g_r_t.astype(np.float64) / norm  # From int64.
+        g_r_t.append(temp_g_r_t)
+
+    g_r_t_final = np.average(g_r_t, axis=0, weights=weights)
```
mattwthompson, line 187: Nice find - my worry with slicing it up like this was that the end "chunks" could be weighted weirdly if they were of strange length. Looks like this accounts for that.

mattwthompson (commented):
Looks good - would it be possible to add a test that ensures the results are not sensitive to the selection of `num_concurrent_pairs`? I don't know if the trajectory included in the repo is long enough to do that or not.

@rmatsum836 I assume this doesn't impact the results?

mdtraj/geometry/rdf.py
rmatsum836: @jacoblumpkins Is there a reason why you changed `type(g_r_t)` from array to list?  

Initializing an empty NumPy array should be more performant than initializing an empty list (specifying exact memory vs overallocting memory for a list and copying objects over to a new list when memory is full) so we should try to stay with NumPy if possible.

mdtraj/geometry/rdf.py
```diff
@@ -109,6 +109,8 @@ def compute_rdf_t(traj, pairs, times, period_length=None, r_range=None,
         Each row gives the indices of two atoms.
     times : array-like, shape=(any, 2), dtype=int
         Each row gives the indices of two frames.
+    num_concurrent_pairs : int, default=100000
```
rmatsum836, line 112: This is a nit but it might be better to name this argument `n_concurrent_pairs` which stays consistent with the naming convention for other arguments (ex: `n_bins`).

mdtraj/geometry/rdf.py
jacoblumpkins: I just made the array type changes and added a test for `n_concurrent pairs`.

mdtraj/geometry/rdf.py
jacoblumpkins: Got it

jacoblumpkins:
@mattwthompson Can you review this again? Thanks!

mdtraj/geometry/rdf.py
```diff
@@ -93,9 +93,9 @@ def compute_rdf(traj, pairs, r_range=None, bin_width=0.005, n_bins=None,
     return r, g_r
 
 
-def compute_rdf_t(traj, pairs, times, period_length=None, r_range=None,
-                  bin_width=0.005, n_bins=None, self_correlation=True,
-                  periodic=True, opt=True):
+def compute_rdf_t(traj, pairs, times, n_concurrent_pairs = 100000,
```
rmcgibbo, line 96: Small request. Can you move the new argument to the end of the argument list, or maybe right before `opt`? I think it would be preferable to have the options that control the meaning of what's computed come first, and the options that control the performance/memory/speed but not the computed values come at the end.

rmcgibbo (changes requested):
Just one small request. Otherwise looks good.

rmcgibbo (approved):

rmcgibbo:
Thanks for adding this, @jacoblumpkins!

Issue #1611, opened 2021-01-07
Incorrect PDB atom indices


SimonBoothroyd:
## Description

It seems that `1.9.5` changed the way that the indices of atoms are [written to PDB files ](https://github.com/mdtraj/mdtraj/pull/1584). This seems to have introduced several issues.

### Round-tripping PDB without TER statements

Loading a PDB file which does not contain TER statements and re-saving out the loaded structure will lead to a PDB file with atom indices which do not take into account the newly added TER statements. 

This can be reproduced by loading and then re-saving the following PDB

```
ATOM      1  O1  HOH A   1       4.012   4.851   0.586  1.00  0.00           O
ATOM      2  H1  HOH A   1       4.826   4.387   0.302  1.00  0.00           H
ATOM      3  H2  HOH A   1       3.243   4.622   0.021  1.00  0.00           H
ATOM      4  C1  MOL B   1       3.024   3.160   3.102  1.00  0.00           C
ATOM      5  O1  MOL B   1       4.131   3.989   3.088  1.00  0.00           O
ATOM      6  H1  MOL B   1       2.484   3.163   4.060  1.00  0.00           H
ATOM      7  H2  MOL B   1       3.406   2.119   2.930  1.00  0.00           H
ATOM      8  H3  MOL B   1       2.305   3.360   2.302  1.00  0.00           H
ATOM      9  H4  MOL B   1       3.955   4.912   3.354  1.00  0.00           H
CONECT    4    5    6    7    8
CONECT    5    4    9
CONECT    6    4
CONECT    7    4
CONECT    8    4
CONECT    9    5
END
```

according to:

```
traj_1 = mdtraj.load("input.pdb")
print([(bond.atom1.index, bond.atom2.index) for bond in traj_1.topology.bonds])

>>> [(0, 1), (0, 2), (3, 4), (3, 5), (3, 6), (3, 7), (4, 8)]

traj_1.save_pdb("output.pdb")

traj_2 = mdtraj.load("output.pdb")
print([(bond.atom1.index, bond.atom2.index) for bond in traj_2.topology.bonds])

>>> [(0, 1), (0, 2), (4, 5), (4, 6), (4, 7), (4, 8)]

```

The mis-indexing of the PDB file then also leads to the bonds in the molecule being incorrectly interpreted as highlighted by the above snippet.

`output.pdb`:

```
MODEL        0
ATOM      1  O   HOH A   1       4.012   4.851   0.586  1.00  0.00           O  
[...]

Issue #1343, opened 2018-04-11
atom_slice: order of the indices


gph82:
traj.atom_slice conserves the (potentially arbitrary) order of atom_indices when selecting the coordinates, but sticks to the original, sequential order in the topology, leading to this behavior. 
``` python
sliced = igeom.atom_slice([0,1])
for aa in sliced.top.atoms:
     print(aa, aa.index, sliced.xyz[:, aa.index])
#MET1-N 0 [[ 4.38999987  3.79559994  1.90639997]]
#MET1-H 1 [[ 4.40560007  3.75        1.81830001]]

sliced = igeom.atom_slice([1,0])
for aa in sliced.top.atoms:
     print(aa, aa.index, sliced.xyz[:, aa.index])
#MET1-N 0 [[ 4.40560007  3.75        1.81830001]]
#MET1-H 1 [[ 4.38999987  3.79559994  1.90639997]]
```

Forcing the sorting prior to slicing here 
https://github.com/mdtraj/mdtraj/blob/2d7ba6458d4bfcd163cdcc940aeb399643343964/mdtraj/core/trajectory.py#L1714
and 
https://github.com/mdtraj/mdtraj/blob/2d7ba6458d4bfcd163cdcc940aeb399643343964/mdtraj/core/trajectory.py#L1717

might be a solution, but changes API behavior. If okay, happy to contribute with PR with that solution.

Otherwise, at least a line in the doc specifying that atom_indices must be sorted or else unexpected behavior can occur.

justinrporter:
Good find!

I am wondering... could `aa.index` be made match xyz, rather than the other way around?

In other words, if I ask for atoms `[1, 0]`, maybe the resulting topology should also store atom 1 first? By analogy to pandas, where `df.iloc[[1, 0]]` will return rows 1 then 0.

bertadenes:
Found the same bug in v1.9.3.

weiHelloWorld:
I got the same issue, is there any update on this?

bertadenes:
In my local repo I implemented a check for sorting and redundancy, also happy to create a pull request if OP had not.

justinrporter:
I don't have merge bits, but maybe @rmcgibbo could say if he'd merge such a PR?

rmcgibbo:
Sure. Which ever behavior you think is more intuitive. We could also just raise an error if they aren't sorted.

justinrporter:
[...]

PR #1665, opened 2021-07-29
don't allow non-sorted or non-unique indices in atom_slice

### LINKED ISSUE #1664: using atom_slice with nonsequential indices gives incorrect topology
## Issue

When using `atom_slice` with indices that are non sequential, there is a miss-match between the topology and the atomic coordinates.

i.e. using `atom_slice` with indices [0, 1, 2, ...] give the correct result, however, if used with indices [2, 1, 0, ...] the output is a `Trajectory` object with atomic coordinates matching the sliced indices, but with topology that is in the order of the sorted indices.

The erroneous function is `_topology_from_subset` found in `mdtraj.core.topology`

## Example code

```
import mdtraj as md
pdb = md.load(pdb_filename)
pdb_slice0 = pdb.atom_slice([0,1,2])
pdb_slice1 = pdb.atom_slice([2,0,1])
print(list(pdb_slice0.top.atoms) == list(pdb_slice1.top.atoms))
print(pdb_slice0.xyz == pdb_slice1.xyz)
```

returns

```
True
False
```
 Meaning that the topology is the same, but the indices are different (the second slice has scrambled coordinates).

## code versions

using mdtraj 1.9.1 with python 3.6

sroet:
closes #1664 

I could not think of a trivial to fix on the `topology` side. A couple things that needed to be taken into account:

1. What happens if atom [0,1] belong to residue 0 and atom [2] to residue 1? Should the residue order also be changed? (This can lead to broken protein chains, and what to do if [0,1] are residue 0, [2,3] are residue 1 and a user ask for a [0,2,1,3] slice?)
2. Same as 1. but with different chains

I don't have the bandwidth to carefully fix this on the topology side, but I did implement a quick fix by not allowing `non-sorted` slicing of the `xyz`. This is now fixed auto-magically with a warning ( by using `np.unique(np.array(indices)`).  It also fixes and warns if you try to slice with a non-unique set of indices `pdb.atom_slice([0,1,2,2])` (which used to raise an error form deep inside the code due to a mismatch in expected xyz shape and the real one) . 

I also added the issue as a test

@mizimmer90 would this behavior (fixing the xyz instead of the topology) be enough to solve your issue?

### DIFF
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -373,8 +373,8 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
     #order to have an easier function later on
     if isinstance(filename_or_filenames, string_types):
         filename_or_filenames = [filename_or_filenames]
-    
-    
+
+
     extensions = [_get_extension(f) for f in filename_or_filenames]
     extension = extensions[0]
     #Make the needed checks
@@ -427,7 +427,7 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
         #TODO make all the loaders accept a pre parsed topology (top) in order to avoid
         #this part and have a more consistent interface and a faster load function
         t = loader(tmp_file, **kwargs)
-        
+
     except TypeError as e:
 
         #Don't want to intercept legit
@@ -468,7 +468,7 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
         # Little monkey-patch to prevent further subsetting Topologies
         # this modified version of the topology will never exit this function
         kwargs['top'].subset = lambda atom_indices : subset_topology
-        
+
 
 
     # We know the topology is equal because we send the same topology
@@ -479,7 +479,7 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
     # files to be read in without using ridiculous amounts of memory.
     for f in filename_or_filenames:
         t = loader(f, **kwargs)
-        
+
         t.topology = None
         trajectories.append(t)
 
@@ -1800,7 +1800,12 @@ class Trajectory(object):
         --------
         stack : stack multiple trajectories along the atom axis
         """
-        xyz = np.array(self.xyz[:, atom_indices], order='C')
+        indices = np.unique(np.array(atom_indices))
+        if (len(indices) != len(atom_indices) or
+            np.not_equal(indices, atom_indices).any()):
+            s = "Atom indices were made unique and sorted before atom slicing"
+            warnings.warn(s, UserWarning)
+        xyz = np.array(self.xyz[:, indices], order='C')
         topology = None
         if self._topology is not None:
             topology = self._topology.subset(atom_indices)
--- a/tests/test_trajectory.py
+++ b/tests/test_trajectory.py
@@ -563,10 +563,10 @@ def test_iterload_chunk_dcd(get_fn):
 
     full = md.load(file, top=top, stride=skip_frames)
     length = len(full)
-    
+
 
     chunks = []
-    for traj_chunk in md.iterload(file, top=top, stride=skip_frames, chunk=frames_chunk):        
+    for traj_chunk in md.iterload(file, top=top, stride=skip_frames, chunk=frames_chunk):
         chunks.append(traj_chunk)
     joined = md.join(chunks)
     assert len(full) == len(joined)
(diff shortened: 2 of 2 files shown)


gph82:
Hi, just this was marked as stale, but I knew this had come up before:
https://github.com/mdtraj/mdtraj/issues/1343

I'm a bit unclear as to why this wasn't solved by:
https://github.com/mdtraj/mdtraj/pull/1578

mizimmer90:
@gph82 thanks for cluing us into the previous issue!

@sroet Sorting the indices is definitely a valid solution! However, my intuition follows @justinrporter from #1343 in that the topology should match the xyz for arbitrary selections of indices. I've had instances where I wish to reorder the topology to match formatting of certain programs.

My vote is that the current PR should be merged at a minimum, although matching the topology to the indices is the ideal solution. I'm open to writing a new PR, unless someone else has preferences as to the best way to accomplish this?

sroet:
@mizimmer90 The maintainers of this codebase seems quite willing to look at PRs, so feel free to open one if you have the time and need.

If you do decide to open a PR, could I ask you to include the following test in `test_pdb.py` (as I have been bitten before by altering the topology, see issue #1611)? Not all of these cases have to pass, but the ones that are not supported should raise an informative error (in my opinion)

```python
@pytest.mark.parametrize("indices", [[0, 1, 2, 3, 4, 5, 6],                     
                                     [0, 1, 2, 3, 4, 5, 6, 6],                  
                                     [6, 5, 4, 3, 2, 1, 0],                     
                                     [0, 2, 4, 6, 1, 3, 5]])                    
def test_multichain_sliced_load_cycle(indices, get_fn):                         
    pdb = load(get_fn('issue_1611.pdb')).atom_slice(indices)                    
    bonds = [(bond.atom1.index, bond.atom2.index)                               
             for bond in pdb.topology.bonds]                                    
    pdb.save(temp)                                                              
    pdb2 = load_pdb(temp)                                                       
    bonds2 = [(bond.atom1.index, bond.atom2.index)                              
              for bond in pdb2.topology.bonds]                                  
    assert len(bonds) == len(bonds2)    
```

mizimmer90:
I've got some code to order based on arbitrary indices, but the issue goes a little deeper than the `_topology_from_subset` function.

I can swap the code

```
newTopology = md.Topology()
old_atom_to_new_atom = {}

for chain in topology._chains:
    newChain = newTopology.add_chain()
    for residue in chain._residues:
        resSeq = getattr(residue, 'resSeq', None) or residue.index
        newResidue = None
        for atom in residue._atoms:
            if atom.index in atom_indices:
                try:  # OpenMM Topology objects don't have serial attributes, so we have to check first.
                    serial = atom.serial
                except AttributeError:
                    serial = None
                if newResidue is None:
                    newResidue = newTopology.add_residue(residue.name, newChain,
                                                         resSeq, residue.segment_id)
                newAtom = newTopology.add_atom(atom.name, atom.element,
                                               newResidue, serial=serial)
                old_atom_to_new_atom[atom] = newAtom
```

for 

```
newTopology = md.Topology()
old_atom_to_new_atom = {}

new_chains = [newTopology.add_chain() for chain in topology.chains]

for atom_index in atom_indices:
    atom = topology.atom(atom_index)
    try: # OpenMM Topology objects don't have serial attributes, so we have to check first.
        serial = atom.serial
    except:
        serial = None
    
    #determine chain
    new_chain = new_chains[atom.residue.chain.index]
    
    # determine residue
    resSeq = atom.residue.resSeq
    resName = atom.residue.name
    newResidue = None
    for residue in new_chain.residues:
        if (residue.resSeq == resSeq) and (residue.name == resName):
            newResidue = residue
            break
    if newResidue is None:
        newResidue = newTopology.add_residue(
            atom.residue.name, new_chain,
            atom.residue.resSeq, atom.residue.segment_id)
    
    # add new atom
    new_atom = newTopology.add_atom(
        atom.name, atom.element, newResidue, serial=serial)
    old_atom_to_new_atom[atom] = new_atom
```

however... i realized that the `atoms` property sorts based on a structural hierarchy. From the `Topology` class:

 ```
   @property
    def atoms(self):
        """Iterator over all Atoms in the Topology.

        Returns
        -------
        atomiter : generator
            Iterator over all Atoms in the Topology.
        """
        for chain in self._chains:
            for residue in chain._residues:
                for atom in residue._atoms:
                    yield atom
```

Thus, the topology gets re-scrambled again upon generation of a `Trajectory` object. The only way to have an arbitrary sliced index would be to change this property, but that seems like a pretty big change. Anyone have thoughts on this or can think of a simpler way?

PR #1668, opened 2021-08-21
New OpenMM imports

dwhswenson:
Starting in OpenMM 7.6, currently available as a release candidate, the `simtk` namespace is deprecated. This PR makes it so that the new import names (`openmm`, `openmm.app`, and `openmm.unit`) are the first ones tried, while the old names in the `simtk` namespace are used as backups if the new names fail.

I haven't thoroughly tested the changes to `import_` yet, but I have checked that I can at least import MDTraj in both an OpenMM 7.6 environment and an OpenMM 7.5 environment, and I don't see the warning about the import name changing when I do that import in 7.6.

### DIFF
--- a/mdtraj/reporters/basereporter.py
+++ b/mdtraj/reporters/basereporter.py
@@ -39,14 +39,19 @@ import numpy as np
 
 try:
     # openmm
-    import simtk.unit as units
-    import simtk.openmm as mm
+    import openmm.unit as units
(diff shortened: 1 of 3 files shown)


rmcgibbo:
Cool. Sounds good to me.

rmcgibbo:
Thanks!

PR #1671, opened 2021-09-10
Add atom selection via `atom_indices` to SASA computation, tests added

gph82:
Limit the SASA computation to some atoms of interest via `atom_indices` when calling `md.shrake_rupley`

The excluded atoms are still considered as potential surface accessibility blockers, but their own SASA doesn't get computed and gets set to zero instead. 

This is particularly helpful for membrane-embedded proteins where you want the lipids "there" but you don't really care about their SASA.

### DIFF
--- a/mdtraj/geometry/src/sasa.cpp
+++ b/mdtraj/geometry/src/sasa.cpp
@@ -48,7 +48,10 @@
  *     a bunch of uniformly distributed points on a sphere
  * n_sphere_points : int
  *    the number of sphere points
- *
+ * atom_selection_mask : 1d array, shape[n_atoms]
+ *    one index per atom indicating whether the SASA
+ *    should be computed for this atom (`atom_selection_mask[i] = 1'
+ *    or or not (`atom_selection_mask[i] = 0`)
  * centered_sphere_points : WORK BUFFER 2d array, shape=[n_sphere_points, 3]
  *    empty memory that intermediate calculations can be stored in
  * neighbor_indices : WORK BUFFER 2d array, shape=[n_atoms]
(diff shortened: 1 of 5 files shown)


mdtraj/geometry/src/sasa.cpp
```diff
@@ -63,11 +66,16 @@
  */
 static void asa_frame(const float* frame, const int n_atoms, const float* atom_radii,
                       const float* sphere_points, const int n_sphere_points,
-                      int* neighbor_indices, float* centered_sphere_points, float* areas)
+                      int* neighbor_indices, float* centered_sphere_points, const int* atom_selection_mask, float* areas)
 {
     float constant = 4.0 * M_PI / n_sphere_points;
 
     for (int i = 0; i < n_atoms; i++) {
+        // Skip atom if not in selection
```
rmcgibbo, line 74: Would it be better to just loop over the selection, rather than looping over all the atoms and skipping?

gph82: absolutely, I just don't really know any c++ lol

rmcgibbo: I would:
1. Don't pass in a mask array of booleans, just pas in the list of atom indicies (pointer + length)
2. Make sure to validate that the atom indices are in-bounds.
3. Instead of `for (int i = 0; i < n_atoms; i++)` to loop over the atoms `i`, do `for (int ii = 0; ii < n_atom_indices; ii++) {... i = atom_indices[ii]` to loop over the atoms.

gph82: Done! Only took me three years!

PR #1675, opened 2021-09-26
Added handling of chimera-style indexing for PDB

PolyachenkoYA:
There were already some workarounds for hex indexing by VMD. This fix generalizes the approach so new indexing styles can be added more easily. It also adds a style that is used by [UCSF Chimera](https://www.cgl.ucsf.edu/chimera/)


PolyachenkoYA:
Checks failed, but they were failing for me before my modifications too. And the test that checks my modification didn't fail.

rmcgibbo:
It doesn't look like the checks are failing on the main branch: https://github.com/mdtraj/mdtraj/commit/ff7f345d65eb33d545664b6cff53a9fcf752a5c2

PolyachenkoYA:
You were right, there were some problems due to trailing spaces in some test files. Also, some tests wanted to create atoms from bare lines (without creating a PdbStructure instance). I think everything works now.

I added an optional parameter for atom creation that defines how 6-11 places are converted to an index. I think there is no way to go around it because of the following: Both chimera and vmd can have an index e.g. a0000, but they represent different numbers since it's now hex chimera. So there is no way a reading function can figure out how to decode this on its own. Decimal will be the case for the majority of cases, which is the default behavior.

PolyachenkoYA:
@rmcgibbo , can you please have a look

PolyachenkoYA:
You are correct in #1722 . My fix does address atom indexing in big files and does not address the same for residue indexing. I am not sure why there is no response since the fix passes all the tests.

PR #1686, opened 2021-11-08
[ci] Bump OS versions on github actions

rmcgibbo:
Change test matrix to python38 and python39 rather than python37 and python38.


sroet:
+1 from me

mattwthompson:
No objection here

PR #1688, opened 2021-11-08
[WIP] enable py310 tests

sroet:
This will be my attempt to enable python 3.10 tests, never dealt with nixos before, but seems like a fun thing to try. 

The only package that is not pip installable at the moment is `tables`

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -20,7 +20,7 @@ jobs:
          - py: "python39"
            nixpkgs: "channel:nixos-21.05"
          - py: "python310"
-           nixpkgs: "channel:nixos-21.05"
+           nixpkgs: "channel:nixos-unstable"
 
 
     runs-on: ubuntu-20.04
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -19,6 +19,9 @@ jobs:
            nixpkgs: "channel:nixos-21.05"
          - py: "python39"
            nixpkgs: "channel:nixos-21.05"
+         - py: "python310"
+           nixpkgs: "channel:nixos-21.05"
+
 
     runs-on: ubuntu-20.04
     steps:


rmcgibbo:
It may work better if you switch `"channel:nixos-21.05"` to like `"channel:nixos-unstable"`, but I'm not sure. Probably a _lot_ of builds will be required because the NixOS build system may not be making pre-built python3.10 stuff yet.

rmcgibbo:
This seems like a build that's going to time out: https://gist.github.com/rmcgibbo/6c5335d6ea05615a3b6ce78184b863bd :P

sroet:
> This seems like a build that's going to time out: https://gist.github.com/rmcgibbo/6c5335d6ea05615a3b6ce78184b863bd :P

Surprisingly it didn't xD it just seem to have build a broken install

rmcgibbo:
It seems possible to me that the test in `sh` here: https://github.com/amoffat/sh/blob/eb9d051f1f948fbbc712c7763cc2c72af60deebd/test.py#L1514-L1548 is actually wrong, based on https://bugs.python.org/issue24283, but that it normally passes anyways out of luck.

sroet:
>  It seems possible to me that the test in sh here: https://github.com/amoffat/sh/blob/eb9d051f1f948fbbc712c7763cc2c72af60deebd/test.py#L1514-L1548 is actually wrong, based on https://bugs.python.org/issue24283, but that it normally passes anyways out of luck.

yeah, from those bugs that seems like an illegal test indeed xD I can't reproduce it locally, though ...

sroet:
let's see if opening/closing still retriggers CI

rmcgibbo:
I submitted a PR to nixpkgs to bump tables: https://github.com/NixOS/nixpkgs/pull/153116

rmcgibbo:
Done.

rmcgibbo:
Thanks for doing all the work on this, @sroet.

PR #1694, opened 2021-11-12
Patch util_arm.h for M1

sroet:
[While trying to build for M1 chips](https://github.com/conda-forge/mdtraj-feedstock/pull/37) this patch was required to get the build to work. (originally implemented by @mattwthompson)


mattwthompson:
Thanks for opening this. I thought a little bit about setting up M1 runners in this repo but didn't want to put in the time and work associated with seeing that to the finish line. It might be useful if somebody else is adventurous

Issue #1695, opened 2021-11-13
Remove use of `simtk` imports


dotsdl:
[OpenMM 7.6.0 removed the top-level `simtk` namespace](https://github.com/openmm/openmm/releases/tag/7.6.0). There is a shim in place for now, but use of `simtk` in imports is now deprecated.

dotsdl:
I will take this on.

dwhswenson:
I think I did this in #1668, no?

mattwthompson:
Yep, should be in 1.9.7!

dotsdl:
Ah, thanks folks. Apologies for the noise; I've been working on [`fah-xchem`](https://github.com/choderalab/fah-xchem) and we're seeing `No module named 'simtk.openmm.app.topology'` from our logs. I suspected `mdtraj` first and saw usages of `simtk` in current `master`, but failed to dig much deeper before raising this issue.

I think our problem is coming from [perses](https://github.com/choderalab/perses); apologies for barking up the wrong tree!

PR #1696, opened 2021-11-15
port openmm.unit 

sroet:
## THIS DROPS OpenMM < 7.6

While looking at #1695 by doing 
```grep -r -n "simtk" --include="*.py"``` 

I saw that the `mdtraj/utils/unit` had a bunch of references to it (in docstrings). I figured that the easy solution for that is to re-export the current `openmm.unit` files. I also updated the `__init__.py` to search for `openmm.unit` in `sys.modules`

[edit] after [this comment](https://github.com/mdtraj/mdtraj/pull/1698#issuecomment-969341160) I decided to not increase the maintenance burden and drop support for the `simtk` namespace


sroet:
this was pulled with the merge of #1698.

Issue #1699, opened 2021-11-18
1.9.7 regression, missing zlib.h


Seanny123:
When running `pip install mdtraj==1.9.7` on a minimal Fedora 33 Docker image with Python 3.9, I get the following error:

```
gcc -pthread -Wno-unused-result -Wsign-compare -DDYNAMIC_ANNOTATIONS_ENABLED=1 -DNDEBUG -O2 -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -D_GNU_SOURCE -fPIC -fwrapv -O2 -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -D_GNU_SOURCE -fPIC -fwrapv -O2 -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -D_GNU_SOURCE -fPIC -fwrapv -fPIC -DUSE_ZLIB=1 -Imdtraj/formats/tng/include -I/usr/include -I/tmp/pip-build-env-dgtekvku/overlay/lib64/python3.9/site-packages/numpy/core/include -I/usr/include/python3.9 -c mdtraj/formats/tng/src/lib/tng_io.c -o build/temp.linux-x86_64-3.9/mdtraj/formats/tng/src/lib/tng_io.o
  mdtraj/formats/tng/src/lib/tng_io.c:32:10: fatal error: zlib.h: No such file or directory
     32 | #include <zlib.h>
        |          ^~~~~~~~
  compilation terminated.
  error: command '/usr/bin/gcc' failed with exit code 1
  ----------------------------------------
  ERROR: Failed building wheel for mdtraj
Failed to build mdtraj
[...]

PR #1700, opened 2021-11-18
protect against issue 1699

sroet:
This should protect against issue #1699 for all future releases. As the `c` code is already wrapped in `ifdef USE_ZLIB` this should work.

I tested setting the value to `("USE_ZLIB", 0)` instead of making `def_macros=None`, but that does still trigger the `ifdef`

I tested this by adding a `#include <potato123.h>` on the problematic line of the issue. This did trigger a similar compile error on regular compilation, and not if I hard-coded `def_macros = None`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -101,7 +101,8 @@ def format_extensions():
         # install of python+zlib or a conda install of python+zlib
         zlib_include_dirs += ["{}/include".format(sys.prefix)]
         zlib_library_dirs += ["{}/lib".format(sys.prefix)]
-    zlib = any([glob(f"{i}/zlib.h") for i in zlib_include_dirs])
+    zlib = any([glob(f"{i}/**/zlib.h", recursive=True)
+                for i in zlib_include_dirs])
     # Protect against Issue 1699
     if zlib:
         def_macros = [("USE_ZLIB", 1)]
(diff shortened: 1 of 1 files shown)


sroet:
Hmm, so good news/bad news, we don't build zlib if we can't find it in `f"{sys.prefix}/include"` :smile: , but our tests still try to run it :disappointed: 
Also,  apparently the zlib.h header files are someplace different on nixos, does anyone have a clue were to also look for these headers? :P

rmcgibbo:
I don't think I understand this. Is zlib actually required for the mdtraj code to work? If so, we need it, so I don't see why we should try to compile if its not there. If we don't need it, then why don't we just get rid of it?

sroet:
> Is zlib actually required for the mdtraj code to work?
Only one specific part of it (`tng_io.c`), the `tng_gzip_compress`  function only exists if zlib is installed and `tng_data_read` errors if the data is compressed and the function was compiled without zlib being present (this is triggering the test errors). 

If you don't interact with tng files, mdtraj works perfectly fine without zlib.

rmcgibbo:
It seems to me that the tng format (tng_io.c) can optionally contain blocks compressed with zlib, and we need to build the library with zlib support in order to be able to uncompress those files. I don't want to build/distribute broken versions of mdtraj that don't actually have support for all the formats we claim to support.

sroet:
> I don't want to build/distribute broken versions of mdtraj that don't actually have support for all the formats we claim to support.

:+1:

PR #1706, opened 2021-12-17
Avoid side effects in mdtraj.load

clonker:
using `filename_or_filenames.pop(0)` leads to a change of the filename list which propagates back into the outer scope and is (in my opinion) quite unexpected, i.e., handing in a list of filenames, that list is suddenly shorter by 1 after a call to `mdtraj.load`. This small change uses slicing rather than `pop` which as an added benefit also supports tuples (which do not possess `pop`).


rmcgibbo:
Yes, that looks better

Issue #1714, opened 2022-01-27
Support for PDB chainID


pgbarletta:
Hi, I implemented this minor feature that has came up before a few times (eg: #873). 
I understand from reading some issues that this is not high priority for maintainers, but it is for me and I believe it could be for many people, so I'd deeply appreciate if you let me make a PR to have this built into mdtraj.

The feature is already half-implemented and I did my best to minimize the modifications to the library:
1. `chain_id` is added to the Topology's `Chain` class. It's set by the constructor and it has a default value so the old way of constructing a chain still works.
2. New `Topology.add_chain()` also accepts a `chain_id` with a default value, so everything else works.
3. `PDBTrajectoryFile.read_model()` now calls `add_chain()` with the `chain_id` attribute from `PdbStructure.chain`, which already reads and stores the chainID from the PDB.
4. When writing PDBs `PDBTrajectoryFile.write()` now checks `chain.chainid` before setting `chainName` to A-Z according to `chainIndex`. If the `chainid` attribute from the chain object is set, then it will use that character.

rmcgibbo:
Sounds reasonable. Looking forward to the PR.

pgbarletta:
Thanks!

PR #1715, opened 2022-01-27
Adding PDB chainID support

pgbarletta:
From #873  and #1714 .

### DIFF
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -636,7 +636,7 @@ class Topology(object):
 
         return True
 
-    def add_chain(self):
+    def add_chain(self, chain_id = None):
         """Create a new Chain and add it to the Topology.
 
         Returns
@@ -644,7 +644,7 @@ class Topology(object):
         chain : mdtraj.topology.Chain
             the newly created Chain
         """
-        chain = Chain(len(self._chains), self)
+        chain = Chain(len(self._chains), self, chain_id)
         self._chains.append(chain)
         return chain
 
@@ -1295,13 +1295,15 @@ class Chain(object):
         Iterator over all Atoms in the Chain.
     """
 
-    def __init__(self, index, topology):
+    def __init__(self, index, topology, chain_id = None):
         """Construct a new Chain.  You should call add_chain() on the Topology instead of calling this directly."""
         # The index of the Chain within its Topology
         self.index = index
         # The Topology this Chain belongs to
         self.topology = topology
         self._residues = []
+        # PDB format chainID
+        self.chain_id = chain_id
 
     @property
     def residues(self):
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -636,8 +636,13 @@ class Topology(object):
 
         return True
 
-    def add_chain(self, chain_id = None):
+    def add_chain(self, chain_id=None):
         """Create a new Chain and add it to the Topology.
+
+        Parameters
+        ----------
+        chain_id : str
+            The PDB chainID of the Chain.
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


pgbarletta:
I hadn't run `test_trajectory.py` on my system, but yeah, it seems those tests are failing.

mdtraj/core/topology.py
```diff
@@ -1295,13 +1295,15 @@ class Chain(object):
...
-    def __init__(self, index, topology):
+    def __init__(self, index, topology, chain_id = None):
         """Construct a new Chain.  You should call add_chain() on the Topology instead of calling this directly."""
         # The index of the Chain within its Topology
         self.index = index
         # The Topology this Chain belongs to
         self.topology = topology
         self._residues = []
+        # PDB format chainID
+        self.chain_id = chain_id
```
sroet, line 1306: please add a description of this attribute to the docstring above

mdtraj/core/topology.py
```diff
@@ -1295,13 +1295,15 @@ class Chain(object):
         Iterator over all Atoms in the Chain.
     """
 
-    def __init__(self, index, topology):
+    def __init__(self, index, topology, chain_id = None):
```
sroet, line 1298: [pep8](https://www.python.org/dev/peps/pep-0008/) style:
```suggestion
    def __init__(self, index, topology, chain_id=None):
```

mdtraj/core/topology.py
```diff
@@ -636,15 +636,15 @@ def __eq__(self, other):
 
         return True
 
-    def add_chain(self):
+    def add_chain(self, chain_id = None):
```
sroet, line 639: [pep8](https://www.python.org/dev/peps/pep-0008/) suggestion
```suggestion
    def add_chain(self, chain_id=None):
```

mdtraj/formats/pdb/pdbfile.py
```diff
@@ -332,7 +332,10 @@ def write(self, positions, topology, modelIndex=None, unitcell_lengths=None,
         if modelIndex is not None:
             print("MODEL     %4d" % modelIndex, file=self._file)
         for (chainIndex, chain) in enumerate(topology.chains):
-            chainName = self._chain_names[chainIndex % len(self._chain_names)]
+            if chain.chain_id == None or chain.chain_id == '':
+                chainName = self._chain_names[chainIndex % len(self._chain_names)]
+            else:
+                chainName = chain.chain_id[0:1]
```
sroet, line 338: can be simplified as both `None` and `''` evaluate to `False`
```suggestion
            if not chain.chain_id:
                chainName = self._chain_names[chainIndex % len(self._chain_names)]
            else:
                chainName = chain.chain_id[0:1]
```

sroet (changes requested):
Some code style suggestions.

Also, the test error seems genuine (missing the added `chain_id` from the `Parameters` section in the docstring of that class):
```
ValueError: In mdtraj.core.topology.add_chain, arguments ['chain_id'] don't match Parameters list []
```

mdtraj/core/topology.py
pgbarletta: done

mdtraj/core/topology.py
pgbarletta: funny, my pep8 was messing with the files and I had to turn it off. Done.

mdtraj/core/topology.py
pgbarletta: Done.

mdtraj/formats/pdb/pdbfile.py
pgbarletta: you're right, thanks. Done.

rmcgibbo (approved):

rmcgibbo:
@sroet: any more changes from you?

sroet (approved):
looks good to me

Issue #1541, opened 2020-04-11
ERROR on pip installation


sperezconesa:
Hello,
I am trying to install mdtraj on my Raspberry Pi which has ARM architecture (I am not sure if this is the problem). I am using python 3.7.3. I get the following given bellow.
Thank you very much in advance!
Best,
Sergio

```text
Looking in indexes: https://pypi.org/simple, https://www.piwheels.org/simple
Collecting mdtraj
  Using cached mdtraj-1.9.3.tar.gz (907 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... error
    ERROR: Command errored out with exit status 1:
     command: /home/sperez/mdsrv_env/bin/python /home/sperez/mdsrv_env/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py prepare_metadata_for_build_wheel /tmp/tmp9w06niy8
         cwd: /tmp/pip-install-w5c5eyqe/mdtraj
    Complete output (15 lines):
    Using built-in specs.
    COLLECT_GCC=arm-linux-gnueabihf-gcc
    COLLECT_LTO_WRAPPER=/usr/lib/gcc/arm-linux-gnueabihf/8/lto-wrapper
    Target: arm-linux-gnueabihf
    Configured with: ../src/configure -v --with-pkgversion='Raspbian 8.3.0-6+rpi1' --with-bugurl=file:///usr/share/doc/gcc-8/README.Bugs --enable-languages=c,ada,c++,go,d,fortran,objc,obj-c++ --prefix=/usr --with-gcc-major-version-only --program-suffix=-8 --program-prefix=arm-linux-gnueabihf- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --enable-bootstrap --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-libitm --disable-libquadmath --disable-libquadmath-support --enable-plugin --with-system-zlib --with-target-system-zlib --enable-objc-gc=auto --enable-multiarch --disable-sjlj-exceptions --with-arch=armv6 --with-fpu=vfp --with-float=hard --disable-werror --enable-checking=release --build=arm-linux-gnueabihf --host=arm-linux-gnueabihf --target=arm-linux-gnueabihf
[...]

Issue #1432, opened 2019-02-13
Building for ppc64le


jchodera:
I'm trying to get `mdtraj` working on [ORNL Summit](https://www.olcf.ornl.gov/summit/), which uses `ppc64le` architecture. 

Building the `_rmsd` extension fails with
```
    building 'mdtraj._rmsd' extension
    creating build/temp.linux-ppc64le-3.7/mdtraj/rmsd
    creating build/temp.linux-ppc64le-3.7/mdtraj/rmsd/src
    $BUILD_PREFIX/bin/powerpc64le-conda_cos7-linux-gnu-cc -DNDEBUG -fwrapv -O2 -Wall -Wstrict-prototypes -mcpu=power8 -mtune=power8 -mpower8-fusion -mpower8-vector -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O3 -pipe -I$PREFIX/include -fdebug-prefix-map=$SRC_DIR=/usr/local/src/conda/mdtraj-1.9.2 -fdebug-prefix-map=$PREFIX=/usr/local/src/conda-prefix -DNDEBUG -D_FORTIFY_SOURCE=2 -O3 -fPIC -Imdtraj/rmsd/include -I$PREFIX/lib/python3.7/site-packages/numpy/core/include -I$PREFIX/include/python3.7m -c mdtraj/rmsd/_rmsd.c -o build/temp.linux-ppc64le-3.7/mdtraj/rmsd/_rmsd.o -fopenmp -msse2 -O3 -funroll-loops -Wno-unused-function -Wno-unreachable-code -Wno-sign-compare
    powerpc64le-conda_cos7-linux-gnu-cc: error: unrecognized command line option '-msse2'; did you mean '-misel'?
    error: command '$BUILD_PREFIX/bin/powerpc64le-conda_cos7-linux-gnu-cc' failed with exit status 1
    Running setup.py install for mdtraj: finished with status 'error'
```
Any ideas what should be updated to allow `ppc64le` builds? Presumably, we need to ditch the `-msse2` flags [here](https://github.com/mdtraj/mdtraj/blob/master/setup.py#L167-L169) and [here](https://github.com/mdtraj/mdtraj/blob/master/setup.py#L182-L184) and [here](https://github.com/mdtraj/mdtraj/blob/master/setup.py#L214-L217) on this architecture. Any idea what these should be replaced by, or if other changes might be needed?

marscher:
[...]

Issue #888, opened 2015-07-10
 IBM


euhruska:
Is it possible to get mdtraj working on a IBM cluster?

swails:
Maybe?  Have you tried?

euhruska:
No, it doesn't work

euhruska:
It fails on 
```gcc -pthread -fno-strict-aliasing -g -O2 -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -Imdtraj/rmsd/include -I/home/eh22/local/lib/python2.7/site-packages/numpy/core/include -I/home/eh22/local/include/python2.7 -c mdtraj/rmsd/_rmsd.c -o build/temp.linux-ppc64-2.7/mdtraj/rmsd/_rmsd.o -fopenmp -msse2 -O3 -funroll-loops
cc1: error: unrecognized command line option "-msse2" ```
Is the msse2 necessary?

swails:
MDTraj uses explicit SSE vector instructions for several of its functions (like RMSD and distance calculations, I think).

You can modify `basesetup.py` to set `compiler_args_sse2` and `compiler_args_sse3` to empty lists instead of those compiler flags and see if that works.  I'm not sure that it will, since the IBM chips have a different architecture (and therefore possibly different instruction sets) than the standard x86_64 chips like AMD and Intel.

euhruska:
Doesn't look like that. I gives then errors in mdtraj/rmsd/_rmsd.c.

rmcgibbo:
Can you point me to the documentation for your c/cpp compiler? Does it
support SSE? What are the flags that it takes?

euhruska:
It's a PowerPC64,  doesn't support SSE2. So I guess I have to switch clusters.

jchodera:
Is it possible to skip the optimized Intel-specific code and fall back to a numpy RMSD computation on `ppc64le` architectures?

rmcgibbo:
There's a bunch of SSE intrinsics in both the geometry modules and the RMSD. Most of the geometry modules have a runtime flag to not use the optimized version written in C and fall back to a slow pure python version, which is used inside the test suite. There is no such option for the RMSD. There's also no options in the build system to skip compilation of these modules.

PR #1718, opened 2022-02-08
Implement reading and writing PDBx/mmCIF files

peastman:
Fixes #652.  This implementation uses OpenMM to do most of the work.

PR #1727, opened 2022-03-16
Implement generic C code paths for non-accelerated architectures.

tonigi:
This should enable compilation on ppc64le (at least). 

The setup code now includes a check for SSE2 availability. If neither SSE nor NEON intrinsics are available, the generic C version is used. It can also be enforced anyway at compile time setting env `MDTRAJ_BUILD_DISABLE_INTRINSICS=1`. 

On ppc64le, all the tests seem to pass, with some differences likely due to floating point tolerance. 

Work done by @org305 . 

Relevant issues: https://github.com/mdtraj/mdtraj/issues/888   https://github.com/mdtraj/mdtraj/issues/1541 https://github.com/mdtraj/mdtraj/issues/1432

It may make sense to CI on ppc64le (likely best done in conda feedstock).


rmcgibbo:
Very cool.

How about adding some tests on x86 that compile in this new generic code and also the specialized code and make sure that the numerics are the same?

I'm worried that if we add code that's only compiled on ppc64le, it's going to bitrot since very few of the developers are interested and have access to in those machines.

tonigi:
Hmm... would it make sense to run the full pytest suite in CI?  If so, the CI could build packages and run tests for cases (i.e. with and without `MDTRAJ_BUILD_DISABLE_INTRINSICS`).

rmcgibbo:
Doubling the total test load just for ppc64le seems like perhaps a bit much, but maybe certain tests that actually use these features could be tagged to run with both DISABLE_INTRINSICS and without.

tonigi:
To clarify: MDTRAJ_BUILD_DISABLE_INTRINSICS only affects build behaviour.  Depending on that variable and supported intrinsics in the compiler, only one of the [generic, sse, neon] version is compiled. So currently it's not possible to test  both code versions with one build. 

Here's two options forward, in my opinion (I apologize if I misunderstood mdtraj's CI strategy):
 * in CI, make a non-accelerated build on x86, run just a few tests, then discard the built object - this gives some confidence that the results will be similar in ppc
 * and/or let conda-forge do an actual ppc build, and test it thoroughly
 
The existing tests seem rather stringent to me, and should catch differences in results, should they arise. How is testing handled for the ARM/NEON version?

I understand the risk of bitrot and that supporting minority architectures is undesirable, but lacking mdtraj on some HPC installations (where OpenMM runs nicely) feels like a bit of a "missing piece" that gets in the way.

rmcgibbo:
Yep, I think I understand.

From my perspective, the ideal solution for CI would probably be to add a “job” the to the CI build matrix that compiles in generic mode, but this job should probably only run “relevant” tests, like the ones that use the geometry and RMSD modules.

For aarch64, there are currently no tests on the CI, so I recognize that I might be holding ppc64le to a slightly higher standard. Sorry. To some extent that’s because people own personal aarch64 macs so I trust that the tests suit is getting some exercise there.

tonigi:
Thanks! Something like the above commit?  (It's not a perfect test, because intrinsics dependencies can still sneak in e.g. from system libraries, but should be good enough)

rmcgibbo:
Something like https://github.com/giorginolab/mdtraj/pull/3 is what I have in mind.

tonigi:
Any idea on this? There is a [segfault](https://github.com/giorginolab/mdtraj/runs/5598510332?check_suite_focus=true#step:5:2598) deep in image_molecules C interface, which seems unrelated to the changes. Possibly a Cython+nix version thing.

rmcgibbo:
I think the same thing happened in this PR and then went away after it was merged: https://github.com/mdtraj/mdtraj/pull/1718

rmcgibbo:
Eh, let's merge and roll the dice.

tonigi:
Apparently it worked. My hunch is that sometimes nix packs a subtly corrupt version, and the failure remains until a new artifact is built. Diffing  the logs was not especially enlightening. Thanks for merging though!

PR #1728, opened 2022-04-01
Adds lower-level `compute_distances_core` function

salotz-sitx:
This adds a lower level interface to the existing `compute_distances`
function in `geometry.distance` that allows the user to pass in only
the position coordinates (and box information if needed) rather than
only having the Trajectory interface.

This allows for implementing algorithms which can easily add in
virtual atoms (such as centers of mass) without needing to actually go
through changing the Topology object of a Trajectory.


rmcgibbo:
Makes sense. Thanks!

PR #1741, opened 2022-05-21
BLD: Add a github actions workflow for building wheels and source distributions

BvB93:


BvB93:
My apologies, I accidentally created a PR while experimenting

PR #1751, opened 2022-08-31
Use oldest-supported-numpy instead of numpy version pinning for build system

chajath:
Older numpy versions pinned were causing issues when those numpy versions were not compatible with newer mac platform.

Instead, let's use https://github.com/scipy/oldest-supported-numpy which does the right thing for most known build envs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,11 +2,12 @@
 requires = [
     "setuptools",
     "wheel",
-    "numpy==1.11.3; python_version=='3.6'",
-    "numpy==1.14.6; python_version=='3.7'",
-    "numpy==1.17.5; python_version=='3.8'",
-    "numpy==1.19.5; python_version=='3.9'",
-    "numpy==1.21.3; python_version=='3.10'",
+    "numpy; sys_platform=='darwin'",
+    "numpy==1.11.3; python_version=='3.6' and sys_platform=='linux'",
+    "numpy==1.14.6; python_version=='3.7' and sys_platform=='linux'",
+    "numpy==1.17.5; python_version=='3.8' and sys_platform=='linux'",
+    "numpy==1.19.5; python_version=='3.9' and sys_platform=='linux'",
+    "numpy==1.21.3; python_version=='3.10' and sys_platform=='linux'",
     "numpy; python_version>='3.11'",
     "Cython>0.28",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,12 +2,10 @@
 requires = [
     "setuptools",
     "wheel",
-    "numpy; sys_platform=='darwin'",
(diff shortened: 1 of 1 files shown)


sroet:
> I'm a bit surprised to see such hard pinning, I personally prefer adding lower constraints until it's known that an upper constraint is also needed. 

The reason for these extreme pinnings is to solve issues like #1617, the problem/solution are discussed from [this comment](https://github.com/mdtraj/mdtraj/issues/1617#issuecomment-771006474) onward.

The TLDR is that `pip` will pull the most recent version of `numpy` it can during building (in build-isolation), completely disregarding whatever `numpy` is present in the environment. However, `numpy` is only backwards compatible; newer `numpy` will run code that was build against an older `numpy` not the other way around.

 This would lead to broken builds if `mdtraj` was installed in an environment that does not have the most recent `numpy` installed.

chajath:
I see. It puts devs like me in a rough spot because if we are tied to an older python version, older NumPy doesn't play nicely with M1 machine. I suggest building wheels per each major NumPy version we want to support. Or, can we relax NumPy pinning on all versions, then advertise users to do `pip install --no-build-isolation` to use whatever NumPy version they have in runtime to do the building?

chajath:
Or, maybe we should use https://github.com/scipy/oldest-supported-numpy like sklearn does. This looks like a good solution for everyone.

sroet:
> Or, maybe we should use https://github.com/scipy/oldest-supported-numpy like sklearn does. This looks like a good solution for everyone.

The current implementation was based on the `scipy` implementation at the time, so this would be my preferred solution. 

However, I am not a recent user of `mdtraj` anymore, so my preference should not be taken too seriously.

mattwthompson:
> It puts devs like me in a rough spot because if we are tied to an older python version, older NumPy doesn't play nicely with M1 machine.

Is there no way for `pip` to use Rosetta to pull down older versions? I know `conda` can switch between `osx-64` and `osx-arm64` for this sort of thing

chajath:
I would rather keep the code portable and don't have to resort to Rosetta. There might be a way, but then I would have to maintain two different toolchains and try to cross-compile and all, it will be a mess tbh.

chajath:
Btw while we are on the topic of platforms, wonder if we should get onboard tools like https://github.com/pypa/cibuildwheel so we publish all bdist wheels and don't have to manage all the different platforms ourselves.

cclauss:
scipy/oldest-supported-numpy#64 is now tested on Python 3.11 release candidate 2.  See #1752

mattwthompson:
If I can ever get #1758 over the finish line, we should take another look at this. I think it's probably a good solution.

mattwthompson (approved):

mattwthompson:
I think this should be good to go - will merge sometime if no objections

PR #1752, opened 2022-09-11
Add Python 3.11 release candidate 2 to testing

cclauss:
Related to #1754

Python 3.11 will be on average 22% faster than Python 3.10 so let’s give it a spin...
https://www.python.org/download/pre-releases

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -19,7 +19,7 @@ jobs:
            intrinsics: native
          - py: "python310"
            intrinsics: native
-         - py: "python311-dev"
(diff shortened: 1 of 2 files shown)


mattwthompson:
I'm unfortunately not familiar enough with Nix/NixOS to help here ... I think MDTraj is packaged externally and pulled down here, so it might need to be rebuilt with the RC. I'm not sure.

cclauss:
Based on the error message, I believe that we need to rebuild with a [current version of Cython](https://pypi.org/project/Cython).

PR #1759, opened 2022-11-03
fix bug: resSeq changed when doing trajectory.atom_slice if resSeq==0

huhlim:
When I was doing trajectory.atom_slice, if a residue's resSeq == 0, the line 93 reassigns resSeq as residue.index because
```python
resSeq = 0 or residue.index # =residue.index
```
As we can assign a default value to the [getattr](https://docs.python.org/3/library/functions.html#getattr) method like
```python
getattr(object, name, default, /)
```
, it would be better to use this instead.

### DIFF
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -90,7 +90,7 @@ def _topology_from_subset(topology, atom_indices):
     for chain in topology._chains:
         newChain = newTopology.add_chain()
         for residue in chain._residues:
-            resSeq = getattr(residue, 'resSeq', None) or residue.index
+            resSeq = getattr(residue, 'resSeq', residue.index)
             newResidue = None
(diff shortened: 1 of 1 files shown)


sukritsingh:
Sorry this fell through the cracks! It looks like you also have added on a bit more parsing code here that alters how things are parsed as well on top of the change you brought up (and I'm not seeing an issue linked here either). 

Perhaps you could better add comments at each change/code block to elaborate on the improvements made. Just so I can get a better picture of what's going on here. 

Not sure if this is irrelevant now (apologies if so!) - just wanted to make sure it got looked at! Happy to close if also irrelevant.

jeremyleung521:
Cherry-picked all the relevant commits into #2013 because your master branch is getting a little messy. :) Closing this now.

Issue #1753, opened 2022-09-12
atom_slice: provides a warning when input indices are not sorted or not unique


yaoyic:
Recently I have been puzzled by the index order issue of `Trajectory.atom_slice` in my project, and it took me sometime to track down the cause:

I want to change the order of some atoms in a residue, such that I use the structure with another force field. Intuitively, I thought it could be done by `atom_slice` method, but in the end I found the coordinates have been correctly reordered, but not the topology, which ends up in the output as wrong labeling for the swapped atoms.

The problem is that the internal logic is inconsistent for trajectory and topology slicing. For the former, it views the provided atom_indices as an ordered list.
https://github.com/mdtraj/mdtraj/blob/1a757274f6f9acd6c9c847c72f5857e9c1046d17/mdtraj/core/trajectory.py#L1804
But for the latter, it views atom_indices as a unordered set,
https://github.com/mdtraj/mdtraj/blob/1a757274f6f9acd6c9c847c72f5857e9c1046d17/mdtraj/core/trajectory.py#L1807
https://github.com/mdtraj/mdtraj/blob/1a757274f6f9acd6c9c847c72f5857e9c1046d17/mdtraj/core/topology.py#L96
and thus the output follows the original topological order. Therefore, there will be an inconsistency between the coordinates and the topology, when the provided atom_indices are not an sorted list.

It seems to be a long-lasting issue (see e.g., #1343) and I can understand that there might not be a simple solution to handle input that is not sorted or has repeating indices. But I think it is worth noting in the docstring, giving a `Warning` or raising an `Error` in the python code when such an input is provided, rather than silently introducing this inconsistency.

Last by not least, thanks for writing and maintaining this package! I have always found it very helpful and handy.

mattwthompson:
I'm neutral on a change that would fix/update this behavior. I'd certainly review a PR that takes a stab at it. It might take a bit of refactoring since the order at the new atoms (`atom_indices`) is determined by the atom ordering in the old topology ([L90-95](https://github.com/mdtraj/mdtraj/blob/1a757274f6f9acd6c9c847c72f5857e9c1046d17/mdtraj/core/topology.py#L90-L95), the outer loops from what you posted). In general, re-ordering atoms in existing objects is playing with fire and I'm not surprised that the bookkeeping has a mismatch.

[...]

PR #1766, opened 2023-02-21
Warn if atom_indices are not monotonically increasing

CharlieLaughton:
As per issue #1753 a fix so that a warning is raised if atom_indices are not monotonically increasing. There is no other tweaking, so existing behaviour continues: atom_indices like [5,4,3,2,1] will be accepted but the resulting topology will be bogus, and atom_indices like [1,2,3,3,4] will lead to a ValueError later in the code, when the subsetted coordinates and topology are found not to match up.

### DIFF
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -54,6 +54,7 @@ import numpy as np
 import os
 import xml.etree.ElementTree as etree
 from collections import namedtuple
+import warnings
 
 from mdtraj.core import element as elem
 from mdtraj.core.residue_names import (_PROTEIN_RESIDUES, _WATER_RESIDUES,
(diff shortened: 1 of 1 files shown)


mdtraj/core/topology.py
```diff
@@ -84,6 +85,13 @@ def _topology_from_subset(topology, atom_indices):
     atom_indices : array_like, dtype=int
         The indices of the atoms to keep
     """
+    np_indices = np.array(atom_indices)
+    if len(np_indices) > 1:
+        if (np_indices[1:] - np_indices[:-1]).min() < 1:
+            warnings.warn('atom_indices are not monotonically increasing')
+        if len(np.unique(np_indices)) < len(np_indices):
+            warnings.warn('atom_indices are not unique')
```
mattwthompson, line 93: Could you provide some rough benchmarks that look at how much this slows down common operations (i.e. loading topologies) for a few different-sized systems? These checks should be quick but they're O(N) and called in most user workflows

CharlieLaughton: Using the topologies from the AMBER20 Benchmark suite (http://ambermd.org/GPUPerformance.php)  I get the following figures for mdt.load_topology():

System         n_atoms    1.9.7       #1753
STMV            23558        0.248       0.285
Factor IX      90906        0.807       0.880
JAC              408609        3.804       3.712
Cellulose 1067055         9.590      9.567

Times (in seconds) are averages over 5 repeats, jobs run on my intel i5 laptop

mattwthompson: Interesting ... I can't imagine this change would actually speed anything up so it's a bit of a surprise that some runs a faster. I suppose the slowdown is simply less (much less) than the variance of 5 runs.

CharlieLaughton: Exactly - I didn’t actually calculate the variance but as I recall there was a fair bit of jitter between runs - these are wallclock times by the way, so that could account for some of it.

Issue #1770, opened 2023-03-14
Iterload fails for large DCD files on Windows


reedharrison:
I encountered an issue using mtraj.iterload with large DCD files on Windows (~12 GB) that results in an OSError as follows:

> File mdtraj\formats\dcd\dcd.pyx:221, in mdtraj.formats.dcd.DCDTrajectoryFile.__cinit__()
> OSError: Could not open file: 6m0j_P1_trajectory.dcd

This DCD file in particular is about 12 GB, and can be loading in chunks successfully on Linux:

`for chunk in md.iterload(filename='6m0j_P1_trajectory.dcd', chunk=100, top=psf):`
`    print(chunk)`

Smaller DCD files can also be read fine on Windows with iterload, so it seems to be a problem specific to large DCD files.

What I think is happening is that "open_dcd_read" is failing at line 219 in "mdtraj/formats/dcd/dcd.pyx", but I'm not sure what the exact problem is. Has anyone run into a problem like this before?

reedharrison:
Tried updating dcdplugin.c to a more recent version. Had to make some additional changes to get it to compile, but it does allow mdtraj.iterload to read the 12 GB DCD successfully. Could use more testing to make sure these changes didn't break anything else.

mattwthompson:
Could you please submit the changes as a PR?

reedharrison:
Closing this issue since it appears fixed.

PR #1771, opened 2023-03-15
Update dcdplugin.c

reedharrison:
Update dcdplugin.c to handle large DCD files more reliably on Windows

From #1770

### DIFF
--- a/mdtraj/formats/dcd/src/dcdplugin.c
+++ b/mdtraj/formats/dcd/src/dcdplugin.c
@@ -1,6 +1,6 @@
 /***************************************************************************
  *cr                                                                       
- *cr            (C) Copyright 1995-2009 The Board of Trustees of the           
+ *cr            (C) Copyright 1995-2016 The Board of Trustees of the           
  *cr                        University of Illinois                       
  *cr                         All Rights Reserved                        
  *cr                                                                   
@@ -11,7 +11,7 @@
  *
  *      $RCSfile: dcdplugin.c,v $
  *      $Author: johns $       $Locker:  $             $State: Exp $
- *      $Revision: 1.76 $       $Date: 2011/05/18 17:29:19 $
+ *      $Revision: 1.88 $       $Date: 2020/12/17 17:14:07 $
  *
  ***************************************************************************
  * DESCRIPTION:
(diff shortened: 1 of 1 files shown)


reedharrison:
Well the mdtraj tests ran fine during testing (as well as the system from #1770 ), but it would be nice to know if it works for others. Haven't run into any issues using it yet, but let me know if there is something you would like to test specifically.

mattwthompson:
It'd be nice to re-jigger CI to also run on Windows. I think MDTraj worked on Windows in the past, not sure.

sukritsingh:
Lol looks like the Linter is being annoying here and making rerunning CI challenging

mattwthompson:
I don't think (:tm:) the linter should be seeing C files, so just accepting the changes in the feature branch should probably work. Maybe the issue is trailing whitespace, which can safely be re-run at any point

rohithmohan:
Sorry to bump this but wanted to check to see if there's any updates on this PR?

janash:
Thanks for the bump! We'll take a look at it. 

@reedharrison or anyone familiar - what is the source of `dcdplugin.c`? It seems like VMD, but it would be helpful to have more context. I see matching code [here](https://www.ks.uiuc.edu/Research/vmd/plugins/doxygen/dcdplugin_8c-source.html)

In #1770, I see that @reedharrison commented

> Had to make some additional changes to get it to compile 

@reedharrison - are you able to remember what those changes are? I think it would be useful to document.

I am fine with merging this assuming tests pass, although I see that the are conflicts now with the main branch that will have to be resolved.

janash:
Thanks @reedharrison! The diff seems to be for the previous plugin version with your ported version rather than the current "raw" plugin version vs. the MDTraj one (is that correct or no?). 

I'm in favor of getting this merged. However, I think we should include a copy of the [VMD plugin license](https://www.ks.uiuc.edu/Research/vmd/plugins/pluginlicense.html) and modify the header to reference it. I'd also like to see a list of modifications made for MDTraj in the header.

janash:
@reedharrison - of course :) Could you make the modifications to the header + add the license that I mentioned?

reedharrison:
Also, it isn't clear to me how you want to handle the license. Is it sufficient to include the URL to the plugin license in dcdplugin.c? Or did you want to add a text file somewhere with the license copied? If the latter, can you please specifically mention where you would like the file placed? MDtraj has been using a modified copy of the plugin for quite some time, but I don't see any mention of the plugin license.

janash:
I'd be happy to clarify. 

I do think we need to include the license file to be fully compliant with the license. It would be acceptable to include it in the same directory as the source file with the name `LICENSE`. Although MDTraj hasn't traditionally included this, it is probably good to include it as long as we are updating this part of the code.

As far as documenting the changes, although we didn't have this previously for this specific plugin, there is an approach taken with some other parsers (for example the PDB parser) which is ported from OpenMM to add a small [README](https://github.com/mdtraj/mdtraj/tree/main/mdtraj/formats/pdb) in the directory that outlines the source and modifications. 

Thinking on this more, in order to be consistent with this pattern, it would be ideal to have a README that listed modifications made to the plugin. A slight expanded version of what you said before would be acceptable:

> For example, I had to remove some instances of the "static" keyword. I also added back in the "dcd_nsets" and "dcd_rewind" functions (which were in the original dcdplugin.c from MDtraj, I believe). There was also a block of code that only seemed relevant to VMD (the VMD "initialization stuff", so I left that out.

Thanks again for your contribution!

reedharrison:
@janash , should be done now. Let me know if there is anything else. Thanks!

janash (changes requested):
Hi @reedharrison - thanks very much for your contribution.

@bennybp (another MolSSI Software Scientist) has also looked at/helped with this review, so I'm tagging him here.

On reviewing the code in-depth, I see that previously, we were using `dcdplugin.h` in an `include` folder, whereas now everything from the header is inline in `dcdplugin.c`. I can see that this is directly from the updated plugin code. 

See [here](https://github.com/mdtraj/mdtraj/blob/main/mdtraj/formats/dcd/include/dcdplugin.h)

This was included in the edited file `dcdplugin.c` AND is also used in [dcdlib.pxd](https://github.com/mdtraj/mdtraj/blob/main/mdtraj/formats/dcd/dcdlib.pxd), which is used by Cython to define the Python interface (I believe). I think a header file is needed here just to provide the function definitions for compilation.

We run into a bit of an issue here because the Python interface is defined in `dcdlib.pxd` which looks at `include/dcdplugin.h` for the rules on the function declarations. However, the actual function definitions themselves aren't linked at all to either of these files.

Considering that, we have to check that function signatures are consistent across all files. For example, in `dcdlib.pxd`, we have

```python
dcdhandle* open_dcd_write(const char *path, const char *filetype, const int natoms, const int with_unitcell)
```
[here](https://github.com/mdtraj/mdtraj/blob/4788995cec889c9cebbdf91c871f63c5a9f4cfb7/mdtraj/formats/dcd/dcdlib.pxd#L24)

in `include/dcdplugin.h`, we have

```python
dcdhandle* open_dcd_write(const char *path, const char *filetype, const int natoms,
                          const int with_unitcell);
```
[here](https://github.com/mdtraj/mdtraj/blob/4788995cec889c9cebbdf91c871f63c5a9f4cfb7/mdtraj/formats/dcd/include/dcdplugin.h#L70C1-L71C52)

while in the new verision of `dcdplugin.c`, we are missing the `with_unitcell` parameter

```python
dcdhandle* open_dcd_write(const char *path, const char *filetype,
    int natoms)
```  

While the tests are passing, this may be because the code defaults to having a unit cell almost always because this decision is now made in the function based on an environment variable. The intended behavior (and that implemented previously) was for it to be passed in as a function parameter. (I'm not sure if I understand why we don't see a problem with CI, but I believe this is a problem regardless that may show up only under certain conditions)

**Before merging**  
1. Change `dcdplugin.c` to use `dcdplugin.h`. This would involve removing the inline function definitions and making sure function signature are consistent across files.
2. Make behavior of `open_dcd_write` match previous behavior, as in read `with_unitcell` from function input rather than a VMD environment variable.
3. The functions `write_timestep` and `close_file_write` also have modified signatures, going from `dcdhandle *v` to `void *v`. I'd prefer this to be `dcdhandle *v` as it was previously.

Considering that you started this PR years ago now, I understand that this may not be your priority. **Please let us know if you would prefer for one of the maintainers to finish this.**

bennybp:
My only question is: the signature to `open_dcd_write` has changed in the C file (removing the `with_unitcell` parameter). Now it basically defaults to `true` unless an env var is set. Two ways to handle this

1. Propagate the change through to the python side. However, it does seem to be set to true/false depending on some logic with cell lengths/angles [here](https://github.com/mdtraj/mdtraj/blob/6b59e16f2f9362b9b99bc2313d1136688f6516a5/mdtraj/formats/dcd/dcd.pyx#L596)

3. Add the parameter back in, and use that value unless the environment variable is set, which would force it to be false.

I would lean towards 2, but any strong feelings from anyone?

bennybp:
I've added the parameter back in. It was almost certainly intentionally put there as a modification to the VMD code, so that `with_unitcell` could be passed in. Also added a comment about why.

I also removed the duplicated code from the source file to the header, and included the header.

Should be ready to go as long as the tests pass.

PR #1785, opened 2023-06-10
Remove stale bot

mattwthompson:
It's been suggested that the stale bot does more harm than good, mostly because closing an issue for inactivity implies it was resolved when it likely wasn't. I don't feel strongly either way; I think it's as simple as removing a file, so here that is.


sukritsingh:
Yeah I agree that in the long-term it'll be useful to remove this bot (I feel like it just causes spam). However, it may be useful just for a moment longer as @apayne97 and I catch up with all the issues and go through the older ones. 

Mind if I take a hot second to merge this in, and in the meantime try and clean up some of the older issues marked by the stalebot?

mattwthompson:
That sounds like a reasonable plan to me. If left to my own devices with feedback as it is (two thumbs up, no commented objections) I'd merge it in a week or two. If you'd prefer, I'll just let you merge this whenever you see fit. Just confirm that's the case and I won't worry about doing it myself.

The only suggestion I'd make to you two, if you want to go through existing and old issues, is to be selective in the old issues you revive. McGibbon commented at some point that most of his time maintaining MDTraj was spent dealing with support requests and not much to do with changing the code. I think it's fine to let those disappear into the ether, or possibly divert them to the Discussions feature or something like that, since users will eventually move on if they can't get their thing to work. Discrete bugs, feature requests, discussion threads, etc. are worth keeping around, though.

sukritsingh:
> If left to my own devices with feedback as it is (two thumbs up, no commented objections) I'd merge it in a week or two. If you'd prefer, I'll just let you merge this whenever you see fit. 

Yep! Happy to confirm that I'll merge it in myself (there's no conflicts anyways) - I think my timeline would be very similar to yours anyways, pending no distracting emergencies. 

> if you want to go through existing and old issues, is to be selective in the old issues you revive.

Agreed! I think we're pretty aligned on what the issues should look like - if anything I'd like to actually be ruthless and close as many issues as possible, probably most likely telling folks to pivot to the Discussions feature or at least giving it a once-over. I think mainly I want to go through all the "old" discussions and make sure I've had my eyes on everything once.

mattwthompson:
+1 to setting up Discussions at some point; I don't have the project permissions at the moment but I'm sure somebody can set it up or run down the people with the authority to pass them on

sukritsingh:
I have permissions to at least set up Discussions. I'll give it a look over, but the suggested default categories looked good to me to start - let me know if you think there are any additional topics to be made!

sukritsingh:
Ok I've been very behind on issue tracking so I'm just going to go ahead and merge this in! Thanks for doing this (and for cutting the new release!)

Issue #1800, opened 2023-07-17
Cython 3 breaks cythonization of _geometry.pyx


jeremyleung521:
[Cython 3](https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html) appears to break cythonization of `mdtraj/geometry/src/_geometry.pyx`.  Reproducible on MacOS and Linux.

To reproduce, install mdtraj without the wheels:
`python -m pip install mdtraj --no-binary :all:` 

Pinning cython to <3 in `pyproject.toml` will prevent the error from happening, but suggests some changes in `_geometry.pyx` is needed for cython 3 support.

```
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [160 lines of output]
      Apple clang version 14.0.3 (clang-1403.0.22.14.1)
      Target: arm64-apple-darwin22.5.0
      Thread model: posix
      InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
      /Users/User/Documents/Important/Pitt/git/mdtraj/basesetup.py:451: UserWarning: Unrecognized setuptools command ('dist_info --output-dir /private/var/folders/06/29rlbptd5jq_rfl__96vd2qh0000gn/T/pip-modern-metadata-a0z_scx1 --keep-egg-info'), proceeding with generating Cython sources and expanding templates
        warnings.warn("Unrecognized setuptools command ('{}'), proceeding with "
      warning: mdtraj/formats/dcd/dcdlib.pxd:12:12: Non-trivial type declarators in shared declaration (e.g. mix of pointers and values). Each pointer declaration should be on its own line.
      warning: mdtraj/formats/dcd/dcdlib.pxd:12:16: Non-trivial type declarators in shared declaration (e.g. mix of pointers and values). Each pointer declaration should be on its own line.
      warning: mdtraj/formats/dcd/dcdlib.pxd:12:20: Non-trivial type declarators in shared declaration (e.g. mix of pointers and values). Each pointer declaration should be on its own line.
      
      Error compiling Cython file:
      ------------------------------------------------------------
      ...
[...]

PR #1801, opened 2023-07-18
Constrain `cython=0.29`

mattwthompson:
Band-aid for #1800, which should be followed-up by a proper fix.

Version 3.0 (!) was released, but the current source code is only compatible with versions 0.29.x

### DIFF
--- a/devtools/conda-envs/test.yaml
+++ b/devtools/conda-envs/test.yaml
@@ -5,7 +5,7 @@ channels:
 dependencies:
   # Core
   - python
-  - cython
+  - cython =0.29
   - numpy
   - zlib
   - pyparsing
--- a/devtools/conda-envs/test.yaml
+++ b/devtools/conda-envs/test.yaml
@@ -5,7 +5,7 @@ channels:
 dependencies:
   # Core
   - python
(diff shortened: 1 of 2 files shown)


mattwthompson:
So it was passing with 0.29.36 a few days ago:
https://github.com/mdtraj/mdtraj/actions/runs/5570950610/jobs/10175514121#step:3:272

It's now failing with ... 0.29.36: https://github.com/mdtraj/mdtraj/actions/runs/5588508191/jobs/10215400973#step:3:272

Something else must have changed - I haven't looked deeper

mattwthompson:
Oh, it's probably that I'm looking at the Cython in the conda environment, but a different one is used for the build process

mattwthompson:
Thanks for the help @jeremyleung521 - this looks good to me, what do you think?

jeremyleung521:
I suggest doing `==0.29.*` or `~=0.29.36` just in case there is a newer `0.29.*` bugfix somewhere down the line, but other than that, it looks good to me!

mattwthompson:
Oh, nice - conda doesn't support the tilde specifier so I didn't think of that

PR #1807, opened 2023-07-30
Use raw strings to fix invalid escape sequences

speleo3:
Add [ruff](https://beta.ruff.rs/) check of [invalid-escape-sequence (W605)](https://beta.ruff.rs/docs/rules/invalid-escape-sequence/) linter rule to CI pipeline.

Fixes https://github.com/mdtraj/mdtraj/issues/1806

Issue #1806, opened 2023-07-29
DeprecationWarning: invalid escape sequence '\g'


speleo3:
I'm getting several "DeprecationWarning: invalid escape sequence".

These can also be detected using a linter like `ruff` or `flake8` with the [W605](https://beta.ruff.rs/docs/rules/invalid-escape-sequence/) rule:
```sh
ruff check --select W605 mdtraj
```
```sh
flake8 --select W605 mdtraj
```

Most of these strings come from math expressions in docstrings, but also several regular expressions are affected.

The easy solution is to use [_raw string_](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals) syntax (`r` prefix) for those strings.

I'd be happy to make a pull request.

mattwthompson:
A PR that raw-ifies regex strings would be welcome, I expect.

I have a slight preference for leaving the existing math expressions in place where possible - I assume `flake8` is just not good with parsing TeX stuff - even if that requires workarounds like `# noqa` or turning off that rule for particular files.

It's worth exploring `flake8` or similar for the entire codebase at some point; no need to do that here

speleo3:
There are actual errors in the TeX strings which would get fixed by raw strings. Example:

Docstring of [mdtraj.asphericity](https://www.mdtraj.org/1.9.8.dev0/api/generated/mdtraj.asphericity.html?highlight=asphericity#mdtraj.asphericity), current:
![asphericity-current](https://github.com/mdtraj/mdtraj/assets/1239490/aa970109-8536-43e8-8f58-185ede3e7c03)

After raw string fix:
![asphericity-raw](https://github.com/mdtraj/mdtraj/assets/1239490/e48e6603-23d1-44ec-afda-815b2fa87c41)

PR #1811, opened 2023-07-31
Add ruff check to CI

speleo3:
Currently checks a single rule (W605) to identitfy invalid escape sequences as reported in https://github.com/mdtraj/mdtraj/issues/1806

Same as https://github.com/mdtraj/mdtraj/pull/1807 but without the code fixes, so the pipeline is expected to fail.


speleo3:
I would also suggest to add a [ruff pre-commit hook](https://beta.ruff.rs/docs/usage/#pre-commit).

speleo3:
The intended CI failure with W605 violoations can now be seen in https://github.com/mdtraj/mdtraj/actions/runs/5715704737/job/15486056503

I noticed that this makes all matrix jobs fail, which is a bit unfortunate. The alternative is to make ruff a separate job, not part of the matrix. I've pushed e0a6b86fd21654e4a7d28ef5c3a0594e7128356b for comparison.

jeremyleung521:
Apparently we're already running ruff on pre-commit, so this PR is no longer needed. We did catch a lot of lint errors though!

PR #1815, opened 2023-08-30
Handle hex residue numbers in CONECT records

zkagin:
Adds a small fix to handle hex residue numbers in CONECT records. mdtraj already handles this elsewhere when parsing ATOM records, but the current CONECT logic will raise an error if those residue numbers show up in a CONECT row as well.

Tested this on a PDB file with CONECT records and validated that it successfully sets `self._current_model.connects` and does not raise an error.

### DIFF
--- a/mdtraj/formats/pdb/pdbstructure.py
+++ b/mdtraj/formats/pdb/pdbstructure.py
@@ -201,13 +201,16 @@ class PdbStructure(object):
                 self._unit_cell_angles = (float(pdb_line[33:40]), float(pdb_line[40:47]), float(pdb_line[47:54]))
 
             elif (pdb_line.find("CONECT") == 0):
-                atoms = [int(pdb_line[6:11])]
-                for pos in (11,16,21,26):
+                atoms = []
+                for pos in (6,11,16,21,26):
                     try:
                         atoms.append(int(pdb_line[pos:pos+5]))
                     except ValueError:
-                        # Optional field, don't worry if it isn't defined
-                        pass
+                        try:
+                            atoms.append(int(pdb_line[pos:pos+5], 16))
+                        except ValueError:
+                            # Optional field, don't worry if it isn't defined
+                            pass
 
                 self._current_model.connects.append(atoms)
         self._finalize()


zkagin:
@mattwthompson @sukritsingh Would one of you be able to review this small PR? Thank you!

sukritsingh:
Sorry for the radio silence! Just digging myself out of a grant writing backlog now - let me take a look this week!

sukritsingh:
So I'm digging through the code now but I'm still not entirely sure what the issue is since CONECT records (to my experience) contain atom indexes and not residue IDs. I've also never encountered the use of "hex residue numbers" before (maybe others have!) - any chance you could include some sample files or tests into this PR so I can get a sense of what this you want to ensure happens?

peastman:
You're correct, CONECT records contain atom numbers, not residue numbers.  @zkagin is that what you meant to say?

Using hex is a nonstandard hack to get around limits of the PDB format.  It only provides 5 columns for storing atom numbers, which limits you to under 100,000 atoms.  To work around that, some programs store atom indices in hex.  Different programs do it in different ways, which makes them hard to interpret consistently.  But in this case that shouldn't be a problem.  All that really matters is whether the value stored in a CONECT record matches the value found in an ATOM record.

The change in this PR could cause problems.  It interprets all atom indices as integers, but parses some as decimal and others as hex.  That could lead to conflicts where two different values get interpreted as the same number.  What about just recording them as arbitrary strings and seeing if they match the ATOM records?

zkagin:
Yes, my apologies - it's the atom serial number. I believe this happens (in our case) when a ligand is added to an already solvated system, leading to more than 10K atoms and the ligand with the CONECT records being at the end of it. An example snippet from the PDB file is:

```
HETATMB34AA  C8  LIG D   1     -17.486  21.604  12.011  1.00  0.00           C  
HETATMB34AB  O1  LIG D   1     -17.223  21.701  13.209  1.00  0.00           O  
HETATMB34AC  N2  LIG D   1     -18.074  22.637  11.306  1.00  0.00           N  
...
CONECTB34AAB34A9B34ABB34AC
CONECTB34ABB34AA
```

I completely agree that converting to hexadecimal is a patch to retrofit an old file format to larger systems. However, it seems to be relatively commonplace among libraries, and mdtraj already handles atom serial numbers that sometimes use hexadecimal in other parts of the code, like [here](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/pdb/pdbstructure.py#L679). I tried to minimize the changes in this PR since right now it just errors when encountering these atoms in a CONECT record, but if necessary we can make a more expansive change if this behavior isn't safe?

sukritsingh:
I think if we're making changes to how we parse CONECT records, particularly with using Hex values, it may be good to include some tests to ensure expected behavior is maintained before after the fix? I think that would be most convincing to me, since this PR does introduce some changes that may be useful in only a small subset of cases.

> What about just recording them as arbitrary strings and seeing if they match the ATOM records?
I like this idea as well!

zkagin:
Unfortunately I don't have the bandwidth to continue pursuing this fix into master. We've patched it in an internal fork and it works correctly for us, and the Hex workaround is already in use in [other parts of mdtraj](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/pdb/pdbstructure.py#L686). Please feel free to resurrect this if you or someone else runs into it.

Issue #1825, opened 2023-10-26
MDTraj reader not compatible with latest GSD


daico007:
One of my labmate bumped into an issue when trying to load in a `gsd` file using `mdtraj`. The error traceback indicates the issue arises from the `gsd.hoomd.open()` step, (invalid mode 'wb'). This is due to the breaking change in open mode options in  `gsd` >= 3 (https://gsd.readthedocs.io/en/stable/python-module-gsd.hoomd.html), basically doing away with `rb` and `wb`.

PR #1826, opened 2023-10-26
Update syntax for GSD reader/writer

daico007:
Address #1825, update syntax in `gsd.py` to match with the latest GSD 3 open syntax (https://gsd.readthedocs.io/en/stable/python-module-gsd.hoomd.html).

### DIFF
--- a/devtools/conda-envs/test.yaml
+++ b/devtools/conda-envs/test.yaml
@@ -14,7 +14,7 @@ dependencies:
   - networkx
   - pandas
   - pytables
-  - gsd =2
+  - gsd>=3
   - pytest
   - pytest-xdist
   - pytest-cov
--- a/devtools/conda-envs/test.yaml
+++ b/devtools/conda-envs/test.yaml
@@ -14,7 +14,7 @@ dependencies:
   - networkx
   - pandas
   - pytables
-  - gsd>=3
+  - gsd>=2.8
   - pytest
   - pytest-xdist
   - pytest-cov
--- a/mdtraj/formats/gsd.py
+++ b/mdtraj/formats/gsd.py
@@ -119,7 +119,7 @@ def load_gsd_topology(filename, frame=0):
     used to construct the MDTraj topology.
     """
     import gsd.hoomd
-    with gsd.hoomd.open(filename, 'rb') as gsdfile:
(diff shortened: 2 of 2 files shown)


natsukium (commented):
Thanks for your investigation.
Could you please patch line 122 in the same way?
You can run pytest to see if you fixed it correctly.

Also, I think we need to replace `gsd.hoomd.Snapshot` with `gsd.hoomd.Frame`.
See https://gsd.readthedocs.io/en/stable/changes.html#id5.

mattwthompson:
Tests are passing, which is great, and the code looks fine, but maybe there's some stuff that isn't covered by tests?

Could you run a [local report](https://coverage.readthedocs.io/en/7.3.2/#quick-start) and let us know how much of the HOOMD-related submodules are covered by tests?

natsukium:
> Tests are passing, which is great, and the code looks fine, but maybe there's some stuff that isn't covered by tests?

Because we use `gsd == 2.9.0` on github actions, but this PR is for `gsd >= 3`.
https://github.com/mdtraj/mdtraj/actions/runs/6657489620/job/18149382762?pr=1826#step:5:53

mattwthompson:
Good catch! I totally missed that. @daico007 could you update the test environment to use GSD 3?

Unless there is a compelling need to maintain backwards compatibility with future versions, I think the release containing this change should just work with just GSD 3

daico007:
Gotcha! Thank you y'all for the feedback! I will try to have those changes in tomorrow.

daico007:
Final commits passed all the gsd test locally

mattwthompson (approved):
It looks like the only relevant API change here was the difference in supported file modes?

If so, that would work with GSD v2 so maybe we can just have `gsd =2,3` in deployment

daico007:
The `gsd.hoomd.Frame` was introduced in gsd v2.8.0 (https://gsd.readthedocs.io/en/stable/changes.html#id5), so that's we will can only support that version forward

mattwthompson:
Either is fine - could you merge upstream back in? I reverted the coverage stuff since it wasn't working and I don't have the time to invest into fixing it.

daico007:
Done

PR #1827, opened 2023-10-28
Set up CodeCov

mattwthompson:


natsukium:
Regarding pytest-xdist, I have also tried parallelizing with it, but the test sometimes fails when the number of threads is large.
In the attached logs I use over 60 threads, but I see the same error on my 8 core macbook.
I guess this is a data race and needs to be addressed in order to parallelize.

<details><summary>Error Log</summary>

```
=================================== FAILURES ===================================
___________________________ test_read_atomindices_1 ____________________________
[gw3] linux -- Python 3.11.5 /nix/store/ffll6glz3gwx342z0ch8wx30p5cnqz1z-python3-3.11.5/bin/python3.11
    def test_read_atomindices_1():
>       with TRRTrajectoryFile(temp) as f:
test_trr.py:153: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
>   ???
E   OSError: Malformed TRR file. Number of atoms <= 0. Are you sure this is a valid GROMACS TRR file?
mdtraj/formats/xtc/trr.pyx:224: OSError
___________________________ test_read_atomindices_2 ____________________________
[gw36] linux -- Python 3.11.5 /nix/store/ffll6glz3gwx342z0ch8wx30p5cnqz1z-python3-3.11.5/bin/python3.11
    def test_read_atomindices_2():
>       with TRRTrajectoryFile(temp) as f:
test_trr.py:166: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
>   ???
E   OSError: Malformed TRR file. Number of atoms <= 0. Are you sure this is a valid GROMACS TRR file?
mdtraj/formats/xtc/trr.pyx:224: OSError
```
</details>

[Full Log](https://logs.ofborg.org/?key=nixos/nixpkgs.263986&attempt_id=876b8992-5d0b-46a1-9e36-db56eb3d705d)

mattwthompson:
60 threads on 8 physical cores is a bit excessive, no?

natsukium:
> 60 threads on 8 physical cores is a excessive

Sorry, this happens on two machines, a 32-core 64-thread x86_64-linux machine and an 8-core 8-thread aarch64-darwin machine.

I haven't tried it, but what about using `xdist_group` like this? https://github.com/pytest-dev/pytest-xdist/issues/385#issuecomment-1304877301
I'm trying to add this project to [NixOS/nixpkgs](https://github.com/NixOS/nixpkgs) and would like to run as many tests as possible without modification to make sure it works.

mattwthompson:
For packaging purposes, you can run tests however you wish. Tests aren't a part of the package itself, so you're free to do whatever you want there (including not run any tests).

For comparison, the conda recipe doesn't do anything but make sure it imports without crashing: https://github.com/conda-forge/mdtraj-feedstock/blob/e7c5ca7dff9c8c9dec36254e2df25248de62baf2/recipe/meta.yaml

I'm sure we'd accept patches that improve the testing infrastructure here, though.

natsukium:
Yes, of course, if we can't fix it, I'll just package it without pytest-xdist.
But I guess contributors suffer from this problem when they run tests locally?

> I'm sure we'd accept patches that improve the testing infrastructure here, though.

Fair enough. I will look into it a bit more and submit a PR.
Thank you so much!

PR #1828, opened 2023-10-28
[DNM] Remove some tests

mattwthompson:
Just seeing if coverage reports are working


mattwthompson:
I'll fiddle with this another time, it's not working now

PR #1833, opened 2023-11-16
setup cibuildwheel

### LINKED ISSUE #1794: Set up `cibuildwheel` to distribute wheels
Btw while we are on the topic of platforms, wonder if we should get onboard tools like https://github.com/pypa/cibuildwheel so we publish all bdist wheels and don't have to manage all the different platforms ourselves.

_Originally posted by @chajath in https://github.com/mdtraj/mdtraj/issues/1751#issuecomment-1234521163_

njzjz:
Fix #1794.

Maintainers need to add the GitHub Actions file as the trusted publisher on PyPI following https://docs.pypi.org/trusted-publishers/adding-a-publisher/. After that, the package can be published automatically.


njzjz:
https://github.com/njzjz/mdtraj/actions/runs/6896938474/

Python 3.12 failed with distutils not found:

```
    File "<string>", line 1, in <module>
    File "/tmp/tmp.95oGBkvyJH/venv/lib/python3.12/site-packages/mdtraj/__init__.py", line 30, in <module>
      from .formats.registry import FormatRegistry
    File "/tmp/tmp.95oGBkvyJH/venv/lib/python3.12/site-packages/mdtraj/formats/__init__.py", line 9, in <module>
      from .netcdf import NetCDFTrajectoryFile
    File "/tmp/tmp.95oGBkvyJH/venv/lib/python3.12/site-packages/mdtraj/formats/netcdf.py", line 40, in <module>
      from distutils.version import StrictVersion
  ModuleNotFoundError: No module named 'distutils'
```

Windows failed with zlib not found:
```
    mdtraj/formats/tng/src/lib/tng_io.c(32): fatal error C1083: Cannot open include file: 'zlib.h': No such file or directory
```

Other builds were successful.

peastman:
See https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html for migrating from distutils to setuptools.  In most cases it just means changing a few imports.

njzjz:
> See https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html for migrating from distutils to setuptools. In most cases it just means changing a few imports.

I submitted #1834 to remove distutils.

.github/workflows/wheel.yaml
```diff
@@ -0,0 +1,106 @@
...
+    name: Build source distribution
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+      - uses: actions/setup-python@v4
+        name: Install Python
+        with:
+          python-version: '3.12'
```
mattwthompson, line 84: This installs Python 3.12 in all builds - do you mean to use `matrix.python`?

njzjz: The `build_sdist` step only packages the source distribution, i.e. the file that ends with `.tar.gz`. The Python version here doesn't matter as all versions and platforms share the same source distribution.

njzjz:
Now there is another problem related to Python 3.12:

```
    File "<string>", line 1, in <module>
    File "/tmp/tmp.NmSnxjZdvT/venv/lib/python3.12/site-packages/mdtraj/__init__.py", line 57, in <module>
      from .geometry import *
    File "/tmp/tmp.NmSnxjZdvT/venv/lib/python3.12/site-packages/mdtraj/geometry/__init__.py", line 53, in <module>
      from .order import *
    File "/tmp/tmp.NmSnxjZdvT/venv/lib/python3.12/site-packages/mdtraj/geometry/order.py", line 25, in <module>
      from pkg_resources import parse_version
  ModuleNotFoundError: No module named 'pkg_resources'
```

`pkg_resources` is also removed from Python 3.12; see https://github.com/python/cpython/issues/95299.

mattwthompson:
`pkg_resources` is not removed, it's just "deprecated" as an API no longer bundled with CPython. It should be provided by `pip` and/or `setuptools`, so I'm not sure why it's not found there.

sukritsingh:
Based on the issue above and the [changelog](https://docs.python.org/3/whatsnew/3.12.html), it looks like `setuptools` was entirely removed as part of the default python 3.12. They recommend running `pip install setuptools` for any `venv` created: 
```
to access these run pip install setuptools in the [activated](https://docs.python.org/3/library/venv.html#venv-explanation) virtual environment.
```

Perhaps we never listed it as a "prerequisite" since it used to be bundled? The solution would just be to include setuptools as part of the requirements right? Do we want that?

mattwthompson:
I expected it to be present in environments used for building since it's listed [here](https://github.com/mdtraj/mdtraj/blob/95d79747deef42c976ca362a57806b61933409f3/pyproject.toml#L3). Notably, the build succeeds but the action fails in a "Testing wheels ..." section, so maybe something like [`CIBW_TEST_REQUIRES`](https://cibuildwheel.readthedocs.io/en/stable/options/#test-requires) is needed? That's my best guess at how dependencies are managed there. I don't use this generation of PyPI tools in my day job so I could be wrong.

It should be list as a runtime dependency, it's going to be present in any environment anyway (conda's `pip` recipe, for example, pulls it in, and it's gotta be present in other build systems given that it's [vendored](https://github.com/pypa/pip/tree/23.3.1/src/pip/_vendor/pkg_resources) in the source code.)

sukritsingh:
Oh weird (and never mind then!) - Seems like then this as issue on testing setup side rather than the codebase side? I'm not familiar with PyPI tools at that level either to have any further insights on possible fixes (but will try and think some about this)

njzjz:
> I expected it to be present in environments used for building since it's listed [here](https://github.com/mdtraj/mdtraj/blob/95d79747deef42c976ca362a57806b61933409f3/pyproject.toml#L3).

The building environment is usually isolated. This is recommended by [PEP-517](https://peps.python.org/pep-0517/#recommendations-for-build-frontends-non-normative).

mattwthompson:
I had a quick look at this. It's only needed at runtime for [this](https://github.com/mdtraj/mdtraj/blob/95d79747deef42c976ca362a57806b61933409f3/mdtraj/geometry/order.py#L100) check, which is obsolete now. I'll write up a quick patch for it.

mattwthompson:
@njzjz when you're able, merge `master` in and 3.12 might build

.github/workflows/wheel.yaml
```diff
@@ -0,0 +1,106 @@
...
+      fail-fast: false
+      matrix:
+        include:
+          # linux-64
+          - os: ubuntu-latest
+            python: 38
+            platform_id: manylinux_x86_64
+          - os: ubuntu-latest
+            python: 39
+            platform_id: manylinux_x86_64
```
mattwthompson, line 21: It'd be lovely if this section was condensed a little bit, since most of it seems to be fairly repetitive.

I assume that this is set up follow some standard naming of wheels? Something like
```
      matrix:
        include:
          - os: [ubuntu-latest, macos-latest]  # Windows missing zlib
          - python: ["3.8", "3.9", "3.10", "3.11", "3.12"]
```

with aliasing to what's currently `platform_id` etc. (not sure how to do this in YAML - I assume it's possible) would be much more compact and easier to maintain.

.github/workflows/wheel.yaml
```diff
@@ -0,0 +1,106 @@
+name: Build and upload to PyPI
+
+on:
+  push:
+  pull_request:
```
mattwthompson, line 5: Maybe worth adding a `workflow_dispatch` so we can trigger this on-demand

mattwthompson (approved):

mattwthompson:
@njzjz thanks for your contribution, this was instructive for me to go through as somebody who is behind the time on wheels

njzjz:
I resolve the comments. I also switch the macOS wheels to [`universal2`](https://cibuildwheel.readthedocs.io/en/stable/faq/#universal2), which can be installed on both `x86_64` and `arm64` systems.

mattwthompson:
@njzjz anything changed in the past ~year with this tool?

njzjz:
> @njzjz anything changed in the past ~year with this tool?

I notice mdtraj has required Python 3.9, so 3.8 should be removed from the matrix.

njzjz:
I'll also bump cibuildwheel to the latest version.

mattwthompson:
Please also drop 3.9

mattwthompson:
It looks like the macOS wheels are failing. I unfortunately don't know what's going on there, without looking into it more deeply

njzjz:
> It looks like the macOS wheels are failing. I unfortunately don't know what's going on there, without looking into it more deeply

The upstream issue is tracked in https://github.com/matthew-brett/delocate/issues/227. I downgrade the delocate version to fix it.

janash:
Merged in #1980

PR #1834, opened 2023-11-17
Drop distutils usage and add Python 3.12 tests

njzjz:
The `distutils` module has been removed from Python 3.12 per [PEP 632](https://peps.python.org/pep-0632/).

### DIFF
--- a/mdtraj/formats/amberrst.py
+++ b/mdtraj/formats/amberrst.py
@@ -30,12 +30,13 @@ removing the functionality that is not needed.
 """
 
 from __future__ import print_function, division
-from distutils.version import StrictVersion
 from math import ceil
 import os
 import warnings
 
 import numpy as np
+from packaging.version import Version
+
 from mdtraj import version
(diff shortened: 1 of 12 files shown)


mdtraj/formats/amberrst.py
```diff
@@ -462,7 +463,7 @@ class AmberNetCDFRestartFile(object):
     def __init__(self, filename, mode='r', force_overwrite=False):
         self._closed = True
         self._mode = mode
-        if StrictVersion(import_('scipy.version').short_version) < StrictVersion('0.12.0'):
+        if Version(import_('scipy.version').short_version) < Version('0.12.0'):
```
mattwthompson, line 466: (don't need to fix here)

This check is hilariously old, can safely be dropped now: 
image

mattwthompson (approved):
LGTM

sukritsingh:
LGTM! Will merge within 24 hours if no changes are made so that #1833 can be addressed

mattwthompson (approved):

PR #1837, opened 2023-11-20
Skip NumPy 1.8 version check

mattwthompson:
See #1833 for context

### DIFF
--- a/mdtraj/geometry/order.py
+++ b/mdtraj/geometry/order.py
@@ -22,10 +22,7 @@
 
 from __future__ import print_function, division
 
-from pkg_resources import parse_version
-
 import numpy as np
-NP18 = parse_version(np.__version__) >= parse_version('1.8.0')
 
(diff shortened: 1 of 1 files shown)


mdtraj/geometry/order.py
```diff
@@ -247,15 +239,13 @@ def _compute_Q_tensor(all_directors):
...
+
+    normed = all_directors / np.linalg.norm(all_directors, axis=2)[..., np.newaxis]
 
     Q_ab = np.zeros(shape=(all_directors.shape[0], 3, 3), dtype=np.float64)
-    for n, directors in enumerate(all_directors):
-        if NP18:
-            normed_vectors = normed[n]
-        else:
-            normed_vectors = directors / np.sqrt((directors ** 2.0).sum(-1))[..., np.newaxis]
+    for n in range(len(all_directors)):
```
ctk3b, line 246: Could probably iterate over `normed` instead

mattwthompson: Yep!

Issue #1835, opened 2023-11-17
Drop old hard-coded version checks 


mattwthompson:
There are some hard-coded version checks to some ~deacde old versions, like this one that stands out in #1834:

This check is hilariously old, can safely be dropped now: 
<img width="451" alt="image" src="https://github.com/mdtraj/mdtraj/assets/7935382/36353a6f-5656-48c8-a104-5892ad9821c1">

There was also a check somewhere for a 0.x version of Pandas, which is no longer needed as it's been on 2.x for a while.

_Originally posted by @mattwthompson in https://github.com/mdtraj/mdtraj/pull/1834#discussion_r1397842794_

PR #1838, opened 2023-11-24
Drop support for ancient upstreams

mattwthompson:
Drop support for
* Python 2
* PyTables 2
* SciPy <0.12
* Pandas <0.14

Also drops the internal `mdtraj.utils.six` module, which I think was just vendored. That's technically an API break, but I don't think there's good reasons for downstream packages to depend on it (both because Python 2/3 compatibility is bad and because this isn't the way to get `six`).


mdtraj/scripts/mdconvert.py
```diff
@@ -310,7 +308,7 @@ def main(args, verbose=True):
                             elif isinstance(v, md.Topology):
                                 _data[k] = v
                             else:
-                                raise RuntineError()
```
ctk3b, line 313: 😬

tests/test_docstrings.py
```diff
@@ -33,6 +32,7 @@ def module(request):
     return mod
 
 
+@pytest.mark.skip()
```
ctk3b, line 35: Is this test broken?

tests/test_nmr.py
```diff
@@ -35,8 +33,8 @@ def test_ppm(get_fn):
     #       2     ASN   CA    999.000  53.004  51.168  51.802  53.081  54.098  52.820  52.379  51.856  53.034  52.754  54.134  54.222  51.210  52.207  50.824  54.459  53.605  54.211  53.688  52.344  53.004  51.168  51.802  53.081  54.098  52.820  52.379  51.856  53.034  52.754  54.134  54.222  51.210  52.207  50.824  54.459  53.605  54.211  53.688  52.344
 
 
-@pytest.mark.skipif((not find_executable(SHIFTX2)) or (not six.PY2) or (os.environ.get('TRAVIS', '') == 'true'),
-                    reason='SHIFTX2 binary not found or Python 2 not found or runningn on travis.')
+@pytest.mark.skipif((not find_executable(SHIFTX2)) or (os.environ.get('TRAVIS', '') == 'true'),
+                    reason='SHIFTX2 binary not found or runningn on travis.')
```
ctk3b, line 37:
```suggestion
                    reason='SHIFTX2 binary not found or running on travis.')
```

ctk3b (approved):
LGTM % fixing or updating docs on the `test_docstrings` test

mattwthompson:
There are more Python 2 artifacts that could be re-jiggered, but for now this does everything I originally set out and more. Will leave open briefly in case anybody else really likes reviewing semi-automated changelogs

sukritsingh:
Is #1835 a relevant issue to this PR? I linked it here just for tracking purposes but can unlink if it's not relevant/addressed (I haven't checked out the commits yet tbh)

mattwthompson:
Yep, this should resolve that issue

a-detiste:
https://wiki.debian.org/Python3-six-removal

> Also drops the internal mdtraj.utils.six module, which I think was just vendored. That's technically an API break, but I don't think there's good reasons for downstream packages to depend on it (both because Python 2/3 compatibility is bad and because this isn't the way to get six).

You're so right. So many project does that and now feels like they should do it forever

PR #1839, opened 2023-11-24
Run tests in parallel

mattwthompson:
Free runners include two cores these days and tests take a few minutes, so this should speed things up.


mattwthompson:
This drops tests from ~5 to ~3 minutes

jeremyleung521:
Sorry for just casually strolling in, but there should be a space between `-n` and `auto` I think?

(Otherwise, nice plugin. Might try to implement pytest-xdist on our projects)

PR #1844, opened 2023-12-27
Add atom selection via atom_indices to SASA computation (try #2, three years after)

gph82:
A little holiday coding for something I had working locally for 3 years and never pushed (#[PR1671](https://github.com/mdtraj/mdtraj/pull/1671) was closed as stale/wontfix). It's been crucial for all my projects all this time.

Here's the original PR message:
> Limit the SASA computation to some atoms of interest via atom_indices when calling md.shrake_rupley
> 
> The excluded atoms are still considered as potential surface accessibility blockers, but their own SASA doesn't get computed and gets set to zero instead.
> 
> This is particularly helpful for membrane-embedded proteins where you want the lipids "there" but you don't really care about their SASA.
>  

I followed @rmcgibbo 's suggestion for the c++ implementation. 

The return SASA-value for 'excluded' atoms/residues is set to np.nan in 9b0d5b81021e0eabb42958462281073d5caf77d0. I can revert that if you prefer a 0 value.

### DIFF
--- a/mdtraj/geometry/sasa.py
+++ b/mdtraj/geometry/sasa.py
@@ -216,7 +216,7 @@ def shrake_rupley(
         for what atoms the SASA is computed. E.g. you can pass a lipid-embedded
         protein (s.t. the lipids are considered blockers) but only compute
         SASA for the protein using atom_selection = traj.top.select("protein").
-        The excluded atoms/residues get a SASA value 0
+        The excluded atoms/residues get a SASA value -1.
 
     Returns
     -------
@@ -287,9 +287,11 @@ def shrake_rupley(
 
     if atom_indices is None:
         atom_selection_mask = np.ones(traj.n_atoms, dtype=np.int32)
+        out = np.zeros((xyz.shape[0], dim1), dtype=np.float32)
     else:
-
         atom_selection_mask = np.array([[1 if ii in atom_indices else 0][0] for ii in range(traj.n_atoms)], dtype=np.int32)
+        out = np.full((xyz.shape[0], dim1), -1, dtype=np.float32)
+        out[:,atom_mapping[atom_indices]]=0
 
     modified_radii = {}
     if change_radii is not None:
@@ -299,7 +301,6 @@ def shrake_rupley(
         for k, v in change_radii.items():
             modified_radii[k] = v
 
-    out = np.zeros((xyz.shape[0], dim1), dtype=np.float32)
     if bool(modified_radii):
         atom_radii = [modified_radii[atom.element.symbol] for atom in traj.topology.atoms]
     else:
--- a/mdtraj/geometry/sasa.py
+++ b/mdtraj/geometry/sasa.py
@@ -183,6 +183,7 @@ def shrake_rupley(
     mode="atom",
     change_radii=None,
     get_mapping=False,
+    atom_indices=None
 ):
     """Compute the solvent accessible surface area of each atom or residue in each simulation frame.
 
@@ -207,6 +208,15 @@ def shrake_rupley(
         Instead of returning only the areas, also return the indices of the
         atoms or the residue-to-atom mapping. If True, will return a tuple
         that contains the areas and the mapping (np.array, shape=(n_atoms)).
+    atom_indices : iterable, optional
+        Selection of atoms indices for which the SASA will be computed.
+        Default is all atoms, but a sub-selection of atoms can be
+        passed here. This selection doesn't affect what
+        atoms are considered accessibility blockers, it only affects
+        for what atoms the SASA is computed. E.g. you can pass a lipid-embedded
+        protein (s.t. the lipids are considered blockers) but only compute
+        SASA for the protein using atom_selection = traj.top.select("protein").
+        The excluded atoms/residues get a SASA value 0
 
     Returns
     -------
@@ -272,9 +282,14 @@ def shrake_rupley(
                 "residues must have contiguous integer indices " "starting from zero",
             )
     else:
-        raise ValueError(
-            'mode must be one of "residue", "atom". "%s" supplied' % mode,
-        )
+        raise ValueError('mode must be one of "residue", "atom". "%s" supplied' %
+                         mode)
+
+    if atom_indices is None:
+        atom_selection_mask = np.ones(traj.n_atoms, dtype=np.int32)
+    else:
+
+        atom_selection_mask = np.array([[1 if ii in atom_indices else 0][0] for ii in range(traj.n_atoms)], dtype=np.int32)
 
     modified_radii = {}
     if change_radii is not None:
@@ -291,7 +306,7 @@ def shrake_rupley(
         atom_radii = [_ATOMIC_RADII[atom.element.symbol] for atom in traj.topology.atoms]
     radii = np.array(atom_radii, np.float32) + probe_radius
 
-    _geometry._sasa(xyz, radii, int(n_sphere_points), atom_mapping, out)
+    _geometry._sasa(xyz, radii, int(n_sphere_points), atom_mapping, atom_selection_mask, out)
 
     if get_mapping is True:
         return out, atom_mapping
--- a/tests/test_sasa.py
+++ b/tests/test_sasa.py
@@ -184,8 +184,8 @@ def test_sasa_6(get_fn):
     # The computed SASA values are the same as in "normal" mode
     np.testing.assert_array_almost_equal(sasa_all_atoms[:,atoms_resid1],
                                          sasa_all_atoms_w_selection[:,atoms_resid1])
-    # The uncomputed SASA values are all 0
-    np.testing.assert_equal(np.unique(sasa_all_atoms_w_selection[:,atoms_resid0_2]),0)
+    # The uncomputed SASA values are all -1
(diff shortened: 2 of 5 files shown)


mdtraj/geometry/sasa.py
```diff
@@ -184,7 +203,10 @@ def shrake_rupley(traj, probe_radius=0.14, n_sphere_points=960, mode='atom', cha
         atom_radii = [_ATOMIC_RADII[atom.element.symbol] for atom in traj.topology.atoms]
     radii = np.array(atom_radii, np.float32) + probe_radius
 
-    _geometry._sasa(xyz, radii, int(n_sphere_points), atom_mapping, out)
+    _geometry._sasa(xyz, radii, int(n_sphere_points), atom_mapping, atom_indices, out)
+
+    # Cast np.nan for residues/atoms not covered by atom_indices
+    out[:, ~np.isin(np.unique(atom_mapping[np.arange(traj.n_atoms)]), np.unique(atom_mapping[atom_indices]))] = np.nan
```
mattwthompson, line 209: The tilda here is just to flip booleans?

tests/test_sasa.py
```diff
@@ -149,3 +148,33 @@ def test_sasa_5():
...
+
+    # The computed SASA values are the same as in "normal" mode
+    np.testing.assert_array_almost_equal(sasa_all_atoms[:,atoms_resid1],
+                                         sasa_all_atoms_w_selection[:,atoms_resid1])
+    # The uncomputed SASA values are all np.nan
+    np.testing.assert_equal(np.unique(sasa_all_atoms_w_selection[:,atoms_resid0_2]),np.nan)
+
+    # Mode="residues"
+    sasa_all_residues = md.geometry.shrake_rupley(t, mode="residue")
+    sasa_all_residues_w_selection = md.geometry.shrake_rupley(t, atom_indices=atoms_resid1, mode="residue")
```
mattwthompson, line 170: Could we also make sure that the SASA values match the expected result when the other residue selection is made? It's nice that you picked a residue other than the first/zeroth for this test, and checking it for the multiple and non-sequential residue selection would be nice as well

mdtraj/geometry/sasa.py
gph82: Yep. If you prefer I can use np.invert. In any case, from the [docs](https://numpy.org/doc/stable/reference/generated/numpy.invert.html)

> The ~ operator can be used as a shorthand for np.invert on ndarrays.

tests/test_sasa.py
gph82: I've updated the tests accordingly. I have also added a test for selecting a subset of atoms (not all) of a residue, checking that e.g. if you select only one atom, mode="atom" or mode="residue" yield the same value (not that I was expecting otherwise but just to try an edge-case)

gph82:
Thanks for reviewing this. One thing that still has me doubting is what's the best value to return for the "other" atoms/residues for which SASA wasn't computed. Currently it is "0", which might be misleading. Alternatively np.nan could work  but might break something elsewhere (you can't do sasa.mean() etc)

gph82:
Hey @mattwthompson  I see you're gearing up for a release. I just saw your comment and the conflicts. Do I have a couple of hrs to solve it before you cut the release? 

I got a paper coming out using this functionality (out of my local mdtraj fork) and it'd be nice if the official mdtraj package already had this functionality released

gph82:
The only other function that faces an equivalent decision is [md.rmsf](https://github.com/mdtraj/mdtraj/blob/4c9bb6e8bc7d6e86890ff0d57814c3c76f7cf792/mdtraj/rmsd/_rmsd.pyx#L225) with:
```
    atom_indices : array_like, or None
        The indices of the atoms to use in the RMSF calculation. If not
        supplied, all atoms will be used(...)
```
Which has an influence over the [returned value's shape](https://github.com/mdtraj/mdtraj/blob/4c9bb6e8bc7d6e86890ff0d57814c3c76f7cf792/mdtraj/rmsd/_rmsd.pyx#L273)
```
    Returns
    -------
    rmsf : np.ndarray, shape=(atom_indices,)
        A 1-D numpy array of the optimal root-mean-square fluctuations of the
       selected atoms.
```

gph82:
But changing the returned SASA-array's shape seemed more invasive. Maybe @rmcgibbo can tell us what he thinks?

sukritsingh:
>  One thing that still has me doubting is what's the best value to return for the "other" atoms/residues for which SASA wasn't computed. Currently it is "0", which might be misleading. Alternatively np.nan could work but might break something elsewhere (you can't do sasa.mean() etc)

Thinking about this issue from a "If I was analyzing data like this, what would I want to see" perspective...**Can I suggest **, it returns "-1"**** instead of returning "0" (which I agree is misleading)? -1 makes _zero_ sense, but is less invasive as an output. It would be easy to filter out and if something broke it'd be obvious to a user.

gph82:
Good compromise @sukritsingh (and thanks for looking into this) . I'll do that and if let you guys decide

gph82:
Can't really understand what happened with macOS-13,3.10 in the CI-tests.
```
FAILED tests/test_trr.py::test_read_atomindices_1 - OSError: Malformed TRR file. Number of atoms <= 0. Are you sure this is a valid GROMACS TRR file?
FAILED tests/test_trr.py::test_read_atomindices_2 - OSError: Malformed TRR file. Number of atoms <= 0. Are you sure this is a valid GROMACS TRR file?
```

gph82:
Anyway, just to have it documented somewhere, @sukritsingh , here's an example for SASA of GDP in the Gs-binding pocket.
```
In [11]: %time SASA_GDP = md.shrake_rupley(geom[:100], mode="residue")[:,GDP_residx]
CPU times: user 4min 4s, sys: 307 ms, total: 4min 4s
Wall time: 32 s

In [12]: %time SASA_GDP = md.shrake_rupley(geom[:100], mode="residue", atom_indices=GDP_atoms)[:,GDP_residx]
CPU times: user 1.03 s, sys: 3.96 ms, total: 1.04 s
Wall time: 262 ms
```
I think it's worth it!

sukritsingh:
That's such a huge improvement! Is this with the "returns -1" fix (ie it returns the same shaped output but just has -1 wherever it wasn't used?). Just want to make sure I understand if the output shape was altered in this speedup. 

I'll dig into the test failures a bit - not clear why that one OS/python version pair would fail when it's the same file (Maybe related to `md_convert`?)

gph82:
Hey, so 
- shape of returned array is always 
  - (n_frames, n_atoms) for mode="atom"
  - (n_frames, n_residues) for mode="residue"
  regardless of having "-1" or "0" for the non-computed SASAs
- speedup comes from not computing SASA for all the Gprotein atoms, while still allowing them to be "accessibility blockers" for the GDP. The fact that it's "-1" or "0" is not important here. It's the fact you excluded them from the SASA computation

sukritsingh:
Huzzah! I suspected that the speedup was coming from the "only computing from limited" atoms part, but just wanted to make sure the shape was being preserved. Great! 
1. I think keeping the shape and having it return -1 (and updating the docs accordingly) will be helpful. We should preserve default behaviors to doing all-atoms/all-residues. 
2. Do we need any new tests for this argument? Just thinking out loud
3. Let's see if we can figure out what the test failure is coming from (seems random and unrelated...might be something in the runner itself)

gph82:
Test's have been added [tests](https://github.com/mdtraj/mdtraj/pull/1844/files#diff-df7a127d5d05c8c85ba767c18a12017dde153e01eb317c97d0ea8cbf23779bb0R174) and docs have been [updated](https://github.com/mdtraj/mdtraj/pull/1844/files#diff-5a1784ea36dbe910493f987abd625219b2a93cfcda4159d2021c4f3546f35606R211).

WRT to CI-tests, can you re-trigger that test alone?

sukritsingh:
> Test's have been added tests and docs have been updated.

Oh derp right I wrote that forgetting I'd seen those. 

Rerunning now with debug logging just in case it crashes again!

gph82:
FWIW same errors on ubuntu-latest,3.10 [here](https://github.com/gph82/mdtraj/actions/runs/9188047181/job/25267001889)

sukritsingh:
Weird - it looks like the `ubuntu-latest, 3.10` passed for the PR [here](https://github.com/mdtraj/mdtraj/actions/runs/9131706368/job/25190391698?pr=1844)?

sukritsingh:
Ok now that the test passed, this looks good to me to merge! @mattwthompson did you have any concerns you want addressed still with this PR? Just want to make sure the "returns `-1` for unselected atoms" functionality was okay with you as well.

sukritsingh:
> I don't like introducing numbers that don't represent the outputs but I don't know of a much better way to do this

Yeah this was honestly the best idea I could come up with. My only thoughts for simplicity were 
1. Avoid "0" because that can be a meaningful output (ie the atom is buried) so we don't want to cause confusion between buried atoms and unmeasured atoms
2. Maintain the shape of the output for consistency

I'm definitely open to other ideas though if someone comes up with one

mattwthompson:
Yep, maintaining the shape is good and inserting zeros would have been worse

gph82:
Thanks for merging. I'm happy this made it into the release. I think its a good addition! I'll post when the paper with the usecase is out

PR #1848, opened 2024-01-14
Use repeated bonds in PDB CONECT records to increase bond orders

### LINKED ISSUE #1847: Bond order not preserved for PDB files
E.g. take benzene here with appropriate CONECT records and load into a viewer and you'll see it display as expected

```
HETATM    1  C1  UNL     1       0.794  -1.149   0.015  1.00  0.00           C  
HETATM    2  C2  UNL     1       1.361   0.130  -0.002  1.00  0.00           C  
HETATM    3  C3  UNL     1       0.590   1.268  -0.017  1.00  0.00           C  
HETATM    4  C4  UNL     1      -0.787   1.145  -0.015  1.00  0.00           C  
HETATM    5  C5  UNL     1      -1.341  -0.112   0.002  1.00  0.00           C  
HETATM    6  C6  UNL     1      -0.581  -1.266   0.017  1.00  0.00           C  
HETATM    7  H1  UNL     1       1.403  -2.043   0.027  1.00  0.00           H  
HETATM    8  H2  UNL     1       2.418   0.211  -0.004  1.00  0.00           H  
HETATM    9  H3  UNL     1       1.007   2.269  -0.030  1.00  0.00           H  
HETATM   10  H4  UNL     1      -1.418   2.029  -0.026  1.00  0.00           H  
HETATM   11  H5  UNL     1      -2.422  -0.217   0.004  1.00  0.00           H  
HETATM   12  H6  UNL     1      -1.024  -2.265   0.030  1.00  0.00           H  
CONECT    1    2    2    6    7
CONECT    2    1    1    3    8
CONECT    3    2    4    4    9
CONECT    4    3    3    5   10
CONECT    5    4    6    6   11
CONECT    6    1    5    5   12
CONECT    7    1
CONECT    8    2
CONECT    9    3
CONECT   10    4
CONECT   11    5
CONECT   12    6
END
```

But if you load and save it again with mdtraj, bond orders are gone:
```
import mdtraj as md
lig = md.load("benzene.pdb")
lig.save("benzene2.pdb")
```

nffaruk:
Closes #1847 . While repeated bonds in CONECT records are not explicitly mentioned in the [spec](https://www.wwpdb.org/documentation/file-format-content/format33/sect10.html#CONECT), they are used for setting bond orders in various widely used packages, especially those that deal with small molecules/cheminformatics (e.g. PyMOL, Schrodinger CADD tools, Open Babel, RDKit). Adding this feature will make it easy to import/export trajectory snapshots that involve small molecules complexed with proteins for processing with these tools.

### DIFF
--- a/mdtraj/formats/pdb/pdbfile.py
+++ b/mdtraj/formats/pdb/pdbfile.py
@@ -51,7 +51,7 @@ import numpy as np
 import xml.etree.ElementTree as etree
 from copy import copy, deepcopy
 from mdtraj.formats.pdb.pdbstructure import PdbStructure
-from mdtraj.core.topology import Topology
+from mdtraj.core.topology import Topology, Bond
 from mdtraj.utils import ilen, cast_indices, in_units_of, open_maybe_zipped
 from mdtraj.formats.registry import FormatRegistry
 from mdtraj.core import element as elem
@@ -418,9 +418,15 @@ class PDBTrajectoryFile(object):
                             'A', 'G', 'C', 'U', 'I', 'DA', 'DG', 'DC', 'DT', 'DI', 'HOH']
         conectBonds = []
         if self._last_topology is not None:
-            for atom1, atom2 in self._last_topology.bonds:
+            for bond in self._last_topology.bonds:
+                atom1, atom2 = bond[0], bond[1]
+                # Add bond repeats for hetatom bonds when bond order > 1
                 if atom1.residue.name not in standardResidues or atom2.residue.name not in standardResidues:
-                    conectBonds.append((atom1, atom2))
+                    if bond.order is not None:
+                        for bo in range(bond.order):
+                            conectBonds.append((atom1, atom2))
+                    else:
+                        conectBonds.append((atom1, atom2))
                 elif atom1.name == 'SG' and atom2.name == 'SG' and atom1.residue.name == 'CYS' and atom2.residue.name == 'CYS':
                     conectBonds.append((atom1, atom2))
         if len(conectBonds) > 0:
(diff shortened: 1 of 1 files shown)


mattwthompson:
I'm super conflicted on whether or not this is a positive change. I don't like going off-spec, but I recognize that other libraries are already doing this and that it has some practical benefits. If we do more forward with this, I suspect the change will be sweeping and we'll need to take a look at a bunch of other behaviors, examples, etc.

I don't actually use PDB files much and I'm inclined to defer to people (like Peter in the linked issue thread) who are more intimately familiar with their structure and usage.

peastman:
The reading side is probably reasonable.  If you encounter a repeated bond that's already outside the spec.  You should give some careful thought to how you represent this, particularly the distinction between "single bond" and "bond order not know".  How do you decide which one it is?

I'm a bit less comfortable with the writing side.  The repeated bonds could lead to errors when reading it with other programs.  What about making it a configurable option, with the default being not to use repeated bonds?

nffaruk:
I made the writing part configurable as Peter suggested.

> You should give some careful thought to how you represent this, particularly the distinction between "single bond" and "bond order not know". How do you decide which one it is?

Unfortunately, it doesn't seem like the existing treatment in mdtraj is rigorous about it. For example, the standard bonds made through `create_standard_bonds` in topology.py don't have any bond order assigned (e.g. `None` instead of 1). Also bond counting and `n_bonds` property for atoms don't consider bond order. So I'm inclined to just treat `None` and 1 bond order equivalently for now.

nffaruk:
Checking in again @mattwthompson and @peastman

mattwthompson:
The code changes look good to me on its face - I'm confused why CI didn't run?

It would be useful to get more feedback from biophysicists on this; Peter's comments lead me to believe this is change is a net positive but that's without hacking away on how other tools parse these files

nffaruk:
So this didn't make it into 1.10, I can update it to work with the new changes when I have time, but it would be good to know if there were any remaining issues that prevented it from inclusion.

sukritsingh:
Sorry about that - still cleaning up other issues/PRs as well. I think the linter is catching some minor conflicts that need to be resolved - then we can run the CI and make sure everything works?

Is there a test included to check that it's handling repeated bonds as expected?

Issue #1853, opened 2024-01-31
chain_id is not preserved when extracting subset of topology


nikolas-fechner:
It seems that the `chain_id` attributes of the chains in a topology object are not preserved when creating it from an atom subset. E.g.
```import mdtraj
print(mdtraj.version.full_version)
protein_traj = mdtraj.load_pdb(protein_f)
print(protein_traj.topology._chains[0].chain_id)
protein_atomsel = protein_traj.topology.select("protein")
protein_traj = protein_traj.atom_slice(protein_atomsel)
print(protein_traj.topology._chains[0].chain_id)
```
prints
```
1.9.9
F
None
```

I suspect this is due to `chain_id=chain.chain_id` is not passed to the `addChain` call [here](https://github.com/mdtraj/mdtraj/blob/36b4cee61d936d9ecee875fbdde37120b59a4d51/mdtraj/core/topology.py#L99). The `chain_id` support with the additional parameter for the `addChain` method was introduced in this [PR](https://github.com/mdtraj/mdtraj/pull/1715/files#diff-e70d2485927f29b1848b9cf3cf3cb74ef3455748641c3dc1cef16901c2f241b3).

Thank you for providing and maintaining this library!  I hope this information is useful. 

Kind regards,
Niko

RobertArbon:
This looks like an easy fix: 

```
def _topology_from_subset(topology, atom_indices):

    ...

    for chain in topology._chains:
        newChain = newTopology.add_chain(chain_id=chain.chain_id)

```

I'm happy to submit a PR and a new test for this.

mattwthompson:
Does the spec offer any guidance? It doesn't look to me like it has any particular opinions on how chain IDs are preserved (or not) when slicing up a PDB file. The only thing that caught my eye was this bit:

> The TER record has the same residue name, chain identifier, sequence number and insertion code as the terminal residue.

https://www.wwpdb.org/documentation/file-format-content/format33/sect9.html

peastman:
Chain IDs are meaningful identifiers.  I'd agree they should be preserved.

RobertArbon:
> Does the spec offer any guidance? It doesn't look to me like it has any particular opinions on how chain IDs are preserved (or not) when slicing up a PDB file. The only thing that caught my eye was this bit:
> 
> > The TER record has the same residue name, chain identifier, sequence number and insertion code as the terminal residue.
> 
> https://www.wwpdb.org/documentation/file-format-content/format33/sect9.html

[...]

PR #1858, opened 2024-03-10
preseving chain_id on creating subset.

RobertArbon:
Hi folks, 

This addresses #1853.  When a new topology is created from a subset of atoms, the `chain_id` from the previous chain is carried across.  

Added three new tests in `test_topology.py`
1. `test_preserve_chain_id`
2. `test_preserve_resseq`
3. `test_preserve_serial`

The last two are for completeness - they are already working but not tested.

### DIFF
--- a/tests/test_topology.py
+++ b/tests/test_topology.py
@@ -327,3 +327,19 @@ def test_topology_join_keep_resSeq(get_fn):
 
     eq(out_resSeq_keepId_True, expected_resSeq_keepId_True)
     eq(out_resSeq_keepId_False, expected_resSeq_keepId_False)
+
+def test_preserve_chain_id(get_fn): 
+    # This pdb has three chains, A, B and C.  
+    traj = md.load(get_fn("1vii_sustiva_water.pdb"))
+    sub_traj = traj.atom_slice(traj.top.select('chainid 1'))
+    eq(sub_traj.top._chains[0].chain_id, 'B')
+
+def test_preserve_resseq(get_fn): 
+    traj = md.load(get_fn("1vii_sustiva_water.pdb"))
+    sub_traj = traj.atom_slice(traj.top.select('chainid 0 and resSeq 2'))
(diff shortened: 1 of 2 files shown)


tests/test_topology.py
```diff
@@ -327,3 +327,19 @@ def test_topology_join_keep_resSeq(get_fn):
 
     eq(out_resSeq_keepId_True, expected_resSeq_keepId_True)
     eq(out_resSeq_keepId_False, expected_resSeq_keepId_False)
+
+def test_preserve_chain_id(get_fn): 
+    # This pdb has three chains, A, B and C.  
+    traj = md.load(get_fn("1vii_sustiva_water.pdb"))
+    sub_traj = traj.atom_slice(traj.top.select('chainid 1'))
+    eq(sub_traj.top._chains[0].chain_id, 'B')
```
mattwthompson, line 335: This seems weird but is probably correct, could you verify my understanding?

* 'chainid 1' grabs the 1-th (second) chain
* this chain (1 residue long) is the ligand: https://github.com/mdtraj/mdtraj/blob/36b4cee61d936d9ecee875fbdde37120b59a4d51/tests/data/1vii_sustiva_water.pdb#L601-L631
* the "chain ID" of this chain happens to be `B`

I think I'm just getting confused by the difference between a chain _index_ and a chain _ID_ ...


```python
In [1]: import openmm.app

In [2]: {
   ...:     (chain.index, chain.id)
   ...:     for chain in [
   ...:         residue.chain
   ...:         for residue in openmm.app.PDBFile("1vii_sustiva_water.pdb").topology.residues()
   ...:     ]
   ...: }
Out[2]: {(0, 'A'), (1, 'B'), (2, 'C')}

mattwthompson: Also, what's the TER record if written from this subset using this patch?

RobertArbon: Sorry - I completely forgot about this.  I will attend to this this week!

PR #1874, opened 2024-04-01
[WIP] Lint codebase

mattwthompson:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,49 @@
...
+  - id: flake8
+- repo: https://github.com/asottile/pyupgrade
+  rev: v3.15.2
+  hooks:
+  - id: pyupgrade
+    args: ["--py39-plus"]
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  rev: v0.3.4
+  hooks:
+    - id: ruff
```
ctk3b, line 35: With ruff, should be able to drop black, isort, flake8

ctk3b (commented):
So linted

.pre-commit-config.yaml
mattwthompson: Yep! That's the goal, just took a bad path to get there.

I added it only to get the NumPy 2.0 rules added in, then started adding in some of its other rules, and then got stuck in a huge mess of stuff and put a decided to deal with it later 🫠

Issue #1831, opened 2023-11-05
Memory leak in NetCDF reporter


sef43:
This is a duplicate of https://github.com/mdtraj/mdtraj/issues/1345 which is currently closed.

It recently appeared again as an OpenMM issue: https://github.com/openmm/openmm/issues/4298

Using the mdtraj.NetCDFReporter causes eventual simulation slow down due to a memory leak. The issue comes from the scipy.io.netcdf library. It can be fixed by using the NetCDF4 library as done here: https://github.com/ParmEd/ParmEd/issues/619

mattwthompson:
This one? https://pypi.org/project/netCDF4/

sef43:
Yep thats the one!. Both ParmEd and MDAnalysis use that library for writing NetCDF if it is available. However they do not have it as a dependency and they both fallback to the scipy version if the NetCDF4 python package is not installed. This seems to be due to some compatibilities with how Amber links to its own NetCDF library. There is more discussion here: https://github.com/MDAnalysis/mdanalysis/issues/506#issuecomment-225081416 (Not sure if this is still a problem in 2023)

mattwthompson:
It looks like this is the same project packaged to conda, which we might want to include or nudge users towards installing in the next version? https://anaconda.org/conda-forge/netcdf4

It's at least used in a bunch of projects I vaguely recognize: https://gist.github.com/mattwthompson/d290dba85668e403b0d5a5451ca256ef

mattwthompson:
#1831 / 1.10

PR #1878, opened 2024-04-18
Sets netCDF4 as default in NetCDFTrajectoryFile, fallback to scipy

janash:
This PR addresses #1831.

I've adopted the strategy suggested by the Issue author, based on an approach taken in ParmEd and MDAnalysis, to use netCDF4 if installed and fall back to SciPy netCDF if not installed.

I based the code for netCDF4 on the original version in the code base. It used netCDF4 before SciPy netCDF, PR that changed it here - https://github.com/mdtraj/mdtraj/pull/360/files

Here is a small demo notebook that demonstrates both backends being used: https://github.com/janash/mdtraj-netcdf/blob/main/openmm.ipynb

I also updated the message about `netCDF4` to give information about `conda` and have an updated URL

One thing I'm not quite sure about is using `import_`. It gives a very verbose message, but it's not actually causing an Exception with my current use, so may be confusing.

### DIFF
--- a/mdtraj/formats/netcdf.py
+++ b/mdtraj/formats/netcdf.py
@@ -125,17 +125,42 @@ class NetCDFTrajectoryFile(object):
         self._closed = True   # is the file currently closed?
         self._mode = mode      # what mode were we opened in
 
-        netcdf = import_('scipy.io').netcdf_file
+        try:
+            # import netcdf4 if it's available
+            netcdf = import_('netCDF4').Dataset
 
-        if mode not in ['r', 'w']:
-            raise ValueError("mode must be one of ['r', 'w']")
+            # mode check for netCDF4
+            if mode not in ['r', 'w', 'a', 'ws', 'as']:
+                raise ValueError(("mode must be one of ['r', 'w', 'a', 'ws', 'as'] if using netCDF4."
+                    " 'r' indicates read, 'w' indicates write, and 'a' indicates"
+                    " append. 'a' and 'w' can be appended with 's', which turns "
+                    " off buffering"))
+            
+            # set input args for netCDF4
+            input_args = {'format': 'NETCDF3_64BIT', 'clobber': force_overwrite}
+
+        except ImportError:
+            netcdf = import_('scipy.io').netcdf_file
+
+            # warn the user, even though the import above also gives
+            # them a big warning. 
+            warnings.warn('Could not find netCDF4 module. Falling back on '
+                              'scipy implementation, which can be significantly'
+                              'slower than the netCDF4 implementation.')
+            
+            # mode check for scipy.io.netcdf_file
+            if mode not in ['r', 'w']:
+                raise ValueError("mode must be one of ['r', 'w'] if using scipy.io.netcdf_file.")
+            
+            # input args for scipy.io.netcdf_file
+            # AMBER uses the NetCDF3 format, with 64 bit encodings, which
+            # for scipy.io.netcdf_file is "version=2"
+            input_args = {'version': 2}
 
         if mode == 'w' and not force_overwrite and os.path.exists(filename):
             raise IOError('"%s" already exists' % filename)
 
-        # AMBER uses the NetCDF3 format, with 64 bit encodings, which
-        # for scipy.io.netcdf_file is "version=2"
-        self._handle = netcdf(filename, mode=mode, version=2)
+        self._handle = netcdf(filename, mode=mode, **input_args)
         self._closed = False
 
         # self._frame_index is the current frame that we're at in the
--- a/mdtraj/formats/netcdf.py
+++ b/mdtraj/formats/netcdf.py
@@ -33,7 +33,6 @@ The code is heavily based on amber_netcdf_trajectory_tools.py by John Chodera.
 
 import os
 import socket
-import sys
 import warnings
 from datetime import datetime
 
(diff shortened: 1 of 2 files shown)


mdtraj/utils/delay_import.py
```diff
@@ -78,15 +78,15 @@
     sharing of array-oriented scientific data.
 
     netcdf4-python can be downloaded from https://pypi.python.org/pypi/netCDF,
```
mattwthompson, line 80: IIUC, just based on following links, this should be 
```suggestion
    netcdf4-python can be downloaded from https://pypi.python.org/pypi/netCDF4,
```

mdtraj/utils/delay_import.py
```diff
@@ -78,15 +78,15 @@
...
-    or installed with the python "easy_install" or "pip" package managers using:
+    or installed with the python "conda" or "pip" package managers using:
 
-    # easy_install netCDF4
```
mattwthompson, line 83: 😅 blast from the past!

mattwthompson (approved):
Thanks, this looks good - I just have one question about how best to proceed

I wonder if it would be better to simply assume that `netCDF4` is installed, but leave the fallback code in place in case there's some other error at import time? We could require it at the packaging level in the next release. I agree the warning is a bit confusing, but I think it's better to have separate "this isn't installed at all" and "we tried to use this module, but something went wrong" warnings.


The `import_` magic is a big confusing to me, too. I think it could be rewritten to better reflect how required & optional dependencies are dealt with today, but that's out of scope for this PR.

janash:
> Thanks, this looks good - I just have one question about how best to proceed
> 
> I wonder if it would be better to simply assume that `netCDF4` is installed, but leave the fallback code in place in case there's some other error at import time? We could require it at the packaging level in the next release. I agree the warning is a bit confusing, but I think it's better to have separate "this isn't installed at all" and "we tried to use this module, but something went wrong" warnings.
> 
> The `import_` magic is a big confusing to me, too. I think it could be rewritten to better reflect how required & optional dependencies are dealt with today, but that's out of scope for this PR.

As far as the `import_` function goes, I think it would be beneficial to evaluate its use. Another option is to not use it at all, let a normal import error occur, and to display a different warning message that simply says that SciPy netcdf is being used, may result in a slow down, and that it is highly recommended to install netcdf4. I actually think this might make the most sense, because the message the user will see (`The code at netcdf.py:130 requires the netcdf4-python module,`) has to do with a file they don't touch at all and probably have no knowledge of. The large message about installing NetCDF4 also detracts from the main point of the warning, which is that they're going to see a performance hit if they don't take some action. If there isn't a strong preference from maintainers, I'd rather remove the use of `import_` in the `try` block, using `import` instead, and make the warning in the `except` block more verbose. I was initially attempting to keep with the original style of the package by using the `import_` function, but it seems like there might not be much attachment to it.

On the other hand, I agree with adding it as a dependency for the next release. It doesn't make sense to me to use it as a default with a big warning message, but to not have it installed as a dependency.  If that's done, the question of whether to use `import_` or `import` becomes much less important :)

mattwthompson:
My preference is

1. Drop the use of `import_` function altogether, along the lines you describe
2. Keep a soft `except ImportError` block, with the warnings you describe and a fallback to the SciPy implementation
3. Just package netCDF4 as a dependency

2. and 3. are in conflict and the warnings may never be emitted to the user, but in case there's some funky with a netCDF implementation it would be better to have the SciPy implementation than not, and the fallback needs to be emitted to the user

mdtraj/formats/netcdf.py
```diff
@@ -127,21 +128,27 @@ def __init__(self, filename, mode='r', force_overwrite=True):
...
+            netcdf = scipy.io.netcdf_file
+
+            warning_message = (
+                "Warning: The 'netCDF4' Python module is not installed. MDTraj is using the 'scipy' "
+                "implementation to read and write netCDF files,which can be significantly slower.\n"
+                "For improved performance, consider installing the 'netCDF4' module. See installation instructions at:\n"
+                "https://unidata.github.io/netcdf4-python/#quick-install"
+            )
             
+            print(warning_message, file=sys.stderr)
```
mattwthompson, line 150: How does printing to STDERR differ from using the built-in `warnings` module? I haven't seen this approach before

janash: When using `warnings` it prints the full file path for the module and also says the warnings type (`UserWarning`). IMO it's a little distracting from the main message.

I noticed that our previous `import_` function printed the message without the extra info in front, and printing to `stderr` is the approach taken in that function - see https://github.com/mdtraj/mdtraj/blob/d785406885f6c65da61f9992122f17a0a8695c12/mdtraj/utils/delay_import.py#L188

janash: If you want to see what the updated warning looks like, I've updated this example notebook: https://github.com/janash/mdtraj-netcdf/blob/main/openmm.ipynb

janash: Although, I suppose one could argue that in the `import_` function it's being used with an exception being raised, so might be more appropriate there. 🤔

mattwthompson: Hmm, I'd prefer everything use `warnings` because it's more configurable and IIUC the standard practice, but it's fine to keep this how it currently is

janash: alrighty, `warnings` it is :)

PR #1879, opened 2024-04-23
chain_id field when using to/from_dataframe

gph82:
ChainID/chain_id attribute was recently (#1715) implemented for pdbfiles, it would be nice if the attribute survives  the to_dataframe/from_dataframe cycle.

Previous behavior:
* `to_dataframe` had a  `chainID` column in the DataFrame, but placed the `atom.residue.chain.index` attribute in it.
* `from_dataframe` read the `chainID` but used it only to detect new chains, it did not get passed to `add_chain`.

New behavior:
* Try to store the actual `chain_id` attribute in the `chainID` column (else fallback to `index`)  when using `top.to_dataframe`
* Add the `chainID` column as `chain_id` when using `from_dataframe`.

### DIFF
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -449,7 +449,8 @@ class Topology(object):
         pd = import_('pandas')
         data = [(atom.serial, atom.name, atom.element.symbol,
                  atom.residue.resSeq, atom.residue.name,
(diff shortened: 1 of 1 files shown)


gph82:
Sorry, lots of tests are failing, I'll get back when I've done my homework

PR #1881, opened 2024-04-26
CI macos-latest fix

jeremyleung521:
Just in case any of the maintainers are wondering, GitHub Action runner images for `macos-latest` has shifted to `macos-14-arm64` a few days ago (from `macos-12`, which is x86-64). That's probably causing the fails in the scheduled CI.

Changing from `macos-latest` to `macos-13` should fix this for the time being, until the arm64 workflows are ready. `macos-13` is the last "standard" (i.e., free) x86-64 runner image from GitHub.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -47,8 +47,8 @@ jobs:
 
     - name: Environment Information
       run: |
-        mamba info
-        mamba list
+        micromamba info
+        micromamba list
 
     - name: Run tests
       if: always()
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -24,7 +24,8 @@ jobs:
 
       matrix:
         os:
-          - macOS-latest
+          - macOS-latest  # arm64
+          - macOS-13  # x86-64
           - ubuntu-latest
         python-version:
           - "3.10"
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -24,7 +24,7 @@ jobs:
 
       matrix:
         os:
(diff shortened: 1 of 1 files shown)


mattwthompson (approved):
Thanks, hopefully we don't need to drop all the way down to 12 like I have had to for some other projects 🫠

mattwthompson:
Could you `:%s/conda/micromamba/` here? https://github.com/mdtraj/mdtraj/pull/1881/files#diff-71cabc4177e41ea8f15a89eb65398fa8187739f68a2762706d84885cd8b2ddc3R50-R51

It's an old, weird alias that for some reason isn't consistent across platforms

jeremyleung521:
> Could you `:%s/conda/micromamba/` here? https://github.com/mdtraj/mdtraj/pull/1881/files#diff-71cabc4177e41ea8f15a89eb65398fa8187739f68a2762706d84885cd8b2ddc3R50-R51
> 
> It's an old, weird alias that for some reason isn't consistent across platforms

Done!

jeremyleung521:
I tested on my own fork and seems like the swap to `macos-13` was not necessary and the problem was just conda/micromamba switch. I'll switch back to macos-latest.

https://github.com/jeremyleung521/mdtraj/actions/runs/8851323246/job/24307625901?pr=2

mattwthompson:
What about testing on both 13 and 14? It seems like things should be working on both https://github.com/mdtraj/mdtraj/actions/runs/8841567954/job/24278867705

mattwthompson:
Thanks again - pretty surprised you haven't gotten a commit through yet, this should make future contributions smoother

PR #1884, opened 2024-05-11
Disable intrinsics when SIMD is not available

natsukium:
`intrinsics` can only be turned off by an environment variable, but even if a CPU does not support SIMD, it will be enabled and the build will fail if this variable is not defined.
I suggest disabling intrinsics by default if sse2 or neon is not available.


mattwthompson:
I'm fine with this, but would like to ask a few questions since I don't know what's going on here

* what CPUs don't support SIMD these days?
* is the problem just that MDTraj has these two SIMD instructions implemented but there are others that might be what somebody's hardware needs?

Issue #1860, opened 2024-03-14
1.10 release planning


mattwthompson:
It's been a while since a release and I'd like to make a new one.

I've created a milestone to track some changes we should make an attempt to include: https://github.com/mdtraj/mdtraj/milestone/12

If I've missed any issues or PRs, please tag them or ping me and I'll squint at them, make a funny face, and guess if each is something we can include in a reasonable timeframe.

This should target Python 3.10, 3.11, and 3.12, following https://scientific-python.org/specs/spec-0000/

chrisjonesBSU:
Should this release add compatibility for MacOSX + python 3.12? From reading around the commits on the feedstock, It looks like the issue was related to the .tng code, which I think was removed in the latest PR.

mattwthompson:
Yep, that's my understanding as well

mattwthompson:
I tagged a `1.10.0rc1` release which should fix a handful of issues that have popped up over the past year, though not all PRs are merged through.

Try it with

```shell
$ pip install mdtraj==1.10.0rc1
```

or
```shell
$ micromamba install mdtdraj -c conda-forge/label/mdtraj_rc -c conda-forge
```

and let us know if anything is surprising. If no feedback this will become a full release in a few weeks

sukritsingh:
I've downloaded this release candidate and quickly am running it on my analysis/MSM scripts and so far no crashes! Thanks for this heroic effort. I'll take a look at some of the other PRs and see if I can help merge/review!

(Also apologies for the big hiatus! I got award a grant and have only just now had life return to some degree of normal pace lol.)

jeremyleung521:
Just wanted to add that the `numpy<2` pin is not yet included in the rc, so it might be worth adding into 1.10 before release (at least until mdtraj supports numpy 2).

https://github.com/mdtraj/mdtraj/issues/1873

mattwthompson:
Good idea - do I do that here? https://github.com/mdtraj/mdtraj/blob/1.10.0rc1/pyproject.toml#L5-L6

mattwthompson:
I made https://github.com/mdtraj/mdtraj/releases/tag/1.10.0rc2

mattwthompson:
https://github.com/mdtraj/mdtraj/discussions/1890

Wheels (#1833) would be nice, too, if we can figure out how to set that up

PR #1886, opened 2024-05-28
Add explicit NumPy 1.x pin where `pip`-like tools look

mattwthompson:
https://github.com/mdtraj/mdtraj/issues/1860#issuecomment-2123324794

@jeremyleung521


jeremyleung521:
Might also want to pin runtime numpy version in setup.py. 

https://github.com/mdtraj/mdtraj/blob/master/setup.py#L309

Issue #1850, opened 2024-01-24
Errors Installing mdtraj from source


jeremyleung521:
I was trying to test my `mdtraj` branch more efficiently but kept encountering significant issues with installing mdtraj from source. I've been able to replicate the same errors on two computers and three different conda installation environments (macos-x86_64 under Rosetta/ macos-arm64/ ubuntu 22.04 on x86_64). I've tried using python 3.9/3.10/3.11/3.12 environments. 

1. Trying to install in editable mode (`python -m pip install -e .`) always results in the following error. Since it failed out, mdtraj is not installed.
```
        File "/private/var/folders/06/29rlbptd5jq_rfl__96vd2qh0000gn/T/pip-build-env-kdg7kr2f/overlay/lib/python3.12/site-packages/setuptools/_distutils/command/build_ext.py", line 467, in build_extensions
          self._build_extensions_serial()
        File "/private/var/folders/06/29rlbptd5jq_rfl__96vd2qh0000gn/T/pip-build-env-kdg7kr2f/overlay/lib/python3.12/site-packages/setuptools/_distutils/command/build_ext.py", line 493, in _build_extensions_serial
          self.build_extension(ext)
        File "/Users/user/Documents/Important/Pitt/git/mdtraj/basesetup.py", line 302, in build_extension
          self.build_static_extension(ext)
        File "/Users/user/Documents/Important/Pitt/git/mdtraj/basesetup.py", line 359, in build_static_extension
          src_tree_output_dir = re.match('build.*(mdtraj.*)', output_dir).group(1)
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      AttributeError: 'NoneType' object has no attribute 'group'
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building editable for mdtraj
Failed to build mdtraj
ERROR: Could not build wheels for mdtraj, which is required to install pyproject.toml-based projects
```

2. Installing in non-editable mode (`python -m pip install .`) would be successful, but I would always end up with this error during `import mdtraj`. **Oddly, this no longer happens if I try it from a new console** (i.e., start a new terminal window). I _SOMETIMES_ get the same error if I install from PyPI source (`python -m pip install mdtraj`).
```
(mdtraj-test2) mc-wifi-10-215-109-101:mdtraj user$ python
[...]

PR #1887, opened 2024-05-28
fix editable mode, add packaging as dependency

jeremyleung521:
#1876 #1850 

Added some try/except blocks during file copy/file name matching in `base_setup.py` as these steps are unnecessary (and fails) during editable mode install.

Also added 'packaging' as an install dependency because it's no longer default as of python 3.12, but needed in `mdtraj/formats/amberrst.py`

### DIFF
--- a/basesetup.py
+++ b/basesetup.py
@@ -307,6 +307,15 @@ release = {release}
         )
 
 
+def copy_same_file_pass(source, destination):
+    try:
+        # Try to copy
+        shutil.copy(source, destination)
+    except shutil.SameFileError:
+        # Ignore the error
+        pass
+
+
 class StaticLibrary(Extension):
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


mattwthompson (approved):
Awesome, this file is mysterious to me and I'm happy somebody else was able to update it

I tested this locally and it behaves just like I expect editable installs to behave; `from mdtraj import __file__` gives me something in the repo and adding an exception at the top of any core file causes imports to crash (without re-installing)

mattwthompson:
It's all pretty confusing to me still - 

1. I know that Setuptools vendors tons of stuff
2. I think `pip` also vendors plenty of stuff
4. `distutils` is gone in 3.12 and shouldn't be imported directly
5. MDTraj is importing it directly, and building fine on 3.12, presumably because of Setuptools's hack?

> I assume in the long run ...

Yep ... "If it works, it works" ... until it doesn't!

PR #1892, opened 2024-06-02
[DTRTrajectoryFile read_as_traj] read only n_frames

OlivierBeq:
This PR ensures only `n_frames` are read (when specified) in a `DTRTrajectoryFile`; it allows `mdtraj.iterload` to work as expected with DESMOND trajectories.

### DIFF
--- a/mdtraj/formats/dtr/dtr.pyx
+++ b/mdtraj/formats/dtr/dtr.pyx
@@ -468,7 +468,7 @@ cdef class DTRTrajectoryFile:
         if atom_indices is not None:
             topology = topology.subset(atom_indices)
 
-        xyz, time, box_length, box_angle = self.read(stride=stride, atom_indices=atom_indices)
+        xyz, time, box_length, box_angle = self.read(n_frames=n_frames, stride=stride, atom_indices=atom_indices)
         if len(xyz) == 0:
             return Trajectory(xyz=np.zeros((0, topology.n_atoms, 3)), topology=topology)
 


sukritsingh:
Thanks for this PR! I'm a bit confused about the intention - we already test for `n_frames` loading in DTR files in all the `test_read` methods in the [DTR Tests](https://github.com/mdtraj/mdtraj/blob/master/tests/test_dtr.py). All these tests pass, so I'm not sure what the issue here being addressed is. Could you expand on the rationale/issue here?

Are we not testing for whether or not we can extract subsets of trajectories? If so we should definitely also add that as a test then i this PR. I confess I'm new to the DTR file side and I haven't gone through them in depth yet.

OlivierBeq:
Only the `read` method of the `DTRTrajectoryFile` class is tested, not the `read_as_traj` method, which is ubiquitously called in the library.

Indeed tests should also focus on the functionality of the `read_as_traj` method.

sukritsingh:
> not the read_as_traj method, which is ubiquitously called in the library.

I think each of the `TrajectoryFile` objects have their own `read_as_traj` method, rather than all of them calling `DTRTrajectoryFile` 

I think I can only find one instance of `DTRTrajectoryFile.read_as_traj` being called is [`_load_desmond_traj`](https://github.com/OlivierBeq/mdtraj/blob/62716f50eee1c2fb6bef857a26f1e5918455b7fd/mdtraj/formats/dtr/dtr.pyx#L73C5-L73C23)
 
> Indeed tests should also focus on the functionality of the read_as_traj method.

All the above aside, it'd be great to have tests for `read_as_traj` as well!

OlivierBeq:
I just added 2 tests for the `DTRTrajectoryFile.read_as_traj` method to ensure either all or the desired number of frames are loaded correctly.

tests/test_dtr.py
```diff
@@ -24,6 +24,7 @@
 import numpy as np
 import pytest
 
+from mdtraj.core.trajectory import _parse_topology
```
sukritsingh, line 27: I understand what you're doing with `_parse_topology` but I'm curious if important an internal method to `Trajectory` objects is the best way we should be loading and parsing topologies for testing. Is there a better way we can load and pass the topologies to `read_as_traj`? I know that in `test_hdf5` we use `md.load_pdb().topology`

tests/test_dtr.py
```diff
@@ -168,6 +169,27 @@ def test_read_8(get_fn):
     assert eq(xyz_ref[:, ::2, :], xyz)
 
 
+def test_read_as_traj_1(get_fn):
+    # Test the default read_as_traj and compare against reference trajectory in dcd format
+    fn_dtr = get_fn("frame0.dtr")
+    fn_dcd = get_fn("frame0.dcd")
+    fn_pdb = get_fn("frame0.pdb")
+    top = _parse_topology(fn_pdb)
```
sukritsingh, line 177: Do we need to load the topology in for testing? Admittedly, I'm basing a lot of my ideas off `test_xtc` and `test_hdf5` (the formats I use most) but in those cases we don't need to load topology for testing. Just considering if this may be unneeded bloat for testing purposes (or maybe topology DTR testing needs to be a separate test)

sukritsingh (changes requested):
This is a necessary addon to the `DTRTrajectoryFile` format where we had most of the capacity to do `read_as_traj` and `mdtraj.iterload` but somehow never enabled `read_as_traj` to load a specified number of frames at a time. This PR addresses this very core need. Thanks for submitting it!

My only questions/sticking points on this PR are regarding the added tests. I'm not the most familiar with `test_dtr` or (`DTRTrajectoryFile` objects as a whole) but based on consistency with `test_xtc` and `test_hdf5`, I'm not convinced we need to import an internal method like `Trajectory._parse_topology` to achieve this testing (mainly just to maintain consistency and avoid unneeded bloating).

Issue #1893, opened 2024-06-03
NETCDF4 Error


jeremyleung521:
If I have `netCDF4` installed, I get the following error with mdtraj `v1.10.0` when I try to load an ``.nc`` file. Probably related to the changes in #1878 . The nc file is created from a pdb file with mdtraj. This works fine in `v1.9.9` (when I downgraded).

@janash 

```
----> 1 a = mdtraj.load('ntl9.nc', top='ntl9.pdb')

File ~/miniconda3/envs/westpa-workshop2024/lib/python3.10/site-packages/mdtraj/core/trajectory.py:433, in load(filename_or_filenames, discard_overlapping_frames, **kwargs)
    423 filename_or_filenames = filename_or_filenames[1:]  # ignore first file
    424 try:
    425     # this is a little hack that makes calling load() more predictable. since
    426     # most of the loaders take a kwargs "top" except for load_hdf5, (since
   (...)
    431     # TODO make all the loaders accept a pre parsed topology (top) in order to avoid
    432     # this part and have a more consistent interface and a faster load function
--> 433     t = loader(tmp_file, **kwargs)
    435 except TypeError as e:
    436     # Don't want to intercept legit
    437     # TypeErrors
    438     if "got an unexpected keyword argument 'top'" not in str(e):

File ~/miniconda3/envs/westpa-workshop2024/lib/python3.10/site-packages/mdtraj/formats/netcdf.py:91, in load_netcdf(filename, top, stride, atom_indices, frame)
     88 topology = _parse_topology(top)
     89 atom_indices = cast_indices(atom_indices)
---> 91 with NetCDFTrajectoryFile(filename) as f:
     92     if frame is not None:
     93         f.seek(frame)

File ~/miniconda3/envs/westpa-workshop2024/lib/python3.10/site-packages/mdtraj/formats/netcdf.py:169, in NetCDFTrajectoryFile.__init__(self, filename, mode, force_overwrite)
    166 if mode == "w" and not force_overwrite and os.path.exists(filename):
    167     raise OSError('"%s" already exists' % filename)
--> 169 self._handle = netcdf(filename, mode=mode, **input_args)
    170 self._closed = False
    172 # self._frame_index is the current frame that we're at in the
    173 #     file
    174 # self._needs_initialization indicates whether we need to set the
    175 #     global properties of the file. This is required before the first
    176 #     write operation on a new file

TypeError: netcdf_file.__init__() got an unexpected keyword argument 'format'
```

[...]

PR #1894, opened 2024-06-03
Fix netCDF file loading to no longer import scipy (and add netCDF to CI reqs)

sukritsingh:
A quick PR to address #1893 - changes below: 

1. Removes the line 

` netcdf = import_("scipy.io").netcdf_file `

from `formats/netcdf.py` which should fix netCDF4 loading issues.

2. Updated the `conda-envs/test.yaml` file to include netCDF4 in the requirements. This should ensure netCDF4 is in the CI env.

Let me know if I'm missing something!

### DIFF
--- a/devtools/conda-envs/test.yaml
+++ b/devtools/conda-envs/test.yaml
@@ -16,3 +16,4 @@ dependencies:
   - gsd>=2.8
   - pytest
   - pytest-xdist
+  - netCDF4
--- a/mdtraj/formats/netcdf.py
+++ b/mdtraj/formats/netcdf.py
@@ -158,8 +158,6 @@ class NetCDFTrajectoryFile:
             # for scipy.io.netcdf_file is "version=2"
             input_args = {"version": 2}
 
-        netcdf = import_("scipy.io").netcdf_file  # noqa: F841
-
         if mode not in ["r", "w"]:
             raise ValueError("mode must be one of ['r', 'w']")
 


sukritsingh:
Ugh looks like all the `test_netcdf.py` tests are now failing - sample below: 

```
FAILED tests/test_netcdf.py::test_read_write_1 - AssertionError: o1 and o2 not the same type: <class 'numpy.ma.core.MaskedArray'> <class 'numpy.ndarray'>
```

Will have to dig into what else needs to change but if someone sees the fix then please shout!

jeremyleung521:
Assuming netcdf4 is returning a [masked view of the array](https://numpy.org/doc/stable/reference/maskedarray.generic.html) instead of a copy (or whatever). Might be worth loosening this assert line so it accepts `MaskedArrays`.

https://github.com/mdtraj/mdtraj/blob/f134696b3d6e3ae26b00221394779edcfc00d898/mdtraj/testing/testing.py#L138

sukritsingh:
Hrmmm looking at the [changelog](https://github.com/Unidata/netcdf4-python/blob/master/README.md) for netCDF4 it looks like it might indeed be returning a Masked Array: 
```
Slicing a netCDF variable will now always return masked array by default, even if there are no masked values. The result depended on the slice before, which was too surprising.
```

> Might be worth loosening this assert line so it accepts MaskedArrays.

I guess either would be fine here...Let me try it

sukritsingh:
Unfortunately it looks like `eq` is too ubiquitously used to be a simple loosening of that line alone.

Also worth noting that it'll probably need to be loosened here: 
https://github.com/mdtraj/mdtraj/blob/f134696b3d6e3ae26b00221394779edcfc00d898/mdtraj/testing/testing.py#L146

Is there be a way to force netCDF4 to output an `ndarray` and not a `MasekdArray`? That would be the quickest fix I see.

jeremyleung521:
Wouldn't something like this work?

```
if isinstance(o1, (np.ndarray, np.ma.MaskedArray)) and isinstance(o2, (np.ndarray, np.ma.MaskedArray)):
    pass
else:
    assert type(o1) is type(o2), f"o1 and o2 not the same type: {type(o1)} {type(o2)}"
```

MaskedArray is a subclass of ndarray, so the isinstance() should work.

jeremyleung521:
Or something even simpler like:

`assert (isinstance(o1, type(o2)) or isinstance(o2, type(o1)))  , f"o1 and o2 not the same type: {type(o1)} {type(o2)}"`

sukritsingh:
Oh right - thanks for the suggestion. Let me try the first one you mentioned.

> Or something even simpler like:
assert (isinstance(o1, type(o2)) or isinstance(o2, type(o1)))  , f"o1 and o2 not the same type: {type(o1)} {type(o2)}"

This line is less specific and might lead to false positives in some cases where o1 and o2 aren't actually compatible down the line, so I'm iffier on it.

sukritsingh:
Success! @jeremyleung521 want to try pulling this PR really quick and testing it to verify on your files?

mattwthompson (approved):
This looks good to me as-is (as long as @jeremyleung521's use case is fixed!)

I'd really like to get @janash's suggestion of mocking `netCDF4` _not_ being installed in here, either now or later

jeremyleung521:
> Oh right - thanks for the suggestion. Let me try the first one you mentioned.
> 
> > Or something even simpler like:
> > assert (isinstance(o1, type(o2)) or isinstance(o2, type(o1)))  , f"o1 and o2 not the same type: {type(o1)} {type(o2)}"
> 
> This line is less specific and might lead to false positives in some cases where o1 and o2 aren't actually compatible down the line, so I'm iffier on it.

It's essentially checking if one is a subclass (or the same class) as the other. If it doesn't work out, it'd raise an error[ down here](https://github.com/mdtraj/mdtraj/blob/f134696b3d6e3ae26b00221394779edcfc00d898/mdtraj/testing/testing.py#L166-L167), anyways.

Anyhow, the current PR works with my files!

sukritsingh:
> I'd really like to get @janash's suggestion of mocking netCDF4 not being installed in here, either now or later

Agreed 100%, but it might be better suited to a separate PR unless you'd like that mocking to be added in parallel? I confess I'm new to mocking so if someone is more comfortable with it then it'd probably be faster as a separate PR.

As an alternative, I could add `netCDF4` as an explicit dependency in this PR, but I wasn't sure if we want that to be explicitly required (or at least, I'd like to hear opinions first).

sukritsingh:
With #1895 handling the mock-testing situation, I think the only thing that remains to decide: 
1. Do we want netCDF4 to be an explicit dependency listed in `setup.py` (alongside the `test.yaml`)?
2. Do we want that to be in this PR or a separate one?

My inclination is 1. Yes, and 2. Including it here can't hurt but no strong feelings if people would rather that be a separate PR.

mattwthompson:
1. Lean yes but don't really care
2. Separate PR preferred IMO

PR #1898, opened 2024-06-05
AST-related modifications

jeremyleung521:
These were some stuff I had written previously and thought to finish.

### Deprecating astunparse

I've swapped out `astunparse` from anything `>= python 3.9` because `ast.unparse` is available in stdlib since python 3.9 (also `astunparse` is no longer actively supported).  Behavior is slightly different (i.e., not as verbose with parenthesis and linebreaks) so I've added/modified the existing tests to test both `astunparse` and `ast.unparse`.

### Removing the last of `ast.Num`/`ast.Str`

In #1838, most instances of `ast.Num`, `ast.Str` (and other deprecated classes) are swapped over to `ast.Constant`. There is this lone TODO where swapping to `ast.Constant` is too broad and breaks behavior. I believe it's intended to catch lone words that result from [this line](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/selection.py#L78), which means it's a lone `ast.Constant` and not a "safe name". 

All tests passed and same behavior as before (v1.9.9) with all sorts of keywords thrown at it.

```
In [1]: from mdtraj.core.selection import parse_selection

In [2]: parse_selection('bar')
ValueError                                Traceback (most recent call last)
Cell In[2], line 1
----> 1 parse_selection('bar')

File ~/Documents/Important/Pitt/git/mdtraj/mdtraj/core/selection.py:418, in parse_selection.__call__(self, selection)
    415 # Special check for a single literal
    417 if isinstance(astnode, ast.Constant) and astnode.value not in {True, False, None}:
--> 418     raise ValueError(
    419         "Cannot use a single literal as a boolean. " f"Choked on node with value {astnode.value}",
    420     )
    422 args = [ast.arg(arg="atom", annotation=None)]
    423 signature = ast.arguments(
    424     args=args,
    425     vararg=None,
   (...)
    430     kw_defaults=[],
    431 )

ValueError: Cannot use a single literal as a boolean. Choked on node with value bar

In [3]: parse_selection('protein')
Out[3]: _ParsedSelection(expr=<function <lambda> at 0x103623a30>, source='atom.residue.is_protein', astnode=<ast.Attribute object at 0x1037318d0>)

In [4]: parse_selection('name CA')
Out[4]: _ParsedSelection(expr=<function <lambda> at 0x11dc65240>, source="atom.name == 'CA'", astnode=<ast.Compare object at 0x10362a080>)
```


mdtraj/core/selection.py
```diff
@@ -43,6 +42,12 @@
     quotedString,
 )
 
+try:
+    from ast import unparse
+except ImportError:
+    from astunparse import unparse
```
mattwthompson, line 48: Any reason the stdlib import wouldn't work? We could just drop the `astunparse` dependency altogether

setup.py
```diff
@@ -308,7 +308,7 @@ def geometry_extensions():
     install_requires=[
         "numpy>=1.23,<2.0.0a0",
         "scipy",
-        "astunparse",
+        "astunparse;python_version<'3.9'",
```
mattwthompson, line 311:
```suggestion
```
I don't think this is worth the hassle, we only say we support SPEC 0 anyway and that's long since dropped 3.8 https://scientific-python.org/specs/spec-0000/

tests/test_selection.py
```diff
@@ -21,15 +21,20 @@
...
 import numpy as np
 import pytest
+from pytest import MonkeyPatch
 
 import mdtraj
 from mdtraj.core.selection import parse_selection
 from mdtraj.testing import eq
 
 
+use_astunparse = False
```
mattwthompson, line 35: Might as well only consider the case in which this is `False` and avoid branching out

mattwthompson (commented):
Only considering the first of your two ideas here, haven't looked closely enough to understand the meat of the changes

jeremyleung521:
@mattwthompson This new commit basically reverts all the test changes so we will only support `ast.unparse` (and not `astunparse`). Also pinned `python_version>=3.9`

mdtraj/core/selection.py
```diff
@@ -75,7 +76,7 @@ def visit_Name(self, node):
         # all other bare names are taken to be string literals. Thus something
         # like parse_selection('name CA') properly resolves CA as a string
         # literal, not a barename to be loaded from the global scope!
-        return ast.Constant(s=node.id, kind=None)
+        return ast.Constant(value=node.id, kind=None)
```
mattwthompson, line 79: Haha, did this have a named argument `s` in years past?

Also, what's `kind` do ... ? I don't see it in the docs https://docs.python.org/3.12/library/ast.html#ast.Constant

mdtraj/core/selection.py
```diff
@@ -408,9 +409,8 @@ def __call__(self, selection):
         astnode = self.transformer.visit(deepcopy(parse_result[0].ast()))
 
         # Special check for a single literal
-        # TODO: These classes are deprecated since 3.8 and replaced with ast.Constant,
-        # but that substitution breaks behavior
-        if isinstance(astnode, (ast.Num, ast.Str)):
+
+        if isinstance(astnode, ast.Constant) and astnode.value not in {True, False, None}:
```
mattwthompson, line 413: I tried this a few months ago and I think something around here is where I got confused and gave up

tests/test_selection.py
```diff
@@ -134,15 +134,15 @@ def test_binary_2():
 
 def test_simple():
     sp = parse_selection("protein")
-    eq(sp.source, "atom.residue.is_protein\n")
+    eq(sp.source, "atom.residue.is_protein")
```
mattwthompson, line 137: I don't care about newlines, but is this change because of something you changed or just to make the code look simpler?

mattwthompson (approved):
LGTM, a couple of stray comments for my curiousity

Mostly what I did in the <5 minutes I looked at this just now are
- [x] Look in the CI logs for deprecations warnings
- [x] Eyeball the diff for anything that looks funky 
- [x] Trust that tests are indeed passing

Dropping dependencies is always fun

mdtraj/core/selection.py
jeremyleung521: The AST documentation is kinda bad.  ```kind``` and `s` are listed in the [grammar tree in the main doc](https://docs.python.org/3/library/ast.html#abstract-grammar), but better explained here: 

> kind is 'u' for strings with a u prefix, and None otherwise, allowing tools to distinguish u"a" from "a".
https://greentreesnakes.readthedocs.io/en/latest/nodes.html#Constant


The s was there for backward compatibility reasons. ```ast.Str(s=something)``` There was a deprecation warning recently: https://github.com/mdtraj/mdtraj/actions/runs/9404499617/job/25903544222#step:6:2330
> A string. The s attribute hold the value. In Python 2, the same type holds unicode strings too.
https://greentreesnakes.readthedocs.io/en/latest/nodes.html#Str

tests/test_selection.py
jeremyleung521: Behavior is slightly different between `astunparse.unparse` and  `ast.unparse`(i.e., latter is not as verbose with parenthesis and linebreaks).

mdtraj/core/selection.py
jeremyleung521: All `Num` and `Str` are `Constants`, but not all `Constants` are `Num` and `Str`. And so I had to filter out what comes out of L73 (True/False/None) vs. L79 (everything else). Everything else from L79 could eventually pass through other nodes (like Compare and what not, those will fail the first check), so those would be safe. But stuff that comes directly out of L79 could be single literal ('not') etc. which needs to choke and error out.

At least that's what i think is happening

Issue #1843, opened 2023-12-13
Overwriting an Existing HDF5Reporter


Heezch:
I'm currently working with a Python script with OpenMM where I need the functionality to overwrite an existing HDF5Reporter with the same name in a simulation. My goal is to either append a new reporter or replace the existing one, depending on whether it's already present.

However, I encountered a `ValueError` indicating that the HDF5 file is already opened and needs to be closed before it can be reopened in write mode. I attempted to address this by checking for the error, closing the existing reporter, and then appending a new one. Here's the code snippet I used:

```
from mdtraj.reporters import HDF5Reporter

try:
    simulation.reporters.append(HDF5Reporter(f'{self.name}.h5', 100))
except ValueError as e:
    if 'already opened' in str(e) and self.override:
        # Attempting to close the existing HDF5Reporter
        for reporter in simulation.reporters[:]:  # Copy of the list for iteration
            if isinstance(reporter, HDF5Reporter) and reporter._filename == f'{self.name}.h5':
                reporter.flush()  # Writing out any pending data
                reporter.close()  # Closing the file
                simulation.reporters.remove(reporter)  # Removing the closed reporter
                break
        # Appending a new HDF5Reporter
        simulation.reporters.append(HDF5Reporter(f'{self.name}.h5', 100))
    else:
        raise
```
Despite these steps, the error persists. I also tried deleting the file using `os` module, but it seems the state remains in RAM, leading to the same error (resetting the kernel would allow me to initialize using the same name and overwrite). Am I missing something obvious? If there is no solution within the same state of the kernel, I will probably adjust the filename, but ideally I do not want to do that. Thank you for having a look at my issue.

sukritsingh:
Thanks for bringing this up! Generally I've found that once you attach the reporter to the `simulation` object all you have to do is run a loop or something simple to append everything as it's running in a single simulation.

> I need the functionality to overwrite an existing HDF5Reporter with the same name in a simulation
 Can you provide a full script example of your usecase where you're running into this issue? You may need to reinitialize the `simulation` object to overwrite/append the existing reporter (otherwise OpenMM will keep it open).

I confess I've only used the OpenMM reporters and have only ever converted completed trajectories into HDF5 files, so I haven't used this reporter myself.

sukritsingh:
Noting that appending in general appears to be tough with HDF5 files - this issue has been brought up before in #1750 - would welcome a PR fix to this!

Heezch:
[...]

Issue #1750, opened 2022-08-25
Support for appending mode for hdf5 writer?


ziyuanzhao2000:
Hi, I'm working on a rather large dataset and would appreciate the ability to continuously append chunks of trajectory to a hdf5 file, much like what the HDF5 reporter does during a MD simulation. I think allowing an optional argument [here](https://github.com/mdtraj/mdtraj/blob/ba929a806bbb5eef14806130cedbfb647132d068/mdtraj/core/trajectory.py#L1396) to specify 'w'/ 'a' append mode can easily achieve this. Please let me know if this is reasonable. Thanks!

sukritsingh:
Hello! Sorry for the massive delay on this as we catch up with older issues - yes being able to append chunks to an hdf5 file seems really reasonable. 

A PR would be welcome here. I might try getting to it on my own time but this wouldn't be mission-critical for me so it'll depend on my bandwidth

PR #1899, opened 2024-06-06
Allow appending to HDF5 files using `trajectory.save_hdf5`

sukritsingh:
Addressing #1750 (and maybe #1843) - this adds an additional argument in `core/trajectory.save_hdf5()` called `savemode` that allows you to define either "w" or "a" for write or append mode respectively. I've set it to `savemode="w"` by default.

The added docstring reads
```
        savemode : str, default='w'
            The mode in which to save the file. 'w' will overwrite any existing
            file, 'a' will append to an existing file.
```

I've also added a snipped to check that only "w" or "a" are being passed and raise a `ValueError` otherwise.

### DIFF
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -1474,7 +1474,7 @@ class Trajectory:
         # run the saver, and return whatever output it gives
         return saver(filename, **kwargs)
 
-    def save_hdf5(self, filename, force_overwrite=True):
+    def save_hdf5(self, filename, savemode="w", force_overwrite=True):
         """Save trajectory to MDTraj HDF5 format
 
         Parameters
@@ -1483,8 +1483,15 @@ class Trajectory:
             filesystem path in which to save the trajectory
         force_overwrite : bool, default=True
             Overwrite anything that exists at filename, if its already there
+        savemode : str, default='w'
+            The mode in which to save the file. 'w' will overwrite any existing
+            file, 'a' will append to an existing file.
         """
-        with HDF5TrajectoryFile(filename, "w", force_overwrite=force_overwrite) as f:
+        # check if savemode is valid (only "w" or "a" are allowed)
+        if savemode not in ["w", "a"]:
+            raise ValueError("savemode must be either 'w' or 'a'")
+        
+        with HDF5TrajectoryFile(filename, savemode, force_overwrite=force_overwrite) as f:
             f.write(
                 coordinates=in_units_of(
                     self.xyz,


sukritsingh:
Worth noting that we already [test if `HDF5TrajectoryFile` can append correctly](https://github.com/mdtraj/mdtraj/blob/15514a57a98bebb3db6f6b6fdc3832f06a6197df/tests/test_hdf5.py#L337-L346) so I don't think we need additional testing? Open to adding more.

jeremyleung521:
I'm in favor of just calling it `mode` (rather than `savemode`) so it's in line with the `HDF5TrajectoryFile` keyword. It's also gnarly defaulting to 'w' but I guess that's just maintaining the previous behavior.

sukritsingh:
> I'm in favor of just calling it mode (rather than savemode) so it's in line with the HDF5TrajectoryFile keyword. 

good point! made the change! 

> It's also gnarly defaulting to 'w' but I guess that's just maintaining the previous behavior.

Yeah it defaults to "w" and I think it forces overwrite to be true which is a) consistent with previous behavior and b) consistent with all the other type of Trajectory file formats as well; Changing just the one seems like a bad idea.

sukritsingh:
This is a minor edit in the codebase and doesn't futz with much but just wanted to make sure nobody else has thoughts particularly @mattwthompson

sukritsingh:
Will go ahead and merge this later in the next few days if there are no further comments/points!

mdtraj/core/trajectory.py
```diff
@@ -1,10 +1,10 @@
 ##############################################################################
 # MDTraj: A Python Library for Loading, Saving, and Manipulating
 #         Molecular Dynamics Trajectories.
-# Copyright 2012-2022 Stanford University and the Authors
+# Copyright 2012-2024 Stanford University and the Authors
 #
 # Authors: Robert McGibbon
-# Contributors: Kyle A. Beauchamp, TJ Lane, Joshua Adelman, Lee-Ping Wang, Jason Swails
+# Contributors: Sukrit Singh, Kyle A. Beauchamp, TJ Lane, Joshua Adelman, Lee-Ping Wang, Jason Swails
```
mattwthompson, line 7: Neither here nor there: I'd rather info in headers like this just exist in one place instead of every file

mattwthompson (approved):
LGTM, I know little about HDF5 but passing an argument through seems straightforward enough

mdtraj/core/trajectory.py
sukritsingh: Yeah I'm just maintaining and updating the existing header on the file since it predates me. I figure that's the "best practice" for now and we can discuss where the singular place should be long term during a bigger refactor

Issue #1739, opened 2022-05-13
CONECT records in PDB files written by mdtraj always use atom index


proteins247:
This is a curious issue brought on by https://github.com/mdtraj/mdtraj/issues/1583, which was solved by https://github.com/mdtraj/mdtraj/pull/1584.

As mdtraj currently is coded, if you are write a pdb file with a structure that has only a single chain, then the atom indexes written in the atom serial number column in the file correspond to the `serial` attribute of each atom. This is so that in situations such as using `atom_slice`, the original atom serial numbers are still preserved in the output, as opposed to being renumbered from 1. 

(Note that as a result of https://github.com/mdtraj/mdtraj/issues/1611 and https://github.com/mdtraj/mdtraj/pull/1612, structures with more than one chain are always given atom serial numbers with a sequential numbering starting from 1.)

The issue is that in the footer of the file, the CONECT records are always written using the sequential serial numbering starting from 1. So, in the scenario of writing a subset of the atoms from a larger system, if the subset is all in one chain, currently you can end up with something like this:

```
REMARK   1 CREATED WITH MDTraj 1.9.7, 2022-05-12
CRYST1   34.320   45.510   98.910  90.00  90.00  90.00 P 1           1 
MODEL        0
ATOM   2497  N1  FOL A 201      -3.795   4.618 -12.297  1.00  0.00           N  
ATOM   2498  C2  FOL A 201      -2.963   5.714 -12.359  1.00  0.00           C  
ATOM   2499  NA2 FOL A 201      -3.462   6.897 -12.700  1.00  0.00           N  
ATOM   2500  N3  FOL A 201      -1.621   5.582 -12.072  1.00  0.00           N  
ATOM   2501  C4  FOL A 201      -1.101   4.354 -11.718  1.00  0.00           C  
ATOM   2502  O4  FOL A 201       0.119   4.226 -11.623  1.00  0.00           O  
ATOM   2503  C4A FOL A 201      -1.940   3.242 -11.647  1.00  0.00           C  
ATOM   2504  N5  FOL A 201      -1.444   2.001 -11.294  1.00  0.00           N  
ATOM   2505  C6  FOL A 201      -2.290   0.912 -11.227  1.00  0.00           C  
ATOM   2506  C7  FOL A 201      -3.638   1.069 -11.520  1.00  0.00           C  
ATOM   2507  N8  FOL A 201      -4.134   2.302 -11.881  1.00  0.00           N  
[...]

PR #1904, opened 2024-06-21
fix pdbstructure "vmd convention" false-positive

### LINKED ISSUE #1903: MDTraj can write some PDB files that it cannot read
I think that solution proposed in this PR is not consistent with `mdtraj` writing functionality of pdb files. This can lead to a situation, when a structure read and then saved by `mdtraj` cannot be read again. Let me provide an example:
Let's assume that I have a cif file with a huge amount of waters (
`.cif` file in the archive [mdtraj_numbering_issue.zip](https://github.com/user-attachments/files/15928227/mdtraj_numbering_issue.zip)
). Then I can load this file to `mdtraj` and save it to `pdb`:
```python
st = md.load("tmp_hoh.cif")
st.save("tmp_hoh_pdb_1.pdb")
```
The residue numbering in the created file for waters will go from 1 to 9999 and then from 0 to 4989. This is in accordance with [writing behaviour](https://github.com/mdtraj/mdtraj/blob/ccd0c874361e4e31127e3803d3670d08d57f7933/mdtraj/formats/pdb/pdbfile.py#L396)

Then if I read the created file again and write it again:
```python
st = md.load("tmp_hohpdb_1.pdb")
st.save("tmp_hoh_pdb_2.pdb")
```
The ordering in the created file is already different: it goes from 1 to 2709, and then it is reset to 0 (thinking that the formatting is `hex`). This is still maybe ok, but then `mdtraj` cannot read this file anymore:
```python
st = md.load("tmp_hohpdb_2.pdb")
```
leads to an error:
```
  File "/home/pbuslaev/conda_env/astex_fep_dev/lib/python3.11/site-packages/mdtraj/core/trajectory.py", line 396, in load
    kwargs["top"] = _parse_topology(
                    ^^^^^^^^^^^^^^^^
  File "/home/pbuslaev/conda_env/astex_fep_dev/lib/python3.11/site-packages/mdtraj/core/trajectory.py", line 170, in _parse_topology
    _traj = load_frame(top, 0, **kwargs)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pbuslaev/conda_env/astex_fep_dev/lib/python3.11/site-packages/mdtraj/core/trajectory.py", line 313, in load_frame
    return loader(filename, frame=index, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pbuslaev/conda_env/astex_fep_dev/lib/python3.11/site-packages/mdtraj/formats/pdb/pdbfile.py", line 152, in load_pdb
    with PDBTrajectoryFile(filename, standard_names=standard_names, top=top) as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pbuslaev/conda_env/astex_fep_dev/lib/python3.11/site-packages/mdtraj/formats/pdb/pdbfile.py", line 289, in __init__
    self._read_models()
  File "/home/pbuslaev/conda_env/astex_fep_dev/lib/python3.11/site-packages/mdtraj/formats/pdb/pdbfile.py", line 608, in _read_models
    pdb = PdbStructure(self._file, load_all_models=True)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pbuslaev/conda_env/astex_fep_dev/lib/python3.11/site-packages/mdtraj/formats/pdb/pdbstructure.py", line 283, in __init__
    self._load(input_stream)
  File "/home/pbuslaev/conda_env/astex_fep_dev/lib/python3.11/site-packages/mdtraj/formats/pdb/pdbstructure.py", line 299, in _load
    self._add_atom(Atom(pdb_line, self))
                   ^^^^^^^^^^^^^^^^^^^^
  File "/home/pbuslaev/conda_env/astex_fep_dev/lib/python3.11/site-packages/mdtraj/formats/pdb/pdbstructure.py", line 875, in __init__
    self.residue_number = _read_residue_number(pdb_line[22:26], pdbstructure, None, self)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pbuslaev/conda_env/astex_fep_dev/lib/python3.11/site-packages/mdtraj/formats/pdb/pdbstructure.py", line 150, in _read_residue_number
    pdbstructure._residue_num_nondec_mode = pdbstructure._residue_num_initial_nodec_vals[
                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: '   0'
```

So it is at least strange that `mdtraj` is producing structures that can not be read back into `mdtraj`

_Originally posted by @pbuslaev in https://github.com/mdtraj/mdtraj/issues/1849#issuecomment-2182757651_

jeremyleung521:
Fixes #1903 

So apparently the reader was misreading all residue numbers 2710+ as vmd overflow? Probably was too focused on getting it to work back in the day that I missed the potential false positives.

I was able to read/write the test files made by the same process in #1903. There was also a bug where the residue number for the TER line would overflowing if more that 9999 residues. This also fixes that.

```
# INCORRECT
TER   44980      HOH D14989
---
# CORRECT
TER   44980      HOH D4989

```

Updated some tests that will catch this as well.


sukritsingh:
So we don't keep finding more and more use cases/edge cases, maybe we should add a test or two in here to ensure we're able to read and write these largePDB files?

jeremyleung521:
> So we don't keep finding more and more use cases/edge cases, maybe we should add a test or two in here to ensure we're able to read and write these largePDB files?

Isn't that what the [mdconvert tests](https://github.com/mdtraj/mdtraj/blob/master/tests/test_mdconvert.py#L113) are doing? Though maybe the test files used either doesn't have enough residues and/or overflow # of atoms to actually catch any of these errors?

Also regarding testing read/write large files, we have to balance the fact that test reference files bloat up the repo (we have about 70MB of [test data](https://github.com/mdtraj/mdtraj/tree/master/tests/data) already), which is why I opted for the [condensed files](https://github.com/jeremyleung521/mdtraj/blob/mdtraj-10000-bugfix/tests/data/vmd_indexing.pdb), which should now catch the false positives. I added the two tests last time (and [updated them this time](https://github.com/jeremyleung521/mdtraj/blob/mdtraj-10000-bugfix/tests/test_pdb.py#L348-L370)). VMD residue number is an special case because the first overflow number is also a number readable in base 10 (2710 in base-16 is 10000 in base-10). In the end, this is a loading/reading error that manifested as a write error symptom (and the writing is already sufficiently tested in mdconvert tests), so checking that it's reading correctly should be fine.

The TER bug's apparently been around since 5+ years ago, so I was just fixing it on the way. I don't think it actually matters too much anyways.

gph82:
Hi! 

I hope is the right place, but while this is open, an FYI that both the current release 1.10.10 (installed via pip) and the master branch at commit ccd0c874361e4e31127e3803d3670d08d57f7933 (which, BTW, is still tagged at as 1.9.10.dev0 [here](https://github.com/mdtraj/mdtraj/blame/ccd0c874361e4e31127e3803d3670d08d57f7933/setup.py#L36)?) stopped reading pdbs from the [OMP (Orientations of Proteins in Membranes Database)](https://opm.phar.umich.edu/proteins/419) giving an error similar to https://github.com/mdtraj/mdtraj/pull/1849#issuecomment-2182757651. 

Since previous releases could read them, I did a `git bisect` which points to 3d5fa756593a57ca90c0ae75d35d98c088172b42 (Allow mdtraj.load() to read various non-standard spec PDB files #1849) as the first bad commit. 

So I followed that commit to issue #1849, which then points to #1903, which itself proposes #1904 (here). 

Letting you know that when I check out [jeremyleung521:mdtraj-10000-bugfix](https://github.com/jeremyleung521/mdtraj/tree/mdtraj-10000-bugfix) then the OPM file is readable again

jeremyleung521:
> Hi!
> 
> I hope is the right place, but while this is open, an FYI that both the current release 1.10.10 (installed via pip) and the master branch at commit [ccd0c87](https://github.com/mdtraj/mdtraj/commit/ccd0c874361e4e31127e3803d3670d08d57f7933) (which, BTW, is still tagged as 1.9.10.dev0 [here](https://github.com/mdtraj/mdtraj/blame/ccd0c874361e4e31127e3803d3670d08d57f7933/setup.py#L36)?) stopped reading pdbs from the [OMP (Orientations of Proteins in Membranes Database)](https://opm.phar.umich.edu/proteins/419) giving an error similar to [#1849 (comment)](https://github.com/mdtraj/mdtraj/pull/1849#issuecomment-2182757651).
> 
> Since previous releases could read them, I did a `git bisect` which points to [3d5fa75](https://github.com/mdtraj/mdtraj/commit/3d5fa756593a57ca90c0ae75d35d98c088172b42) (Allow mdtraj.load() to read various non-standard spec PDB files #1849) as the first bad commit.
> 
> So I followed that commit to issue #1849, which then points to #1903, which itself proposes #1904 (here).
> 
> Letting you know that when I check out [jeremyleung521:mdtraj-10000-bugfix](https://github.com/jeremyleung521/mdtraj/tree/mdtraj-10000-bugfix) then the OPM file is readable again

Thanks for testing (and making sure this works), @gph82 ! 

@mattwthompson Since this appears to be causing a lot of problems everywhere, it would be great of this (and the netcdf/scipy bug) could be merged and be pushed as part of a new release soon (along with the version number update).

pbuslaev:
Hi @jeremyleung521, thanks for fixing. One thing that I should mention, is that I am having quite a hard time reading this code. Do you mind if I try to change variable/method names and add some comments so that the logics you've implemented is a bit clearer?

jeremyleung521:
> Hi @jeremyleung521, thanks for fixing. One thing that I should mention, is that I am having quite a hard time reading this code. Do you mind if I try to change variable/method names and add some comments so that the logics you've implemented is a bit clearer?

Hopefully that's enough. I also removed some of the legacy args/logic from #1675 (like ``indexing_fncs``) that I never bothered to delete when I started #1849.

mattwthompson:
I think pushing out a 1.10.1 release sooner than later is a good idea once this is reviewed and merged

jeremyleung521:
> I think pushing out a 1.10.1 release sooner than later is a good idea once this is reviewed and merged

@mattwthompson This PR is ready to merge too. Obviously depends on when we want 1.10.1 to be released.

mattwthompson:
I think this warrants a review

1.10.1 might happen next week?

Rasput1n00:
Hi!

I hope this is the right place to write. I'm currently using the "mdconvert" function to output several frames from dcd files into pdb files(output). As an input topology file, my PDB input file has residue sequence number larger than 10000.
When I use your "mdtraj-mdtraj-10000-bugfix" version([924c2eb](https://github.com/jeremyleung521/mdtraj/commit/924c2ebcd9779ac8d63457b7f29af982c45598d8)) to output some PDB files, there are still several mistakes in the output results, as shown in Fig.
image

So, the problem is that the atom ID just changed due to the mistake lines. 

It would be much appreciated if you could fix this bug, and thanks for your time!

Rasput1n00:
> Hi!
> 
> I hope this is the right place to write. I'm currently using the "mdconvert" function to output several frames from dcd files into pdb files(output). As an input topology file, my PDB input file has residue sequence number larger than 10000. When I use your "mdtraj-mdtraj-10000-bugfix" version([924c2eb](https://github.com/jeremyleung521/mdtraj/commit/924c2ebcd9779ac8d63457b7f29af982c45598d8)) to output some PDB files, there are still several mistakes in the output results, as shown in Fig. image
> 
> So, the problem is that the atom ID just changed due to the mistake lines.
> 
> It would be much appreciated if you could fix this bug, and thanks for your time!

I'm sorry that I haven't used your newest commit version(6c9543d23d7a2c665108b4bcf6b2b82999f3b799), but I ran some tests on your newest commit version, such a situation remains.

jeremyleung521:
> Hi!
> 
> I hope this is the right place to write. I'm currently using the "mdconvert" function to output several frames from dcd files into pdb files(output). As an input topology file, my PDB input file has residue sequence number larger than 10000. When I use your "mdtraj-mdtraj-10000-bugfix" version([924c2eb](https://github.com/jeremyleung521/mdtraj/commit/924c2ebcd9779ac8d63457b7f29af982c45598d8)) to output some PDB files, there are still several mistakes in the output results, as shown in Fig. image
> 
> So, the problem is that the atom ID just changed due to the mistake lines.
> 
> It would be much appreciated if you could fix this bug, and thanks for your time!

Thanks for pointing this out.

One thing is that the input PDB file is not up to spec. it's overflowing the residue number into chain column, which is why it's reading it as chain 1, residue 127. instead of residue 10127. You're also jumping around a lot in the residue number which will not trigger the overflow. 

The TER line I haven't changed, so it seems like it's a bug from a long while back (does it behave the same in 1.9.9?). I tried commenting out [L414](https://github.com/mdtraj/mdtraj/pull/1904/files#diff-8ac1d0f3c3130835a8c75b05097cf0e34929ed1cf6c43366d66ab6c3b7cb8e45R414) but `test_multichain_load_cycle`failed. 

I don't have time to fix this until later in August so if someone has time to do it before that, please do.

mattwthompson:
Could we get somebody with more expertise than me with PDB files to review this?

Rasput1n00:
> > Hi!
> > I hope this is the right place to write. I'm currently using the "mdconvert" function to output several frames from dcd files into pdb files(output). As an input topology file, my PDB input file has residue sequence number larger than 10000. When I use your "mdtraj-mdtraj-10000-bugfix" version([924c2eb](https://github.com/jeremyleung521/mdtraj/commit/924c2ebcd9779ac8d63457b7f29af982c45598d8)) to output some PDB files, there are still several mistakes in the output results, as shown in Fig. image
> > So, the problem is that the atom ID just changed due to the mistake lines.
> > It would be much appreciated if you could fix this bug, and thanks for your time!
> 
> Thanks for pointing this out.
> 
> One thing is that the input PDB file is not up to spec. it's overflowing the residue number into chain column, which is why it's reading it as chain 1, residue 127. instead of residue 10127. You're also jumping around a lot in the residue number which will not trigger the overflow.
> 
> The TER line I haven't changed, so it seems like it's a bug from a long while back (does it behave the same in 1.9.9?). I tried commenting out [L414](https://github.com/mdtraj/mdtraj/pull/1904/files#diff-8ac1d0f3c3130835a8c75b05097cf0e34929ed1cf6c43366d66ab6c3b7cb8e45R414) but `test_multichain_load_cycle`failed.
> 
> I don't have time to fix this until later in August so if someone has time to do it before that, please do.

I knew the residue sequence number is overflowing, but it’s a compromised choice. The input PDB file originally generated by VMD “topo readlammpsdata” function, in which the residue sequence number is wrote in decimal  when it <10000, and in hex. when it >=10000.  And “mdconvert” function cannot read such a hex. number without any marker. So I have to change the hex. seq. number into dec.

I will try to sort my input PDB file in residue sequence number to solve the problem bothering me. Thank you so much for your reply!

jeremyleung521:
> > > Hi!
> > > I hope this is the right place to write. I'm currently using the "mdconvert" function to output several frames from dcd files into pdb files(output). As an input topology file, my PDB input file has residue sequence number larger than 10000. When I use your "mdtraj-mdtraj-10000-bugfix" version([924c2eb](https://github.com/jeremyleung521/mdtraj/commit/924c2ebcd9779ac8d63457b7f29af982c45598d8)) to output some PDB files, there are still several mistakes in the output results, as shown in Fig. image
> > > So, the problem is that the atom ID just changed due to the mistake lines.
> > > It would be much appreciated if you could fix this bug, and thanks for your time!
> > 
> > 
> > Thanks for pointing this out.
> > One thing is that the input PDB file is not up to spec. it's overflowing the residue number into chain column, which is why it's reading it as chain 1, residue 127. instead of residue 10127. You're also jumping around a lot in the residue number which will not trigger the overflow.
> > The TER line I haven't changed, so it seems like it's a bug from a long while back (does it behave the same in 1.9.9?). I tried commenting out [L414](https://github.com/mdtraj/mdtraj/pull/1904/files#diff-8ac1d0f3c3130835a8c75b05097cf0e34929ed1cf6c43366d66ab6c3b7cb8e45R414) but `test_multichain_load_cycle`failed.
> > I don't have time to fix this until later in August so if someone has time to do it before that, please do.
> 
> I knew the residue sequence number is overflowing, but it’s a compromised choice. The input PDB file originally generated by VMD “topo readlammpsdata” function, in which the residue sequence number is wrote in decimal when it <10000, and in hex. when it >=10000. And “mdconvert” function cannot read such a hex. number without any marker. So I have to change the hex. seq. number into dec.
> 
> I will try to sort my input PDB file in residue sequence number to solve the problem bothering me. Thank you so much for your reply!

So I went down the rabbit hole and dug through #1611 #1612 #1739, which were all about choosing an intuitive behavior for atom numbering for pdb. I added in the `ter=False`option of so when you run `traj.save('blah.pdb', ter=False)` or any pass `ter=False` to any equivalent function, it will no longer include the TER lines, match that in CONECT section, which should hopefully allow you to maintain your atom number order.

It will be difficult to do anything about atom numbers/residue numbers out of order. Anything else probably require a behavioral change and we would probably need to form some sort of consensus in the community.

jeremyleung521:
@mattwthompson Is there anything I can do to speed this PR (and thus 1.10.1 release) along? Are we just waiting for some fresh eyes to review?

mattwthompson:
I don't think so - at a certain point I wonder if merging as-is and taking a (small) risk of another unexpected quirk is less harmful than delaying this for months

Issue #1907, opened 2024-07-05
six module removed but still referenced, breaking mdinspect


drew-parsons:
mdtraj 1.10.0 removed `utils/six.py`.  But it is still referenced by the geometry submodule at

https://github.com/mdtraj/mdtraj/blob/ccd0c874361e4e31127e3803d3670d08d57f7933/mdtraj/geometry/internal.py#L37

`geometry/internal.py` is almost not used at all, except that it is used by `scripts/mdinspect.py`

Hence the mdinspect script does not work with mdtraj 1.10.0

```
$ mdinspect 
Traceback (most recent call last):
  File "/usr/bin/mdinspect", line 33, in <module>
    sys.exit(load_entry_point('mdtraj==1.9.10.dev0', 'console_scripts', 'mdinspect')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/bin/mdinspect", line 25, in importlib_load_entry_point
    return next(matches).load()
           ^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/importlib/metadata/__init__.py", line 205, in load
    module = import_module(match.group('module'))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 995, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "/usr/lib/python3/dist-packages/mdtraj/scripts/mdinspect.py", line 38, in <module>
    from mdtraj.geometry.internal import COVALENT_RADII
  File "/usr/lib/python3/dist-packages/mdtraj/geometry/internal.py", line 37, in <module>
[...]

PR #1908, opened 2024-07-05
remove remaining instances of six

jeremyleung521:
#1907 

Since it's an easy fix, I've gone ahead and made them.

`filter` in python 3 now returns a `generator` instead of a `list`, but shouldn't matter in since we're using it in a `for` loop.
`xrange` is only used in `for` loops, so a `range` should work the same.

Can confirm the error doesn't show up, but I haven't used mdinspect before.


mattwthompson (approved):
LGTM, I wonder why this was missed? Are we just skipping the tests that would hit this file?

jeremyleung521:
> LGTM, I wonder why this was missed? Are we just skipping the tests that would hit this file?

mdinspect appears to not have any tests... And none of the tests touch the `get_***_connectivity()` functions in geometry/internal.py

Issue #1896, opened 2024-06-04
Set up better versioning scheme


mattwthompson:
https://pypi.org/project/versioneer/

https://pypi.org/project/versioningit/

mattwthompson:
MDTraj now uses versioneer as of #1909. Let us know if there are any issues

PR #1909, opened 2024-07-09
Set up versioneer

mattwthompson:
tag https://github.com/mdtraj/mdtraj/issues/1896

`versioningit` doesn't play nicely with what we have in `setup.py` so I went with `versioneer`

This is mostly ~automated,

```shell
$ pip install versioneer
$ versioneer install --no-vendor
```

and then fiddle with the remainder


setup.py
```diff
@@ -32,13 +32,6 @@
     disable_openmp = False
 
 
-##########################
-VERSION = "1.9.10.dev0"
-ISRELEASED = False
-__version__ = VERSION
-##########################
-
```
mattwthompson, line 40: IMHO the main value add is having something less error-prone than this; in the release wheels I have this set (properly) to 1.10.0 but I had to do that with a local modification that differs from the release assets, which is not good.

mattwthompson:
I've put no effort into fixing the import loop - happy to have somebody else look at this before me

jeremyleung521:
> I've put no effort into fixing the import loop - happy to have somebody else look at this before me

@mattwthompson Please pull my branch.  I fixed it here: https://github.com/jeremyleung521/mdtraj/tree/versioneer

Some file formats needed the version number to stamp the files, but versions (with versioneer) should be called via `mdtraj.__version__` instead of `mdtraj.version.version` or `mdtraj.version.short_version` using `versioningit`

Example working tests here: https://github.com/jeremyleung521/mdtraj/pull/3

mattwthompson:
Seems good to me, thanks @jeremyleung521 

For any reviewers, `mdtraj/_version.py` is roughly auto-generated from the commands above

jeremyleung521 (commented):
I think these need to be uncommented and replaced with the following. I think the [sphinx backend](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-version) pulls version number from here for the html and such:

```
import mdtraj

version = mdtraj.__version__
release = mdtraj.__version__
```

mattwthompson:
Good catch, I saw the local variable wasn't used and didn't dig deeper

Issue #1818, opened 2023-09-06
xrdlib deprecation warning


richardjgowers:
I'm getting this warning in a test suite:  `/Users/runner/micromamba-root/envs/openfe_env/lib/python3.11/site-packages/mdtraj/formats/__init__.py:6: DeprecationWarning: 'xdrlib' is deprecated and slated for removal in Python 3.13`

Over in mda land we made this as a drop in replacement:

https://pypi.org/project/mda-xdrlib/

mattwthompson:
Those imports do different things, one for the Python layer and one for the C layer (note it's a `.pyx` file, not a normal Python file). I also think that looks a namespace collision, but that [appears to be](https://stackoverflow.com/a/20268478) how it's done in Cython world

padix-key:
That is true, having it both imported and cimported simply makes attributes from Python and C world available. However, as far as I saw in this case, `xdrlib` is only used to call functions/types from [`xdrlib.pxd`](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/xtc/xdrlib.pxd) and nothing from the Python standard library module `xdrlib.py` is called.

mattwthompson:
Maybe the Python import can be skipped, I'm not sure having almost no Cython experience myself. I'd happily review a PR that gives it a shot

PR #1911, opened 2024-07-13
Replace `xdrlib` with `mda_xdrlib`

chrisjonesBSU:
I'm trying to reduce some of the deprecation warnings we get in the mosdef packages, and one of them was from mdtraj's use of `xdrlib`.  https://github.com/mdtraj/mdtraj/issues/1818 mentions using the `mda_xdrlib` package as a replacement, which this PR does.

### DIFF
--- a/devtools/conda-envs/test.yaml
+++ b/devtools/conda-envs/test.yaml
@@ -8,6 +8,7 @@ dependencies:
   - numpy
   - zlib
   - pyparsing
+  - mda-xdrlib
   - scipy
   - networkx
   - pandas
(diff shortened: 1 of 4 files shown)


jeremyleung521:
@chrisjonesBSU Seems like you need to add `mda-xdrlib` into https://github.com/mdtraj/mdtraj/blob/master/docs/requirements.txt as well.

mattwthompson (approved):
Okay so it actually is a drop-in replacement? Nice

I assume the code hit by this is tested, otherwise it's not likely to be throwing the deprecation warnings we've seen for ages

mattwthompson:
Let's make sure to add https://prefix.dev/channels/conda-forge/packages/mda-xdrlib to the conda builds in the next release

PR #1917, opened 2024-08-20
Fix RMSD externs to be compatible with Cython 3.0.

Augustin-Zidek:
* We needed this fix internally to make mdtraj work with Cython 3.0.
* This fixed it, however, we build rdkit ourselves using Blaze (Bazel), so this fix might not be applicable upstream.

### DIFF
--- a/devtools/conda-envs/test.yaml
+++ b/devtools/conda-envs/test.yaml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   # Core
   - python
-  - cython =0.29.36
+  - cython
   - numpy
   - zlib
   - pyparsing
--- a/devtools/conda-envs/test.yaml
+++ b/devtools/conda-envs/test.yaml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   # Core
   - python
-  - cython
+  - cython =3
   - numpy
   - zlib
   - pyparsing
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ requires = [
     "versioneer[toml]",
     "wheel",
     "numpy~=1.25",
-    "Cython~=3",
+    "Cython>=3.0.0",
 ]
 
 [tool.versioneer]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ requires = [
     "versioneer[toml]",
     "wheel",
     "numpy~=1.25",
-    "Cython>0.29",
+    "Cython=3",
 ]
 
 [tool.versioneer]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ requires = [
     "versioneer[toml]",
     "wheel",
     "numpy~=1.25",
-    "Cython~=0.29.36",
+    "Cython>0.29",
 ]
 
 [tool.versioneer]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ requires = [
     "versioneer[toml]",
     "wheel",
     "numpy~=1.25",
-    "Cython=3",
+    "Cython~3",
 ]
 
 [tool.versioneer]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ requires = [
     "versioneer[toml]",
     "wheel",
     "numpy~=1.25",
-    "Cython~3",
+    "Cython~=3",
 ]
 
 [tool.versioneer]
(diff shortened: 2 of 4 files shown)


mattwthompson:
Are these changes backwards-compatible?

Augustin-Zidek:
Yes, I tested backwards compatibility with Cython 0.29 in our environment ad it works.

Also, another data point to support backwards compatibility is that some of the externs are already wrapped in the `cdef extern from`.

mattwthompson:
Well this seems like an improvement that stands on its own if it works with 3.0; tests pass with the current environment (notably not on 3.0) but that doesn't necessarily mean it works with 3.0. Testing it with 3.0 would require updating all of the Cython code, though, and that's not something that can be taken on too easily I don't think.

I'm not sure what to do here - what do others think?

Augustin-Zidek:
Note that with this change and one further fix in `image_molecules.pxi` (which I will send a PR for once I internally rebase on top of the latest mdtraj version), I am able to build and have all mdtraj tests green under Cython 3.0.

Augustin-Zidek:
Update: with the latest MDTraj version, this PR is all that is needed to get it to build with Cython 3.0 (or at least in our setup), since the issue in `image_molecules.pxi` has been fixed in https://github.com/mdtraj/mdtraj/commit/f17266a1bbd5f659549ca03ff0c3dcea7535e362.

mattwthompson:
Care to bump the Cython version here?

Augustin-Zidek:
Could I do that in a separate PR? (To signify that this change is backwards-compatible.)

mattwthompson:
I'd rather not maintain backwards compatibility, or maintain two versions of this upstream, personally

mattwthompson:
@Augustin-Zidek AFAICT, these changes continue to work on the old Cython. Do you know how much work is associated with getting 3.0 working? Updating this line of the test environment would get CI working on Cython 3.0: https://github.com/mdtraj/mdtraj/blob/master/devtools/conda-envs/test.yaml#L7

Augustin-Zidek:
Sorry for the long delay, this is now done.

* In `devtools/conda-envs/test.yaml` I removed any version constraints like it was before https://github.com/mdtraj/mdtraj/commit/2ba38d830b89c33da37e291582ad289abf3a3ff4
* In `pyproject.toml` I changed `Cython~=0.29.36` to `Cython>0.29` which is similar to the `Cython>0.28` that used to be there before.

mattwthompson:
No apologies necessary! We appreciate your work!

pyproject.toml
```diff
@@ -4,7 +4,7 @@ requires = [
     "versioneer[toml]",
     "wheel",
     "numpy~=1.25",
-    "Cython~=0.29.36",
+    "Cython>0.29",
```
mattwthompson, line 7: I think this should be 
```suggestion
    "Cython=3",
```
?

devtools/conda-envs/test.yaml
```diff
@@ -4,7 +4,7 @@ channels:
 dependencies:
   # Core
   - python
-  - cython =0.29.36
+  - cython
```
mattwthompson, line 7: I checked the logs and 3.x is pulled down, but just to be safe (and explicit about dependencies)
```suggestion
  - cython =3
```

pyproject.toml
Augustin-Zidek: Done.

devtools/conda-envs/test.yaml
Augustin-Zidek: Done.

Augustin-Zidek: Ah, this actually doesn't work and it should be `~3`.

pyproject.toml
Augustin-Zidek: Unfortunately, this didn't work. 

I tried multiple things, only `Cython>=3.0.0` worked.

mattwthompson: Weird, I find TOML formatting to be a headache. If it works it works 🤷

mattwthompson:
I remain surprised that the codebase now passes with Cython 3.0 - I thought it was going to be a huge, messy refactor - but I checked the logs and that's what's installed.

IIUC, the change in `pyproject.toml` will handle the dependency update for all thing that pull from PyPI.

My proposed path forward here is relatively simple:
1. Merge this PR
2. (Once other PRs are merged or deferred) cut #1926
3. Ensure `cython =3` is in conda dependencies

What do we think? @jeremyleung521 @janash @Lnaden @sukritsingh

Augustin-Zidek:
Looks like @jeremyleung521 and @janash are happy -- thanks for taking a look. Shall we land this? (I need approval from your side.)

PR #1918, opened 2024-08-21
Add option for HDF5 openmm reporter to save forces

andychenbruce:
I didn't see any way in the [documentation](https://mdtraj.org/development/api/generated/mdtraj.reporters.HDF5Reporter.html) to save the `forces`, which I wanted as training data for a neural net.

I tested the reporter and it seems to work:
```
simulation.reporters.append(mdtraj.reporters.HDF5Reporter('ala2.h5', 1000, forces=True))
```

I'm not sure how the unit conversions are intended to work. I set the units for the forces to be the same as OpenMM which [says](http://docs.openmm.org/latest/userguide/theory/01_introduction.html):
```
For example, a force always has the same units (kJ/mol/nm)
```

### DIFF
--- a/mdtraj/reporters/basereporter.py
+++ b/mdtraj/reporters/basereporter.py
@@ -62,6 +62,7 @@ class _BaseReporter:
         velocities=False,
         atomSubset=None,
         enforcePeriodicBox=None,
+        forces=False
     ):
         """Create an OpenMM reporter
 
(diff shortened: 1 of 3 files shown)


mdtraj/reporters/basereporter.py
```diff
@@ -62,6 +62,7 @@ def __init__(
         velocities=False,
         atomSubset=None,
         enforcePeriodicBox=None,
+        forces=False
```
mattwthompson, line 65: If no other reporters store forces, could this just be changed in the child class?

mattwthompson (changes requested):
These passes the eye test, but needs some unit tests added

PR #1919, opened 2024-08-23
Remove unused Python xdrlib import

### LINKED ISSUE #1818: xrdlib deprecation warning

padix-key:
Following up the discussion of #1818, this PR removes the unused Python `xdrlib`, as only the internal C `xdrlib` is used. Closes #1818

### DIFF
--- a/devtools/conda-envs/test.yaml
+++ b/devtools/conda-envs/test.yaml
@@ -8,7 +8,6 @@ dependencies:
   - numpy
   - zlib
   - pyparsing
-  - mda-xdrlib
   - scipy
   - networkx
   - pandas
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -7,7 +7,6 @@ nbsphinx_link
 msmb_theme
 jupyter
 ipython
-mda-xdrlib
 matplotlib
 pandoc
 jinja2
--- a/mdtraj/formats/xtc/trr.pyx
+++ b/mdtraj/formats/xtc/trr.pyx
@@ -28,7 +28,6 @@
 
 import os
 import warnings
-import mda_xdrlib as xdrlib
 
 import numpy as np
 
--- a/mdtraj/formats/xtc/xtc.pyx
+++ b/mdtraj/formats/xtc/xtc.pyx
@@ -27,7 +27,6 @@
 ###############################################################################
 import os
 import warnings
-import mda_xdrlib as xdrlib
 
 import numpy as np
 
--- a/setup.py
+++ b/setup.py
@@ -302,7 +302,6 @@ metadata = dict(
         "scipy",
         "pyparsing",
         "packaging",
-        "mda-xdrlib",
     ],
     package_data={"mdtraj.formats.pdb": ["data/*"]},
     zip_safe=False,


mattwthompson:
Unfortunately we can't use the version that's in the standard library because it's being removed in 3.13

https://docs.python.org/3/library/xdrlib.html

mattwthompson:
We switched over to MDAnalysis's version in #1911, which should land in a release once #1904 is finished up

jeremyleung521:
Because we're using xdrlib in the "python" context and also "Cython" context, we need to load both. Former for lines like [these](https://github.com/mdtraj/mdtraj/blob/6d77f4eeb398cee2b14460f9ffbc1f263f26c6b7/mdtraj/formats/xtc/trr.pyx#L273), latter for the header files for lines like [these](https://github.com/mdtraj/mdtraj/blob/6d77f4eeb398cee2b14460f9ffbc1f263f26c6b7/mdtraj/formats/xtc/trr.pyx#L560). Convention is to load them [under the same name](https://numpy.org/doc/stable/reference/random/extending.html#cython).


Easy way to test is:
In a python 3.13 environment (`conda create -n 313 -c conda-forge/label/python_rc python=3.13.0rc1 mda-xdrlib`), install cython that is compatible with 3.13 (`conda activate 313; python -m pip install cython`)

In a test file (`test.pyx`) containing the following:
```
import mda_xdrlib as xdrlib
cimport mda_xdrlib as xdrlib

xdrlib.xdrtel('abc.txt')
```

Compilation will fail (`cythonize -i test.pyx`) if the first line is commented out.


What we need to do is though, also modify it so the cimport also pulls from `mda_xdrlib`. Will be pushing a PR shortly for that.

padix-key:
> Former for lines like [these](https://github.com/mdtraj/mdtraj/blob/6d77f4eeb398cee2b14460f9ffbc1f263f26c6b7/mdtraj/formats/xtc/trr.pyx#L273)

Specifially these are taken from the shipped C library, I think (https://github.com/mdtraj/mdtraj/blob/6d77f4eeb398cee2b14460f9ffbc1f263f26c6b7/mdtraj/formats/xtc/include/xdr_seek.h#L13). I also tested removing the `import xdrlib` line (back when `mdtraj` used the one form the Python stdlib) and it still worked fine

jeremyleung521:
> > Former for lines like [these](https://github.com/mdtraj/mdtraj/blob/6d77f4eeb398cee2b14460f9ffbc1f263f26c6b7/mdtraj/formats/xtc/trr.pyx#L273)
> 
> Specifially these are taken from the shipped C library, I think
> 
> https://github.com/mdtraj/mdtraj/blob/6d77f4eeb398cee2b14460f9ffbc1f263f26c6b7/mdtraj/formats/xtc/include/xdr_seek.h#L13
> I also tested removing the `import xdrlib` line (back when `mdtraj` used the one form the Python stdlib) and it still worked fine

Ah, I missed that. Also I didn't see that we actually bundled the header files in mdtraj and so the cython parts we've been using were never part of mda-xdrlib (or python stdlib). 

In that case, then yea this LGTM. Thanks for catching this :) One less dependency, for sure.

mattwthompson:
So MDTraj bundles this library? Where in the source tree is it?

jeremyleung521:
> So MDTraj bundles this library? Where in the source tree is it?

These are the header files for the cython... 
https://github.com/mdtraj/mdtraj/tree/master/mdtraj/formats/xtc/include

padix-key found out that we don't really use the regular "python" imported portion, which is vendored by `mda_xdrlib`.

mattwthompson:
Thanks for the explanation.

I looked at these logs yesterday and didn't see the deprecation warnings that would indicate problems when 3.13 launches.

mattwthompson (approved):

PR #1923, opened 2024-08-30
Numpy 2 Support

jeremyleung521:
So I began working on Numpy 2 support because (surprise) it doesn't seem like there will be a Numpy v1.27 for Python 3.13.

The biggest headache right now is [NEP50](https://numpy.org/neps/nep-0050-scalar-promotion.html), where dtype will no longer dynamically adjust based on the magnitude of the values. There are a lot of examples in the NEP50 link, but effectively, you can very easily get RuntimeError (or rounding errors) just from basic math operations because the output value will be of the wrong type, overflowing or rounding numbers to the wrong values. (https://numpy.org/neps/nep-0050-scalar-promotion.html#backward-compatibility)

Currently there are 7 failed tests and it's a wack-a-mole trying to figure out which basic math operation is causing the wrong output/dtype. I suspect there are a lot of other hidden cases too because the whole unit conversion is a rabbit hole + a huge mess because of all the basic math going on.

Example run with errors:
https://github.com/jeremyleung521/mdtraj/actions/runs/10622593572/job/29447235437

Would be great if someone could help!


devtools/conda-envs/test.yaml
```diff
@@ -5,7 +5,7 @@ dependencies:
   # Core
   - python
   - cython =0.29.36
-  - numpy
+  - numpy>=2
```
mattwthompson, line 8:
```suggestion
  - numpy=2
```
I doubt v3 will drop anytime soon, but we probably wouldn't want that pulled in 😅

mattwthompson (commented):
You should drop AmberTools from the testing environment (and unfortunately skip tests that need it, if that's not already in place) as it's not going to be getting new builds anytime soon

jeremyleung521:
Note to self and/or others: `tests/__init__.py` activates warnings (e.g. [this](https://github.com/mdtraj/mdtraj/actions/runs/10688152873/job/29627382023?pr=1923#step:6:2625)) every time the output dtype has changed due to NEP 50. This should be deleted after everything is fixed.

devtools/conda-envs/test.yaml
```diff
@@ -5,7 +5,7 @@ dependencies:
   # Core
   - python
   - cython =0.29.36
-  - numpy
+  - numpy=2
```
mattwthompson, line 8: It'd be nice to test on version 1 and 2, injecting it in CI like we already do Python versions

jeremyleung521:
See #1928 instead.

devtools/conda-envs/test.yaml
jeremyleung521: Added in #1928.

PR #1928, opened 2024-09-18
Numpy 2 Support (New)

jeremyleung521:
This PR replaces #1923 so it's easier to collaborate.

------

Original Post:

So I began working on Numpy 2 support because (surprise) it doesn't seem like there will be a Numpy v1.27 for Python 3.13.

The biggest headache right now is [NEP50](https://numpy.org/neps/nep-0050-scalar-promotion.html), where dtype will no longer dynamically adjust based on the magnitude of the values. There are a lot of examples in the NEP50 link, but effectively, you can very easily get RuntimeError (or rounding errors) just from basic math operations because the output value will be of the wrong type, overflowing or rounding numbers to the wrong values. (https://numpy.org/neps/nep-0050-scalar-promotion.html#backward-compatibility)

Currently there are 7 failed tests and it's a wack-a-mole trying to figure out which basic math operation is causing the wrong output/dtype. I suspect there are a lot of other hidden cases too because the whole unit conversion is a rabbit hole + a huge mess because of all the basic math going on.

Example run with errors:
https://github.com/jeremyleung521/mdtraj/actions/runs/10622593572/job/29447235437

Would be great if someone could help!

---
Note to self and/or others: ``tests/__init__.py`` activates warnings (e.g. [this](https://github.com/mdtraj/mdtraj/actions/runs/10688152873/job/29627382023?pr=1923#step:6:2625)) every time the output dtype has changed due to NEP 50. This file should be deleted after everything is fixed.


mdtraj/formats/pdb/pdbfile.py
```diff
@@ -201,7 +201,7 @@ def load_pdb(
         # cell is not absurdly high. Standard water density is ~55 M, which
         # yields a particle density ~100 atoms per cubic nm. It should be safe
         # to say that no particle density should exceed 10x that.
-        particle_density = traj.top.n_atoms / traj.unitcell_volumes[0]
+        particle_density = np.float64(traj.top.n_atoms) / np.float64(traj.unitcell_volumes[0])
```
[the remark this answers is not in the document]
jeremyleung521: unitcell_lengths and unitcell angles are often inherited from the exact file you load in so the volume calculated is often in <float64, especially from a pdb.  I did try to fix it [upstream](https://github.com/mdtraj/mdtraj/pull/1928/files/c2a5d162b00a5f084b86abd63ed95d0f453f99a3#diff-1b104b5984a85d8b849914ff10c636e498d9b75091dc845b9595a740f6580503R877) during volume calculation, but it didn't seem to be enough to fix the test. 

Also since I apparently broke the tests on 1.26, I should probably begin rolling back on some of my changes just so it's at an easier place for other people to start.

mdtraj/utils/unit/mymatrix.py
```diff
@@ -235,7 +235,7 @@ def __mul__(self, rhs):
             for j in range(n):
                 for k in range(r):
                     result[i][j] += self[i][k] * rhs[k][j]
-        return result
+        return np.float64(result)
```
janash, line 238: I have so many thoughts about this module. 

Does it make sense to cast the result as np.float64 when the calculations weren't necessarily done with `np.float64` in the first place?

More broadly, does it make sense to have a **pure Python** linear algebra matrix when NumPy is already a dependency? Should we be coverting to and working with NumPy arrays here instead of looping over lists?

janash: I don't see numpy imported in this module. Is it coming from somewhere else?

jeremyleung521:
Thanks for looking at this, @janash.

So I just found out that the 1.26* tests are failing because I forced Numpy2 behavior when I changed the NPY_PROMOTION_STATE to 'weak_and_warn'. (https://github.com/mdtraj/mdtraj/pull/1928/commits/d4f1a5cfdf7e36f7815e5750ae85b529b58149ca)

Now that I know this, we could force 'legacy' behavior in numpy 2 by running ``np._set_promotion_state('legacy')`` on a global scale, which would allow us to bring numpy 2 support quickly, so we can take the time to actually implement all these fixes.

Lnaden:
> we could force 'legacy' behavior in numpy 2 by running np._set_promotion_state('legacy') on a global scale, which would allow us to bring numpy 2 support quickly, so we can take the time to actually implement all these fixes.

If you do that, you will need to add a warning so that anyone else who is already on Numpy 2 support and uses MDTraj knows this is happening since its a global setting and not just an alternate import path.

janash:
> Thanks for looking at this, @janash.
> 
> So I just found out that the 1.26* tests are failing because I forced Numpy2 behavior when I changed the NPY_PROMOTION_STATE to 'weak_and_warn'. ([d4f1a5c](https://github.com/mdtraj/mdtraj/commit/d4f1a5cfdf7e36f7815e5750ae85b529b58149ca))
> 
> Now that I know this, we could force 'legacy' behavior in numpy 2 by running `np._set_promotion_state('legacy')` on a global scale, which would allow us to bring numpy 2 support quickly, so we can take the time to actually implement all these fixes.
> 
> (Here's a PR of a clean branch to demonstrate this: [jeremyleung521#6](https://github.com/jeremyleung521/mdtraj/pull/6))

This is seems convenient for the moment and conserving expected behavior! I agree that we'd want to think carefully about real fixes and compatibility with NumPy 2.0 in the long run if we took this approach.

mdtraj/formats/pdb/pdbfile.py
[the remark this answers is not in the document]
janash: Yes, right. Looking at it, I do see that the `unitcell_lengths` and angles are coming from the format readers. Thanks for pointing that out! It looks like the (or formats or whatever I should be calling them) are making instances of trajectories (let me know if that isn't always the case, I saw Amber and LAMMPS do this). For the particular case of unitcell vectors and angles, it seems to me like we could cast to `np.float64` in the `Trajectory` object for things like unit cell by using a `property` and a setter to make sure unitcells are always `np.float64`.

In some cases, it looks like data types are already "hard-coded", e.g. LAMMPS explicitly sets its lengths to `np.float32`.

https://github.com/mdtraj/mdtraj/blob/036e0d5f514d7e0344a3d7d90248c0c1e3384673/mdtraj/formats/lammpstrj.py#L313

Just thinking through things, we could take the approach of property + setter in `Trajectory` to make sure `unitcell_lengths` etc are in `np.float64` for trajectory objects, or we could go into the code for each format and make sure they are using `np.float64`. I guess I would prefer the first option unless there is a need to change data types within the specific format objects (seems like we'd have to carefully review what is going on in each format reader).

I'd be curious to know why your solution of casting during the volume calculation didn't work :) Maybe catching it even further upstream (such as enforcing the type of unitcell_lengths would help, though we'd have to evaluate if there were any implications)

jeremyleung521:
> > we could force 'legacy' behavior in numpy 2 by running np._set_promotion_state('legacy') on a global scale, which would allow us to bring numpy 2 support quickly, so we can take the time to actually implement all these fixes.
> 
> If you do that, you will need to add a warning so that anyone else who is already on Numpy 2 support and uses MDTraj knows this is happening since its a global setting and not just an alternate import path.

Good point. Other than that, I do think most (if not all) of the NEP50 changes are backwards compatible (we're being more explicit/verbose with the casting after all). 

So I tested [this](https://github.com/jeremyleung521/mdtraj/actions/runs/11076119695/job/30778520081?pr=6) on a [clean branch](https://github.com/jeremyleung521/mdtraj/pull/6) and all the test passed on numpy2! 
Here's a placeholder warning message that shows up during import. Would something like this be sufficient? We could also put a link to the PR with explanation and what not as well.

```
In [1]: import mdtraj
/Users/User/Documents/Important/Pitt/git/mdtraj/mdtraj/__init__.py:221: UserWarning: To quickly make mdtraj compatible with Numpy 2, we are globally changing the numpy promotion behavior to match that from Numpy 1.X to bypass changes introduced in NEP50. This might affect other programs.
  warn('To quickly make mdtraj compatible with Numpy 2, we are globally changing the numpy promotion behavior to match that from Numpy 1.X to bypass changes introduced in NEP50. This might affect other programs.', UserWarning)
```

mdtraj/utils/unit/mymatrix.py
jeremyleung521: I must've missed it. :/

mdtraj/formats/pdb/pdbfile.py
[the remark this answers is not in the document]
jeremyleung521: It's been like a month but I recall that when I set all of the `unit_lengths` to float64 in the Trajectory object, when we do traj.save(), it will effectively write something different due to precision difference, causing some of the other conversion tests to fail.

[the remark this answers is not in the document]
janash: Ah, okay. Maybe I should actually try it out and see :) I'm wondering if it's going to be possible to maintain the exact same behavior without the same type promotion rules or if we'll eventually have to make some decisions that might change some numbers slightly.

mdtraj/utils/unit/mymatrix.py
mattwthompson: Adding to the (my) confusion, the entire units module is vendored from OpenMM and ... well I'm not sure what it does here

jeremyleung521:
Okay. I just pushed a few commits which did the following:
- Rollback some of the bad commits that caused 1.26 to fail
- Fix most of the tests (mostly in alignment and geometry) that made most tests pass
- Merged in the cython 3.0 commit.

The CI for 1.26 still fails for `test_rdf.py`, due to a single bad value in one of the reference files (`O_O_rdf_t.txt`). I think it probably needs to be remade... The file name contains alphabet capital O not a zero. This test is faulty in numpy 2.X as well, but with more bad values.
```
FAILED tests/test_rdf.py::test_compare_rdf_t_master - AssertionError: 
Arrays are not almost equal to 5 decimals

Mismatched elements: 1 / 20000 (0.005%)
Max absolute difference: 0.0123327
Max relative difference: 9.37646821e-06
 x: array([[1.16901e+05, 0.00000e+00, 0.00000e+00, ..., 9.75763e-01,
        1.02728e+00, 1.05117e+00],
       [0.00000e+00, 0.00000e+00, 0.00000e+00, ..., 1.04546e+00,...
 y: array([[1.16901e+05, 0.00000e+00, 0.00000e+00, ..., 9.75763e-01,
        1.02728e+00, 1.05117e+00],
       [0.00000e+00, 0.00000e+00, 0.00000e+00, ..., 1.04546e+00,...
```

What's left is to deal with all the casting errors in the warning.

mdtraj/utils/unit/mymatrix.py
jeremyleung521: > Adding to the (my) confusion, the entire units module is vendored from OpenMM and ... well I'm not sure what it does here

It does all sorts of unit conversions. I did a quick look and it seems like not much have changed. Just looking ahead, I'm assuming all our changes here could be backported to openmm if/when they decide to support numpy 2.

janash:
@jeremyleung521 - thanks! I'm planning to look at this in-depth tomorrow.

jeremyleung521:
The latest set of commits include #1932 's commits and also revert the changes that caused the rdf test to fail in 1.26, pending more investigation.

jeremyleung521:
Some more digging and fixing:

Most of the problems are errors propagated from the np.histogram line in [`mdtraj/geometry/rdf.py`](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/geometry/rdf.py#L210). There are no options in terms of casting, and there are [a lot of math involved over 200+ lines](https://github.com/numpy/numpy/blob/v2.0.0/numpy/lib/_histograms_impl.py#L680-L897), so I'm not even sure it's worth going through. So I did the next best thing. Here's the docs for [`np.histogram`](The values of the histogram).

Of the 61 mismatched elements, 8 of them result from calculations of `r` and `V` with"edges" (`bin_edges`, numpy1 outputs `float32`, numpy2 in `float64`), the 52 are from the "tmp" (`hist` from `np.histogram`). I fixed it by pre-computing the bin_bounds in `float32` using `np.linspace`, and then feeding it to `np.histogram`. That way, everything that comes out would be of the same precision because the input would all be float32. Not sure if that's ok?

There is also a discrepancy in value (first one in the output) of which I am unable to fix, so I replaced that value with the output. It's not really a huge problem because the value is in 1e5 with an error of ~0.0064. The old test still passes, so 🤷‍♂️.

Anyhow I'll let other people decide on whether these are good approaches or not. It's really arbitrary when cast to float64/float32 and maybe the easiest way is to remake all the reference file? Who knows.

jeremyleung521:
@janash With that, I think the PR is ready for review (again).

tests/test_lammpstrj.py
```diff
@@ -67,7 +67,7 @@ def test_read_write_0():
         xyz2, new_lengths, new_angles = f.read()
 
     eq(lengths, new_lengths)
-    eq(angles, new_angles)
+    eq(angles, new_angles, decimal=3)
```
janash, line 70: Hey Jeremy, I was very interested in this test in particular since the way it has been fixed has been to basically increase the allowed difference between arrays. I did an investigation of the test and also looked at the effect of the data type in `LAMMPSTrajectory`, see here: https://github.com/janash/lammps_mdtraj_comparison

I think we may want to want to increase the data type in LAMMPS Trajectory to be `np.float64`. I don't really understand why it is `np.float32`.

mattwthompson: +1 for using `float64` anywhere it keeps LAMMPS tests passing

jeremyleung521: I agree with using float64 as well. That's a very comprehensive repo @janash! 

I dug through the lammps code just a little (I never used it before), and it's funny that [all angles writes/calculations are essentially double](https://github.com/lammps/lammps/blob/59bbc5bcc1104bdb4fb45107cd65b5d4d76dbc00/src/angle_write.cpp) everywhere I looked anyways.

janash: I got a little preoccupied with the test, I suppose 😇 

As far as the data type - `np.float64`, vs `np.float32` - is it that `np.float64` is the default data type for an array if not specified? (Like `np.ones(100)` will be `np.float64`?). Is there a default float type? If there is, was it different in the past/do we expect it to be different in the future?

jeremyleung521: > I got a little preoccupied with the test, I suppose 😇
> 
> As far as the data type - `np.float64`, vs `np.float32` - is it that `np.float64` is the default data type for an array if not specified? (Like `np.ones(100)` will be `np.float64`?). Is there a default float type? If there is, was it different in the past/do we expect it to be different in the future?

Default's np.float64, it seems
https://numpy.org/doc/stable/reference/arrays.dtypes.html#specifying-and-constructing-data-types

janash: @jeremyleung521 - also great to know that a need for `np.float32` isn't coming from anywhere in the LAMMPS writer

jeremyleung521: The new commit below (4ca025a) should effectively swap over to `np.float64` for `LAMMPSTrajectory`'s angles, lengths and xyz.

mattwthompson:
I'll try reviewing this in the next week - any pointers on how best to go about it?

jeremyleung521:
> I'll try reviewing this in the next week - any pointers on how best to go about it?

I've mostly been focusing on getting the NEP50 warnings out of the way and fixing the tests to ensure they pass (and have parity) between numpy1 and numpy2.  

* Having multiple conda/venv environments with editable installs to switch between (one for numpy1, numpy2). I've effectively pinned to `numpy<3` for mdtraj so you could just pip install (or conda install) whichever version on top.
* Adding `weak_and_warn` promotion state to the tests for numpy2 only to warn about NEP50 changes (essentially, revert this commit: https://github.com/mdtraj/mdtraj/pull/1928/commits/8d18a5ae1e5280984009574d848248b75ab38f13)
* First approve and merge #1932 on re-activating DSSP tests, which should reduce the number of tracked changes/changed files in this PR

janash:
Hey @jeremyleung521 - Looking at this again! I was just taking a look at the RDF test that you mentioned. These are my findings currently and I'm wondering how they go with your observations.

I ran the RDF analysis using "old Numpy" and saved the RDF (in the test `rdf_O_O`) as a pickle. Then I did it using NumPy 2.0 and also saved the same thing.

When plotting just the raw RDF (excluding the frame/first element), the plots look like the following:

Old:
rdf_old_values

New:
rdf_new_values

These look pretty similar, but here is the difference (new - old):
rdf_difference_values

This was without any of the fixes in this PR. Can you point out which points in your PR address this? Your most recent comment says it's currently just one different value in the RDF, but I'd like to understand where these different values are coming from, because it looks like some kind of shift in the calculation (big positive difference then big negative difference in the opposite direction) and only occurs sometimes (not every time step). It's also odd to me that the difference seem to get progressively smaller. Compared to the RDF values themselves the difference values are pretty small though.

janash:
Sorry @jeremyleung521 - I see your earlier comment on `np.histogram` now.  I'll take a closer look at your described solution

> Of the 61 mismatched elements, 8 of them result from calculations of r and V with"edges" (bin_edges, numpy1 outputs float32, numpy2 in float64), 52 are from the "tmp" (hist from np.histogram). I fixed it by pre-computing the bin_bounds in float32 using np.linspace, and then feeding it to np.histogram. That way, everything that comes out would be of the same precision because the input would all be float32. Not sure if that's ok?

janash:
> Some more digging and fixing:
> 
> Most of the problems are errors propagated from the `np.histogram` line in [`mdtraj/geometry/rdf.py`](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/geometry/rdf.py#L210). There are no options in terms of casting, and there are [a lot of math involved over 200+ lines](https://github.com/numpy/numpy/blob/v2.0.0/numpy/lib/_histograms_impl.py#L680-L897), so I'm not even sure it's worth going through. So I did the next best thing. Here's the docs for [`np.histogram`](https://numpy.org/doc/stable/reference/generated/numpy.histogram.html).
> 
> Of the 61 mismatched elements, 8 of them result from calculations of `r` and `V` with"edges" (`bin_edges`, numpy1 outputs `float32`, numpy2 in `float64`), 52 are from the "tmp" (`hist` from `np.histogram`). I fixed it by pre-computing the bin_bounds in `float32` using `np.linspace`, and then feeding it to `np.histogram`. That way, everything that comes out would be of the same precision because the input would all be float32. Not sure if that's ok?
> 
> There is also a discrepancy in value (first one in the output) of which I am unable to fix, so I replaced that value with the output in the reference file. It's not really a huge problem because the value is of magnitude 1e5 with an error of ~0.0064. The numpy 1.26 test still passes, so 🤷‍♂️.
> 
> Anyhow I'll let other people decide on whether these are good approaches or not. It's really arbitrary when cast to float64/float32 and maybe the easiest way is to remake all the reference file? Who knows.

@jeremyleung521 - thanks so much for your thorough investigation of this problem. This seems like a lot to figure out. Based on your analysis, it seems like it would be reasonable to regenerate the reference files letting np.float64 be the default. We've taken the approach in the LAMMPSTrajectory class to use np.float64 (though that preserved old behavior). Based on the RDF difference I looked at, it doesn't seem like it would affect any scientific conclusions, so probably either decision is okay. It just seems more straightforward and cleaner to use the default behavior.

mattwthompson:
Shall we give Jeremy the honors of pressing the big green button?

PR #1929, opened 2024-09-19
Sort bonds to avoid make_molecules_whole failures

### LINKED ISSUE #1925: make_molecules_whole fails when bonds are not well ordered
Hello,

mdtraj's image_molecules was failing for some of our systems, and I could trace the source of the bug to `make_whole` in [image_molecules.pxi](https://github.com/mdtraj/mdtraj/blob/fd80ab1c14a83175a4f17829887cada99b9204ed/mdtraj/geometry/src/image_molecules.pxi#L15C1-L34C78) .

Since it always moves the second atom in the bond with respect to the first, it will fail if your bonds list is [(1,3), (2,3)] because it moves 3 -> 1 and then again 3 -> 2.

An example with molecules: This gets wrapped properly
```
CRYST1    2.000    5.000    2.000  90.00  90.00  90.00 P 1           1
ATOM      1  OW  SOL     1       1.000   4.715   1.000  1.00  0.00           O  
ATOM      2  HW1 SOL     1       0.400   0.500   1.000  1.00  0.00           H  
ATOM      3  HW2 SOL     1       0.400   4.000   1.000  1.00  0.00           H  
CONECT    1    2    3
END
```
But this doesn't
```
CRYST1    2.000    5.000    2.000  90.00  90.00  90.00 P 1           1
ATOM      1  HW1 SOL     1       0.400   0.500   1.000  1.00  0.00           H  
ATOM      2  HW2 SOL     1       0.400   4.000   1.000  1.00  0.00           H  
ATOM      3  OW  SOL     1       1.000   4.715   1.000  1.00  0.00           O  
CONECT    1    3
CONECT    2    3
END
```
(Of course water molecules are not an issue since they never get written the second way, but we've had issues for larger systems because of this)

This behavior might be intended and is probably the reason for the `sorted_bonds` argument (?), but the sorting done in the `image_molecules` & `make_molecules_whole` code by default doesn't solve the case above.
```
if sorted_bonds is None:
    sorted_bonds = sorted(self._topology.bonds, key=lambda bond: bond[0].index)
    sorted_bonds = np.asarray(
        [[b0.index, b1.index] for b0, b1 in sorted_bonds],
        dtype=np.int32,
    )
```

One potential solution could be to change the bond sorting algorithm to always add bonds to known atoms only
```
def sort_bonds(bonds):
    """Sorts a list of bonds such that they are in a continuous chain.
    bonds: np.array of shape (n, 2)
    """
    # Start with the atom of the first bond
    sorted_bonds = []
    known_atoms = set([bonds[0][0]])
    bonds = list(bonds)
    # Continue until there are no bonds left to sort
    while bonds:
        remaining_bonds = []
        for bond in bonds:
            # Check if the edge connects to any known atom
            if bond[0] in known_atoms:
                sorted_bonds.append(bond)
                known_atoms.add(bond[1])
            elif bond[1] in known_atoms:
                sorted_bonds.append(bond[::-1]) # Put the known atom first!
                known_atoms.add(bond[0])
            else:
                remaining_bonds.append(bond)
        if remaining_bonds and (len(remaining_bonds) == len(bonds)):
            # No new bonds were added - we completed a disconnected subgraph / molecule
            # Just start a new molecule by adding one atom, and continue
            known_atoms.add(remaining_bonds[0][0])
        bonds = remaining_bonds
    return np.array([[a1.index, a2.index] for a1, a2 in sorted_bonds], dtype=np.int32)
```
This is what we've been using as a workaround and seems to work ok.

eruijsena:
Fixes #1925 by adding a function which sorts the list of bonds such that they continuously traverse the molecular graph, thereby preventing wrapping atoms to disconnected neighbours in `make_molecules_whole`.

The function now uses a Breadth-First Search to order the bonds in near linear time complexity.

I did some quick tests, here is how it compares to the previous version I posted in the issue, and to python's `sorted` which it would replace, on a fully disconnected graph.

image

And here is a comparison between the `image_molecules` of this PR and master on a simulation of a slightly large system:

image

As a side-note, the BFS returns only the minimum number of bonds to connect all atoms in a molecule (minimum spanning trees), so technically there could even be performance improvements for densely connected molecular graphs as the loop iterates over less bonds.


fwaibl:
I agree that this would be a very welcome change, I also sometimes had problems with molecules not being processed properly by "make_molecules_whole"

Best,
Franz

tests/test_trajectory.py
```diff
@@ -891,6 +891,21 @@ def test_smooth(get_fn):
...
+    t2 = md.load(get_fn("water2_pbc.pdb"))
+
+    t1.make_molecules_whole(inplace=True)
+    t2.make_molecules_whole(inplace=True)
+
+    # Test that bond lengths are preserved
+    bonds1 = [(a1.index, a2.index) for a1, a2 in t1.topology._bonds]
+    bond_lenghts1 = [np.linalg.norm(t1.xyz[0, a1] - t1.xyz[0, a2]) for a1, a2 in bonds1]
+    bonds2 = [(a1.index, a2.index) for a1, a2 in t2.topology._bonds]
+    bond_lenghts2 = [np.linalg.norm(t2.xyz[0, a1] - t2.xyz[0, a2]) for a1, a2 in bonds2]
```
jchen0506, line 905: minor typo here.

jchen0506 (approved):
Overall the code looks good to me and it should fix the issue. There are 2 minor typos in the test case but it still runs fine.

PR #1932, opened 2024-10-02
Fixing DSSP Tests

jeremyleung521:
The DSSP tests are currently not working or running, mostly because of dependencies (dependent on an old version of mkdssp, not easily installable) This modifies the tests so they actually run with a modern version of [dssp](https://github.com/PDB-REDO/dssp), at least locally. 

1. New option/feature to skip outputting PDB header lines (MDL, Remarks etc.), since `mkdssp` apparently won't run with them in the way. Defaults to true (like usual) but if there are problems, you can invoke this with `traj.save('output_file.pdb', header=False)` or `traj.save_pdb('output_file.pdb', header=False)`.
2. There are some discrepancies with secondary structure assignments between [DSSP v4.4.7](https://github.com/PDB-REDO/dssp) and what mdtraj outputs. Pytest results: 
https://gist.github.com/jeremyleung521/3ff52fcc9c44cb5ad18b73d0e6e58931

Will dig deeper on why we're getting 2., but here's some of the work, if anyone wants to build upon that.

### DIFF
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -1545,7 +1545,7 @@ class Trajectory:
                 types=[a.name for a in self.top.atoms],
             )
 
-    def save_pdb(self, filename, force_overwrite=True, bfactors=None, ter=True):
+    def save_pdb(self, filename, force_overwrite=True, bfactors=None, ter=True, header=True):
         """Save trajectory to RCSB PDB format
 
         Parameters
@@ -1561,6 +1561,10 @@ class Trajectory:
         ter : bool, default=True
             Include TER lines in pdb to indicate end of a chain of residues. This is useful
             if you need to keep atom numbers consistent.
+        header : bool, default=True
+            Include header in pdb. Useful if you want the extra output, but sometimes prevent
+            programs from running smoothly.
+
         """
         self._check_valid_unitcell()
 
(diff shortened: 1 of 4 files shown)


jeremyleung521:
So I had some time to think about the discrepancies. The missing 'P' (PPII) is fine, since those are not in the base DSSP definitions and likely not in the dssp code in mdtraj. But the conflation between B/S/T (Beta Sheets, Bend, Turn) like [here](https://gist.github.com/jeremyleung521/3ff52fcc9c44cb5ad18b73d0e6e58931#file-gistfile1-txt-L302) is a little concerning and worth mulling/debugging on. Are our definitions here reasonable?

Here's a list of all the codes in mkdssp: https://github.com/PDB-REDO/dssp/blob/trunk/doc/mkdssp.md

jeremyleung521:
Last update. The outputs are completely the same if we use [ostrokach::dssp](https://anaconda.org/ostrokach/dssp), which contains dssp `v2.2.1`. It seems mdtraj uses the [same code](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/geometry/src/dssp.cpp) from v2.2.0, so that makes sense. But that conda package is only available for linux, so I don't know if we want to do all our tests on that, but it is self contained and contains libcifpp which necessary to run things. That is not the case for a lot of the other install options.

dssp 4.4.7 on conda-forge doesn't work. dssp 4.2.2 on Ubuntu/apt behaves differently (assumes mcif file input). The state of the dssp program is very perplexing.

jeremyleung521:
The last few commits remove the need for `mkdssp v2.2.1` to be installed. Added in the reference files for comparison.

Should be ready for review.

Issue #1937, opened 2024-10-18
`Trajectory.atom_slice` uses different orders for coordinates and topology


pbuslaev:
[`Trajectory.atom_slice` method](https://github.com/mdtraj/mdtraj/blob/1748df86e8c56968ad129b151ff3ce0f5e67617e/mdtraj/core/trajectory.py#L2006) attempts to generate a subtrajectory for the atom slice. However, if the `atom_indices` parameters are not sorted, this can lead to an error. What the method does is it [takes coordinates](https://github.com/mdtraj/mdtraj/blob/1748df86e8c56968ad129b151ff3ce0f5e67617e/mdtraj/core/trajectory.py#L2028) for `atom_indices` without sorting:

```python
xyz = np.array(self.xyz[:, atom_indices], order="C")
```

and then [creates a topology for a subset](https://github.com/mdtraj/mdtraj/blob/1748df86e8c56968ad129b151ff3ce0f5e67617e/mdtraj/core/trajectory.py#L2031):
```python
topology = self._topology.subset(atom_indices)
```

[`Topology.subset` method](https://github.com/mdtraj/mdtraj/blob/1748df86e8c56968ad129b151ff3ce0f5e67617e/mdtraj/core/topology.py#L1097) is a wrapper over [`Topology._topology_from_subset` method](https://github.com/mdtraj/mdtraj/blob/1748df86e8c56968ad129b151ff3ce0f5e67617e/mdtraj/core/topology.py#L65). This method iterates over all atoms in sorted way and checks if atom index is in atom indices. Thus, if provided atom_indices is not a sorted array, the order of coordinates in `xyz` and atoms in `topology` does not match. The real use case for this to happen was application of the following function to [`3QAP` cif structurte downloaded from rcsb](https://www.rcsb.org/structure/3QAP):

```python
def cif_to_pdb(path_to_cif: str, ligand_names_list: list[str] = []) -> str:
    """
    Path to cif file.
    """
    base = os.path.basename(path_to_cif)
    dir_name = os.path.dirname(path_to_cif)
    
    name, _ = os.path.splitext(base)
    
    out = os.path.join(dir_name, f"{name}.pdb")
    
    st = md.load(path_to_cif)
    
    prot = st.topology.select("protein").astype(int)
    water = st.topology.select("water").astype(int)
    ligands = (st.topology.select("resname " + " ".join(ligand_names_list)) if ligand_names_list else np.array([])).astype(int)
    
[...]

PR #1938, opened 2024-10-18
Sorting atom_indice in atom_slice

pbuslaev:
This should solve #1937


jeremyleung521 (approved):
LGTM. 

But I suggest rerunning the checks/CI due to large amounts of changes from the `numpy2` branch we just merged in. If someone with permission could do it or you could just pull all upstream `master` changes into your fork.

Issue #1939, opened 2024-10-23
mdconvert doesn't work with the latest version of NetCDF4


xiki-tempula:
when running `mdconvert T4lys_ligand_solvated_pruned.nc -o ddddsdss.xtc` and using NetCDF4 as backend, mdconvert would give the error of

```
Traceback (most recent call last):
  File "/Users/zwu/miniconda3/envs/mdtraj/bin/mdconvert", line 11, in <module>
    sys.exit(entry_point())
             ^^^^^^^^^^^^^
  File "/Users/zwu/miniconda3/envs/mdtraj/lib/python3.11/site-packages/mdtraj/scripts/mdconvert.py", line 597, in entry_point
    main(args)
  File "/Users/zwu/miniconda3/envs/mdtraj/lib/python3.11/site-packages/mdtraj/scripts/mdconvert.py", line 324, in main
    with InFileFormat(fn, "r") as infile:
         ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zwu/miniconda3/envs/mdtraj/lib/python3.11/site-packages/mdtraj/formats/netcdf.py", line 169, in __init__
    self._handle = netcdf(filename, mode=mode, **input_args)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: netcdf_file.__init__() got an unexpected keyword argument 'format'
```
The environment is created with `conda create -n mdtraj -c conda-forge mdtraj NetCDF4`

conda list

```
astunparse                1.6.3              pyhd8ed1ab_2    conda-forge
blosc                     1.21.6               h5499902_0    conda-forge
bzip2                     1.0.8                h99b78c6_7    conda-forge
c-ares                    1.34.2               h7ab814d_0    conda-forge
c-blosc2                  2.15.1               h5063078_0    conda-forge
ca-certificates           2024.8.30            hf0a4a13_0    conda-forge
certifi                   2024.8.30          pyhd8ed1ab_0    conda-forge
cftime                    1.6.4           py311h0f07fe1_1    conda-forge
hdf4                      4.2.15               h2ee6834_7    conda-forge
hdf5                      1.14.4          nompi_h99fbd1e_101    conda-forge
icu                       75.1                 hfee45f7_0    conda-forge
krb5                      1.21.3               h237132a_0    conda-forge
libaec                    1.1.3                hebf3989_0    conda-forge
[...]

PR #1940, opened 2024-10-23
test mdconvert for NetCDF4 backend

xiki-tempula:
Try to see if I could reproduce the issue #1939

### DIFF
--- a/tests/test_mdconvert.py
+++ b/tests/test_mdconvert.py
@@ -45,7 +45,7 @@ on_py3 = sys.version_info >= (3, 0)
 
 def test_nc_input(tmp_path, get_fn):
     output_file = tmp_path / "output.xtc"
-    command = f"mdconvert {get_fn("mdcrd.nc")} -o {output_file}"
+    command = f"mdconvert {get_fn('mdcrd.nc')} -o {output_file}"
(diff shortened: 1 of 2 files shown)


mattwthompson:
Please keep `mdconvert` tests in `tests/test_mdconvert.py`

xiki-tempula:
@mattwthompson Moved. Do you mind kick up the workflow?

mattwthompson:
Is there a way to authorize your changes so you can continue to iterate without us needing to click? I know commits from anybody who has committed to the repo doesn't need this click, but it also seems obtuse to need to add you to an organization role, `CODEOWNERS` file, or similar.

I get why GitHub defaults to this behavior for public forks, but you're not just some random stranger here.

xiki-tempula:
Thanks. Sorry, I'm just being lazy and just change it on the fly in website so it takes a couple iterations. I kind of opened this branch to demonstrate that this is a bug. I tried to fix it but cannot figure out why netcdf.Dataset doesn't take format keyword. Do you mind take over and try to see if you could fix the issue? I think the current test shall definitely reproduce the bug.

xiki-tempula:
This problem will be fixed by 1.10.1 release

PR #1943, opened 2024-11-07
RDF Reference for NumPy 2

janash:
This PR regenerates the RDF reference files using NumPy 2.0 and rolls back the RDF calculation function from the current NumPy 2 branch.


mdtraj/geometry/rdf.py
```diff
@@ -173,7 +173,7 @@ def compute_rdf_t(
         shape=(2,),
         warn_on_cast=False,
     )
-
+
```
jeremyleung521, line 176: some white space here.

jeremyleung521 (approved):
LGTM otherwise.

PR #1949, opened 2024-11-19
Consolidate/update config files

mattwthompson:
A few of these files are outdated, others can be wrapped up into `pyproject.toml`. I personally think TOML is a bad format but I think the benefits of consolidating config into fewer files is a value add.

Some adjacent tasks that would be nice to do in separate PRs are
* Switch version handling to `versioningit` or [`setuptools-scm`](https://github.com/pypa/setuptools-scm)
* Update ... *gestures broadly* the build system
* Look through CI logs for other warnings/noise
* Rename default branch to `main`


jeremyleung521 (approved):
LGTM.

Issue #1947, opened 2024-11-14
`atom_slice` loses information on chain IDs


jakublala:
```
traj = md.load(f'{output_dir}/{filename}_solvated.pdb')
print(traj.topology.atom(1).residue.chain.chain_id)
traj = traj.atom_slice(traj.topology.select('protein'))
print(traj.topology.atom(1).residue.chain.chain_id)
>> A
>> None
```

I am not sure if this is the expected behaviour, but it's unexpected and caused a bug in my code. If this is expected due to some philosophical nature of how `atom_slice` should work, then feel free to close this.

mattwthompson:
I don't think that's intended

I think chain information might be lost here: https://github.com/mdtraj/mdtraj/blob/bf857af4aaf9b269af9d111214500f572b3f52a2/mdtraj/core/topology.py#L92

k9cdt:
I realized there has been an existing issue before I opened PR #2014 just now. That one-line fix should address this issue.

jakublala:
Thanks for fixing!

PR #1951, opened 2024-11-21
Preserve `chain_ids` for `to_openmm` and `from_openmm`

rjdirisio:
Hi all,

This PR perserves `chain_id`s when going to/from openmm topologies. It is pretty analogous to https://github.com/mdtraj/mdtraj/pull/1858. 

This allows preservation of chain IDs when reading/writing PDBx/mmCIF files, since, for reading, `PDBXTrajectoryFile.__init__` uses `from_openmm`, leveraging openmm's `PDBxFile` object that maintains chain IDs itself,  and, for writing, the `PDBxTrajectoryFile.write` method uses `to_openmm`.

The `topology` property, and the `topology.setter` in `HDF5TrajectoryFile` need extra work to carry through chain IDs in topologies there since they use a `topology_dict` that doesn't carry through chain IDs.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,9 +1,6 @@
 name: CI
 on:
   push:
-    branches:
-      - master
-      - main
   pull_request:
     branches:
       - master
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,6 +1,9 @@
 name: CI
 on:
   push:
+    branches:
+      - master
+      - main
   pull_request:
     branches:
       - master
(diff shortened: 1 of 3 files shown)


mattwthompson:
#1947?

rjdirisio:
@mattwthompson  #1947 is a duplicate of #1853 . PR #1858 addresses those two issues.

This PR addresses a separate issue, where chain IDs are lost on `to`/`from_openmm` calls, which happen inside the loads/writes of `PDBxTrajectoryFile`. This also gets loads/writes of hdf5 files part of the way to supporting chain IDs as well, because currently they don't .

janash (approved):

Issue #1956, opened 2024-12-06
Store formal charges from OpenMM 8.2+


mattwthompson:
Context: https://github.com/openmm/openmm/issues/4739#issuecomment-2523750194

janash:
Closed by #1967

PR #1967, opened 2025-01-18
Store and write formal charges from PDB/OpenMM

janash:
This PR adds capabilities to store formal charge from OpenMM and write to PDB , addressing #1956.

* A `formal_charge` attribute to the `Atom` class that is set to `None` by default. I've added it as the last argument (and it is optional), so this should not break any existing code. 
* When reading an OpenMM topology, the atom object is inspected for the `formalCharge` attribute. If present, it is read from the OpenMM topology and stored in the MDTraj topology (`Atom` with `formal_charge` attribute mentioned above).
* The PDB writer is modified to write the formal charge as the last two columns in an ATOM entry. If `formal_charge` is `None` for the atom, two spaces are written. Otherwise, the charge value is written.
* **While completing this PR, I also stumbled on a bug with reading the charge from PDB**. The charge was not likely to ever be read correctly for two reasons (1) the string associated with a charge was cast to `int`. This would not succeed in the case that the charge was format {number}{sign}, which is common. (2) The charge read from the PDB was not used in the internal topology (perhaps this is obvious since the attribute didn't exist before). Both of these have been corrected.
* Tests have been added for the PDB charge reading/writing, and for checking that formal charges from OpenMM ->MDTraj topology are preserved.


mdtraj/core/topology.py
```diff
@@ -385,7 +395,12 @@ def to_openmm(self, traj=None):
                         element = None
                     else:
                         element = app.Element.getBySymbol(atom.element.symbol)
-                    a = out.addAtom(atom.name, element, r)
+                    
+                    # If we are using a compatible version of OpenMM, add formal charge
+                    if formal_charge:
```
mattwthompson, line 400: Could this be handled by directly checking the attribute?

```suggestion
                    if hasattr(atom, "formalCharge"):
                        ...
```

I lazily checked, starting from an environment which had 8.2:


```console
$ ipython                                                                                                15:00:59  ☁  0.16.7 ☂
Python 3.11.11 | packaged by conda-forge | (main, Dec  5 2024, 14:21:42) [Clang 18.1.8 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.30.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import openmm.app

In [2]: topology = openmm.app.PDBFile("../openff-interchange/protein.pdb").topology

In [3]: first_atom = next(topology.atoms())

In [4]: hasattr(first_atom, "formalCharge")
Out[4]: True

In [5]:
Do you really want to exit ([y]/n)? y
(openff-interchange-env) [openff-toolkit]
$ micromamba install "openmm =8.1" -c conda-forge -yq                                                    15:03:45  ☁  0.16.7 ☂
(openff-interchange-env) [openff-toolkit]
$ ipython                                                                                                15:04:05  ☁  0.16.7 ☂
Python 3.11.11 | packaged by conda-forge | (main, Dec  5 2024, 14:21:42) [Clang 18.1.8 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.30.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import openmm.app

In [2]: topology = openmm.app.PDBFile("../openff-interchange/protein.pdb").topology

In [3]: first_atom = next(topology.atoms())

In [4]: hasattr(first_atom, "formalCharge")
Out[4]: False

janash: Yes, that seems much cleaner for the `from_openmm` method.

I can update that, then put a `try/except` in `to_openmm`

janash:
Can someone approve? @jeremyleung521 or @mattwthompson

jeremyleung521 (approved):
Just tested it and works well!

PR #1969, opened 2025-01-25
Allow `mdtraj.load` with `top=None`

### LINKED ISSUE #1968: Unexpected behaviour of `mdtraj.load` with `top=None`
I noticed that `mdtraj.load` behaves differently when the `top` kwarg is explicitly set to `None` compared to ommitting the `top` kwarg entirely:

````python
mdtraj.load("frame.pdb")
# Works as expected, loading the PDB file without the need of a topology
````

````python
mdtraj.load("frame.pdb", top=None)
# TypeError: A topology is required. You supplied top=None
````

I think this is unexpected and probably unintended. It could be easily fixed with a change to [this line here](https://github.com/mdtraj/mdtraj/blob/bc401410aec9d862dbc92d2622b360e07eb32462/mdtraj/core/trajectory.py#L396) by avoiding passing `None` on to `_parse_topology`. Happy to submit a PR if you are interested.

As a bit of additional motivation, beeing able to call `load` with `top=None` would allow for consistent calls in both contexts with and without topology without checking how to call `load` beforehand:

````python
# Grabbing trajectory and optional topology from somewhere else
trajectory_file, topology_file = some_function()

# Right now we need to do:
if topolgy_file is None:
    trajectory = mdtraj.load(trajectory_file)
else:
    trajectory = mdtraj.load(trajectory_file, top=topology_file)

# Instead we could just always do:
trajectory = mdtraj.load(trajectory_file, top=topology_file)
````

janjoswig:
Fixes #1968

### DIFF
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -385,18 +385,18 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
             "Each filename must have the same extension. " "Received: %s" % ", ".join(set(extensions)),
         )
 
-    # pre-loads the topology from PDB for major performance boost.
+    # Pre-loads the topology from PDB for major performance boost
     topkwargs = kwargs.copy()
-    # if top is not given try with one of the trajectory files
-    topkwargs.pop("top", None)
     topkwargs.pop("atom_indices", None)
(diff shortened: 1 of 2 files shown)


jeremyleung521 (approved):
LGTM

PR #1972, opened 2025-01-30
add flaky to a few test_distance tests

jeremyleung521:
A few tests were failing on one specific configuration (one python version, macos) and require manual restarting. I added `pytest-rerunfailures` as a pytest plugin, which would automatically rerun the tests 3 times before officially declaring a failure. 

https://github.com/jeremyleung521/mdtraj/actions/runs/13057906340/job/36433449570
https://github.com/mdtraj/mdtraj/actions/runs/12101259345/job/33740887047

Also had to restart `test_superpose_refinds` FOUR TIMES to get it working. Not sure what that was about. Leaving that be for now.
https://github.com/mdtraj/mdtraj/actions/runs/12966061236/job/36247585956

### DIFF
--- a/devtools/conda-envs/test.yaml
+++ b/devtools/conda-envs/test.yaml
@@ -15,4 +15,5 @@ dependencies:
   - gsd>=2.8
   - pytest
   - pytest-xdist
+  - pytest-rerunfailures
   - netCDF4
--- a/tests/test_distance.py
+++ b/tests/test_distance.py
@@ -37,6 +37,10 @@ from mdtraj.geometry.distance import (
 )
 from mdtraj.testing import assert_allclose, eq
 
+
+flaky_on_macos = pytest.mark.flaky(condition=sys.platform.startswith('darwin'), reruns=3, reason='flaky on macos')
+
+
 N_FRAMES = 20
 N_ATOMS = 20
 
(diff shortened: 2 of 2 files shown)


janash:
Do these issues happen on macs, or does it just happen with the CI?

jeremyleung521:
> Do these issues happen on macs, or does it just happen with the CI?

Mostly just the CI (as far as I can tell).

janash:
I pulled out my Mac to try out these tests. I ran one of the problem tests 500 times (`test_compute_distances_core_periodic`) using another pytest plugin and it passed every time (this is a slow test). This does make it seem more like an issue with GHA, though I'm really curious why this would be happening.

janash (approved):
I'm fine with adding this plugin, but would still like to know why the CI fails intermittently on MacOS.

mattwthompson (approved):
🪄🪄🪄 magic 🪄🪄🪄

jeremyleung521:
> I'm fine with adding this plugin, but would still like to know why the CI fails intermittently on MacOS.

It's only doing it on apple silicon (aarch64) and numpy 1 only, but I've seen it happen on different python versions. Who know what's happening since it's spitting out NaN instead of floats or ints.

Issue #1950, opened 2024-11-21
Type annotations


mattwthompson:
It'd be nice to eventually have type annotations distributed with MDTraj. I don't think this has been discussed, so this can serve as a meta-issue if we make progress on this.

I have some experience gradually adding annotations to existing codebases, though none with Cython like we have here. NumPy now has pretty good support for annotations, especially on newer Pythons.

Before trying anything out, the direction I'd like to start is to slowly add annotations to the public Python classes.. `Topology`/`Trajectory`/etc. are how most users interact with the codebase and where type annotations could have an "80/20" sort of impact. Type-checkers are sorta designed to give maintainers flexibility to start from zero annotations and add small amounts incrementally[^1], even while much of a codebase doesn't pass. The first few changesets would be pretty goofy-looking[^2] and pedantic[^3] , but over time the value[^4] compounds on itself.

Another common stopgap is to generate stub files ([the automation](https://mypy.readthedocs.io/en/stable/stubgen.html) is okay) and distribute them as a separate package. This would be a safe option since the stub package could be optional and effectively "opt-in" a.k.a. won't break existing workflows.

[^1]: https://adamj.eu/tech/2022/08/23/python-type-hints-gradually-add-types-for-third-party-packages/
[^2]: lots of `type: ignore`, lots and lots!
[^3]: lots of "we know the argument is an integer, but okay I guess now we have `x: int`"
[^4]: finding bugs, squishier "old Python" paradigms like duck-typing, and, to a certain extent, code that is more self-documenting

janash:
@jchen0506 is going to work on this! She's starting with `Topology`. 

@mattwthompson - we can discuss this, but what are your thoughts on type annotation "versions". For example, should we use `List` from the typing module, or the built-in `list`?

mattwthompson:
👋 hello @jchen0506! Happy to have you on board

Anything Python 3.9+ can use built-ins for `list`, `dict`, `tuple`, and a few others (https://peps.python.org/pep-0585/) which I believe is now the standard and also my preference. I think we should be targeting 3.11 or 3.12 for new development (especially if we want to use generics! https://peps.python.org/pep-0695/).

At the end of the day, I let linters handle it: https://github.com/asottile/pyupgrade?tab=readme-ov-file#pep-585-typing-rewrites / https://docs.astral.sh/ruff/rules/non-pep585-annotation/

Typing an existing codebase is usually quite the journey, I'm happy to support and help out where I can

[...]

PR #1977, opened 2025-02-07
Type annotation - Topology

jchen0506:
This PR is working on issue #1950
As suggested  in the issue, this PR added type annotation  to the `Topology` module because of its larger impact to the users. I'll start another PR for other core module like `Trajectory`


mattwthompson:
Are you running a type-checker on the codebase (maybe mypy)?

It'd be useful to run it in CI as well. This would probably require a new section in `pyproject.toml` to handle the config

jchen0506:
> Are you running a type-checker on the codebase (maybe mypy)?
> 
> It'd be useful to run it in CI as well. This would probably require a new section in `pyproject.toml` to handle the config

I'm not. I will first try to run this locally and then working on CI implementation later. This could benefit the future type annotation process.

mdtraj/core/topology.py
```diff
@@ -1715,76 +1848,76 @@ def __eq__(self, other):
...
+    def __repr__(self) -> str:
         return "Single"
 
-    def __float__(self):
+    def __float__(self) -> float:
         return 1.0
 
 
-Single = Single()
+Single: Singleton = Single()
```
janash, line 1873: The class is defined then immediately written over with an instance of the class (and this is why `mypy` is giving errors about `Single` and `Double`, etc being defined twice). It's kind of a weird pattern.

The easiest option is to tell `mypy` to ignore this error using `#type: ignore[no-redef]`

Another option would be to restructure the code. However, for the sake of this PR, I'd say the tag is the most appropriate.

mdtraj/core/topology.py
```diff
@@ -114,16 +121,15 @@ def _topology_from_subset(topology, atom_indices):
                     )
                     old_atom_to_new_atom[atom] = newAtom
 
-    bondsiter = topology.bonds
+    bondsiter: Iterator[Bond] | list[Bond] = topology.bonds
     if not hasattr(bondsiter, "__iter__"):
-        bondsiter = bondsiter()
+        bondsiter = list(bondsiter)
```
janash, line 126: Making this a `list` seems like a behavior change, and I'm leaning towards not doing it. Do we have any insight into when this condition might be used? (what would `topology.bonds` have to be to go through here?

If the tests are not hitting this line, it's possible that we might not see the effect of a change through tests. We may want to check the codecov :)

janash (changes requested):
I didn't get to make it through the whole thing, but I wanted to leave this comment/ask for clarification

mdtraj/core/topology.py
jchen0506: I checked the codecov and line 126 is not covered by the tests although while I was reading it, this part of the code is just trying to ensure that `bondsiiter` is an iterator  and mypy is right about `bondsiter` is not a callable. I think after adding all the type annotations, it should be already gaurenteed as an iterator so this check may not be necessary.

mattwthompson: I'm not sure what kind of an iterator this is, but potentially unwrapping a generator is an undesirable change

mattwthompson:
Are there plans to include type-checking in CI? That's typically what's done, even for gradual typing approaches

mdtraj/core/topology.py
janash: Definitely agree! We do not want to unwrap a generator. For the typing on line 124, it would be *best* if we could figure out what was the intention of this `if` (when did they expect it to go through this?)

However, lacking that what we can tell is that if we go through the if statement `bondsiter` is a callable that gives us something we can iterate through (as seen later in the `for` loop.

So I think the typing on like 124 would be

```
bondsiter: Iterable[Bond] | Callable[[], Iterable[Bond]] = topology.bonds
```

And to change the portion under the `if` statement back to the original code.

janash:
> Are there plans to include type-checking in CI? That's typically what's done, even for gradual typing approaches

We definitely should!

mdtraj/core/topology.py
jchen0506: @janash I agree this is the better solution. I'll change it accordingly. Thank you!

mdtraj/core/topology.py
```diff
@@ -312,17 +317,16 @@ def join(self, other, keep_resSeq=True):
                     atom_mapping[atom] = a
 
         for bond in other.bonds:
-            a1, a2 = bond
             out.add_bond(
-                atom_mapping[a1],
-                atom_mapping[a2],
+                atom_mapping[bond.atom1],
```
janash, line 321: There are a few spots in the PR where changes like this have been made. It's notable (and should be discussed) because in the original case, we have something like

```
a1, a2 = bond
```

Then use of `a1` and `a2`. This is essentially saying that whatever is in `bond` unpacks to two values (one assumes the atoms in the bonds).

However, in the new usage, we use `bond.atom1` and `bond.atom2`. 

While it may work, these two approaches show different assumptions about the behavior of `bond`. For the first, it's assumed it's something like a tuple that can be unpacked. For the second, it's an object with attributes named `atom1` and `atom2`. 

We may be able to say based on other parts of the code that we expect `bond` to have `atom1` and `atom2` attributes. However, I'd prefer to keep these closer to the original code just so we know for sure that we aren't changing anything unintentionally.

mdtraj/core/topology.py
```diff
@@ -385,19 +395,22 @@ def to_openmm(self, traj=None):
...
+                    ):  # This will occur if the version of OpenMM does not support formal charge (version<=8.2)
                         a = out.addAtom(atom.name, element, r)
                     atom_mapping[atom] = a
 
         for bond in self.bonds:
-            a1, a2 = bond
             out.addBond(
-                atom_mapping[a1],
-                atom_mapping[a2],
+                atom_mapping[bond.atom1],
```
janash, line 412: Same comment as above.

mdtraj/core/topology.py
```diff
@@ -577,18 +594,35 @@ def from_dataframe(cls, atoms, bonds=None):
...
+                "You supplied a %s" % type(bonds),
             )
 
         if not np.all(np.arange(len(atoms)) == atoms.index):
             raise ValueError(
                 "atoms must be uniquely numbered " "starting from zero.",
             )
-        out._atoms = [None for i in range(len(atoms))]
+
+        # create a default atom to fill in the topology to replace the original None placeholder
```
janash, line 611: This seems like it would be a significant change. I'd be in favor of using a tag to ignore this list comprehension (it's internal to a function and replaced anyway) along with a comment like

```
# Create a list of None value placeholders
# type ignore is to avoid mypy error
out._atoms = [None for i in range(len(atoms))] # type: ignore
```

mdtraj/core/topology.py
```diff
@@ -613,7 +651,7 @@ def from_dataframe(cls, atoms, bonds=None):
                     atom["resSeq"],
                     atom["segmentID"],
                 )
-
+            assert r is not None, "Expected a Residue, but got None"
```
janash, line 654: Is this a new check?

mdtraj/core/topology.py
```diff
@@ -1144,7 +1219,7 @@ def select_expression(self, selection_string):
         fmt_string = "[atom.index for atom in topology.atoms if {condition}]"
         return fmt_string.format(condition=condition)
 
-    def select(self, selection_string):
+    def select(self, selection_string: str) -> NDArray[np.int32]:
```
janash, line 1222: Is `np.int32` right here? This function creates a NumPy array on line 1235

```
indices = np.array([a.index for a in self.atoms if filter_func(a)])
```

I would actually guess it would be `np.int64` most of the time. It may be safest here to just type hint it as `NDArray[np.integer]`

mdtraj/core/topology.py
```diff
@@ -1167,11 +1242,14 @@ def select(self, selection_string):
...
-        filter_func = parse_selection(selection_string).expr
+        filter_func = parse_selection(selection_string).expr  # type: ignore
         indices = np.array([a.index for a in self.atoms if filter_func(a)])
         return indices
 
-    def select_atom_indices(self, selection="minimal"):
+    def select_atom_indices(
+        self,
+        selection: str = "minimal",
+    ) -> NDArray[np.int32]:
```
janash, line 1252: Same comment here. Perhaps use `np.integer`

mdtraj/core/topology.py
```diff
@@ -1200,15 +1278,41 @@ def select_atom_indices(self, selection="minimal"):
...
+            )
         elif selection == "heavy":
-            atom_indices = [a.index for a in self.atoms if a.element != elem.hydrogen and a.residue.is_protein]
+            atom_indices = np.array(
+                [
+                    a.index
+                    for a in self.atoms
+                    if a.element != elem.hydrogen and a.residue.is_protein
+                ],
+                dtype=np.int32,
```
janash, line 1305: I'm not sure we should set the dtype like this explicitly. It may conflict with the default integer type on most systems (which is what would have been returned previously)

I'm also thinking of the previous PR we did with NumPy 2.0 where LAMMPS was using 32 bit numbers (causing some issues) and it was unclear why that choice was made. If we make an explicit choice on types like this for a data structure, it should be documented and for a reason rather than seeming somewhat arbitray.

mdtraj/core/topology.py
```diff
@@ -1200,15 +1278,41 @@ def select_atom_indices(self, selection="minimal"):
...
+            )
         elif selection == "water":
-            atom_indices = [a.index for a in self.atoms if a.name in ["O", "OW"] and a.residue.is_water]
+            atom_indices = np.array(
+                [
+                    a.index
+                    for a in self.atoms
+                    if a.name in ["O", "OW"] and a.residue.is_water
+                ],
+                dtype=np.int32,
```
janash, line 1314: Same comment as above.

janash (changes requested):
Thanks @jchen0506 for this PR. I know this has been a TON of work.

That being said, I do still have some requests for changes.

There are a few places where I've requested changes to keep it closer to the original code.

Beyond that, there is also a choice in this PR so far to use a particular data type in many cases (`np.int32`) that I think goes beyond the scope of adding type hinting. I'd like to type hint more generically for `np.integer` and leave the previously occurring behavior (where numpy arrays were created but not explicitly typed)

mdtraj/core/topology.py
mattwthompson: Strongly agree with Jessica here - even if functional code changes do happen, I'd prefer they be in smaller and more atomic PRs and not smuggled into a larger and ostensibly unrelated change. (I'm sympathetic to the desire to kill two birds with one stone and have done it plenty myself - in fact it's from those experiences that I've learned how easy it is to accidentally sneak in unintentional behavior changes/regressions that look fine at the time)

mdtraj/core/topology.py
jchen0506: Yes, this is to check the Residue (r) is not None before it's used, similar to the Chain is not None check on line 627

mypy.ini
```diff
@@ -0,0 +1,8 @@
+[mypy]
+files = /Users/jingchen/Documents/MolSSI/mdtraj/mdtraj/core/topology.py
```
janash, line 2: This file path should be relative or otherwise not "hard coded"

mdtraj/core/topology.py
jchen0506: I agree. I have changed all the `np.int32` to a more general `np.integer` to be on the safe side.

mdtraj/core/topology.py
jchen0506: Thanks for the explanation. This makes sense. I have changed them back to the original.

janash (approved):
I think it's ready to go! Thanks for your work, @jchen0506.

I'm fine with adding the mypy check to CI in a different PR.

mdtraj/core/topology.py
```diff
@@ -114,16 +121,15 @@ def _topology_from_subset(topology, atom_indices):
...
     if not hasattr(bondsiter, "__iter__"):
         bondsiter = bondsiter()
 
     for bond in bondsiter:
         try:
-            atom1, atom2 = bond
             newTopology.add_bond(
-                old_atom_to_new_atom[atom1],
-                old_atom_to_new_atom[atom2],
+                old_atom_to_new_atom[bond.atom1],
```
janash, line 131: Oops! Missed this one. Can you change this back to the old usage like we discussed?

mdtraj/core/topology.py
```diff
@@ -246,26 +252,25 @@ def copy(self):
                     out.add_atom(atom.name, atom.element, r, serial=atom.serial)
 
         for bond in self.bonds:
-            a1, a2 = bond
-            out.add_bond(a1, a2, type=bond.type, order=bond.order)
+            out.add_bond(bond.atom1, bond.atom2, type=bond.type, order=bond.order)
```
janash, line 255: same here

janash (changes requested):

janash (approved):

janash:
I'm going to add CI for mypy in a PR.

@mattwthompson - are you okay with a merge for this?

PR #1980, opened 2025-02-07
Set up `cibuildwheel`

mattwthompson:
Upstream/refresh of #1833


.github/workflows/wheel.yaml
```diff
@@ -4,6 +4,9 @@ on:
   push:
   pull_request:
   workflow_dispatch:
```
mattwthompson, line 6: We'd probably ... not want these triggers, right?

jeremyleung521: maybe PR, but not push and workflow_dispatch... They won't upload unless it's a tagged version anyways.

pyproject.toml
```diff
@@ -7,6 +7,18 @@ requires = [
     "Cython>=3.0.0",
 ]
 
+[tool.cibuildwheel]
```
jeremyleung521, line 10: I personally prefer putting all of these into the yaml file (just so they're all in one place) but that'll lead to duplicate lines. Not sure what people prefer.

jeremyleung521:
Seems like cibuildwheel is pinning delocate somewhere, so that delocate fix in pyproject.toml is still needed. I assume we can be remove it next time when cibuildwheels get updated.

Otherwise I'm done with all the commits i wanted to make.

jeremyleung521:
Doubt that hardly anyone would be using the windows wheels, but here we are...  (Assuming they have zlib installed, of course.)

pyproject.toml
mattwthompson: Putting commands which will only been seen by CI strikes me as a little strange. (Maybe this is because I haven't come across it in such a black-and-white case?). But I guess some options in this file in my other projects are effectively CI-only. I see the value in consolidating configs in one place!

jeremyleung521: Moved them all into the build.yaml. The one thing that pushed me over the line is that we had two places to declare which python versions to use and which builds to skip, which is unnecessarily complicated.

.github/workflows/wheel.yaml
janash: Do we want the `cron`? It says its set up for every Monday.

janash (commented):
LGTM - left one comment about the pinning of `delocate`

.github/workflows/wheel.yaml
janash: I didn't notice before that the last step (uploading to PyPI) only happens when a release is published. So I suppose running every week seems fine :)

.github/workflows/build.yaml
```diff
@@ -13,41 +13,31 @@ jobs:
...
-          CIBW_ARCHS: all
-          CIBW_BUILD: ${{ env.CIBW_BUILD }}
+          CIBW_ARCHS_WINDOWS: auto
+          CIBW_ARCHS_MACOS: "universal2"
+          CIBW_ARCHS_LINUX: native
+          CIBW_BUILD: "cp3${{ matrix.python }}-*"
+          # delocate 0.11.0 has this issue:
+          # https://github.com/matthew-brett/delocate/issues/229
+          # https://github.com/matthew-brett/delocate/issues/227
+          CIBW_REPAIR_WHEEL_COMMAND_MACOS: "pip install 'delocate<0.11.0' && delocate-wheel --require-archs {delocate_archs} -w {dest_dir} -v {wheel}"
```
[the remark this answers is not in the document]
janash: Nevermind, I had missed Jeremy's comment last week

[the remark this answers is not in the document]
jeremyleung521: We could technically use `CIBW_DEPENDENCY_VERSIONS: 'latest'`, but that's at the expense of reproducibility.

 https://cibuildwheel.pypa.io/en/stable/options/#dependency-versions

mattwthompson:
I'm fine with merging this whenever you two are happy with it, I might not have the time to dig into it myself and don't want to block it

janash (approved):

janash:
I went through the workflow file and options for `cibuildwheel` and it all seems reasonable to me. I'm not quite sure I understand the version pinning, but I trust Jeremy :)

jeremyleung521:
> I went through the workflow file and options for `cibuildwheel` and it all seems reasonable to me. I'm not quite sure I understand the version pinning, but I trust Jeremy :)

I usually use 'latest' for the `cibuildwheel` dependencies but I'll be more conservative here with mdtraj and stick with the default :)

PR #1988, opened 2025-02-24
Remove scipy import in `_lprmsd.pyx` to make mdtraj submodules importable without scipy

rjdirisio:
Hi all,

The cython file `_lprmsd.pyx` imports `scipy.spatial.distance`, even though it doesn't use anything from `scipy` at all. This prevents any submodule of mdtraj from being imported without scipy installed, since mdtraj's `__init__.py` imports lprmsd for `__all__` . 

It also appears that care was used to _selectively_ import scipy in various places throughout the repository:
- [here, only in smooth()](https://github.com/mdtraj/mdtraj/blob/76a7563dc708c7a2c3918c0b506b3eaaca34272d/mdtraj/core/trajectory.py#L2142)
- [here, only in AmberNetCDFRestartFile](https://github.com/mdtraj/mdtraj/blob/76a7563dc708c7a2c3918c0b506b3eaaca34272d/mdtraj/formats/amberrst.py#L540-L545)
- [here, only inside kabsch_sander()](https://github.com/mdtraj/mdtraj/blob/76a7563dc708c7a2c3918c0b506b3eaaca34272d/mdtraj/geometry/hbond.py#L314)

So, I imagine that scipy is meant to be a looser dependency than, say, numpy.  The `_lprmsd.pyx` file is currently preventing that from being the case.

If one removes the import of scipy in `_lprmsd.pyx`, then one can `import mdtraj`, or, e.g. `from mdtraj.core.trajectory import Trajectory` successfully without scipy installed.

### DIFF
--- a/mdtraj/rmsd/_lprmsd.pyx
+++ b/mdtraj/rmsd/_lprmsd.pyx
@@ -26,7 +26,6 @@
 ##############################################################################
 import cython
 import numpy as np
-import scipy.spatial.distance
 from mdtraj.utils import ensure_type
 
 cimport numpy as np


mattwthompson (approved):
LGTM as a standalone change

rjdirisio:
> Do you have some custom build infrastructure that gets around this, that uses some MDTraj submodules

Yes, I have a docker container whose dependencies I am trying to make as lean as possible. One can `pip install mdtraj --no-deps` and be less heavy-handed with dependencies. Yes, I do not use the scipy modules in this environment, and scipy is a very heavy dependency. One can observe the zipped size as [40MB](https://pypi.org/project/scipy/#files), which, unpacked, becomes even larger

mattwthompson:
Wow, I recall it being closer to 10 MB a few years ago. It looks like my recollection isn't on target, but it's also climbed quite a bit over the year. In the first half of 2021, for example, it climbed from a 27 MB tarball to 36 MB

https://pypi.org/project/scipy/1.6.0/#files
https://pypi.org/project/scipy/1.7.0/#files

But, then again, I'm used to dealing with libraries on the order of 100s of MB so I forget people elsewhere in the Python world actually benefit from these (relatively) smaller changes

mattwthompson:
Thanks for the patch!

kyleabeauchamp:
Good stuff, this is what OSS is all about 👍

PR #1989, opened 2025-02-25
Add logo in SVG

janash:
I recently needed to use the MDTraj logo in a presentation recently and I redrew it as an SVG. Might as well include it in the repository!


jeremyleung521 (approved):
LGTM

PR #1999, opened 2025-03-11
move rng/ptraj generation for distance tests into fixture

jeremyleung521:
The generation of a random `ptraj` and other associated variables for `test_distances.py` are currently at the top of the test file. Unfortunately, some tests below share the same variable names (e.g. `xyz`), which means they could be overwritten when the tests are run out of order using `pytest-xdist`. My hypothesis is that this is [causing some](https://github.com/jeremyleung521/mdtraj/actions/runs/13777969584/job/38530872667#step:6:3054) [flaky behavior in the CI](https://github.com/mdtraj/mdtraj/actions/runs/13732317066/job/38411437988#step:6:3026).

I moved all that to `conftest.py` as fixture class variables, which should prevent those variables from being overwritten. Some egregious errors include the parts of the `xyz` array being rewritten [into NaNs](https://github.com/mdtraj/mdtraj/blob/main/tests/test_distance.py#L279), which prevents [other tests from completing](https://github.com/mdtraj/mdtraj/actions/runs/13732317066/job/38411437988#step:6:3026).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -21,8 +21,12 @@
 ##############################################################################
 
 import os
+import itertools
 
 import pytest
+import numpy as np
+
+import mdtraj as md
 
 flaky_pdb_dl = pytest.mark.flaky(rerun=3, reason='github-node flaky pdb dl')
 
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -35,3 +39,24 @@ def _get_fn(fn):
         return f"{test_dir}/data/{fn}"
 
     return _get_fn
+
+
+@pytest.fixture
+def gen_random_ptraj(request, tmp_path):
```
janash, line 45: Can `tmp_path` be removed here? I don't see that it's used (the tests pass locally for me without it).

If you remove `tmp_path`, you could set the scope of the fixture to `class`, but this really isn't critical.

tests/conftest.py
```diff
@@ -35,3 +39,24 @@ def _get_fn(fn):
...
+    """
+    Fixture for preparing a temp location
```
janash, line 47: Docstring update, doesn't seem like the right description.

janash (commented):
Looks great, thanks for catching this!

Just a slight comment/adjustment, but looks good to me

tests/conftest.py
jeremyleung521: All of that are leftover from testing. Made those changes :)

jeremyleung521: Apparently scopes need to be per function (default) or it'll rewrite certain variables :)

jeremyleung521:
@janash If you could approve this, then we could merge this in soon-ish. Thanks!

janash (approved):

janash:
Oops! Thanks for the ping. I guess I forgot to hit "approve" :) Done now, ready to merge.

Issue #1897, opened 2024-06-04
Add netCDF4 as an upstream dependency


sukritsingh:
Building on discussions in #1893 and #1894, wanted to start an issue thread discussing explicitly adding netCDF4 as an upstream dependency for installation.

This thread can also serve as a spot for noting all the changes/issues required to achieve this and all considerations

jeremyleung521:
Reopening this for discussion because I'm now leaning towards removing `netCDF4` as a hard dependency. This is currently a dependency on pip installs (including local) and not on conda installs. (dependency added here: https://github.com/mdtraj/mdtraj/pull/1944) We should really decide on having it on both or neither, for consistency sake. Here's a draft PR to add the hard dependency to conda-forge (https://github.com/conda-forge/mdtraj-feedstock/pull/78).

Tbh it's _really_ weird `netCDF4` is a hard dependency here, since pandas/pytables/networkx isn't. And all of those are equally handled by `delay_import`. Plus functionality is already via `scipy`, so removing `netCDF4` is not going to break anything, unlike some of the other imports.

snippet of `delay_import`:
 https://github.com/mdtraj/mdtraj/blob/500d31ed280457d3fc7649cc2189adf34c73a6f5/mdtraj/utils/delay_import.py#L43-L100

PR #2000, opened 2025-03-11
Remove netCDF4 dependency, set up PyPI build variants

jeremyleung521:
Removing netCDF4 as a hard dependency (see https://github.com/mdtraj/mdtraj/issues/1897).

While I was at it, I added in `mdtraj[tests]` and `mdtraj[optional]`, which should allow you to install most of the other dependencies without having to resort to the test conda env recipe, and `pip install mdtraj[all]` should install both sets. I also updated the `delay_import` documentation to remove `scripttest` (not used anymore) and `scipy` (a hard dependency now).

I believe we are dropping python 3.9 (EOL later this year) so did that in the metadata as well.

### DIFF
--- a/mdtraj/utils/delay_import.py
+++ b/mdtraj/utils/delay_import.py
@@ -95,17 +95,6 @@ MESSAGES = {
 
     conda install -c conda-forge openmm
     """,
-    "scripttest": """
-    The code at {filename}:{line_number} requires the scripttest package,
-    which is a python package for testing command-line applications
-
-    scripttest can be downloaded from https://pypi.python.org/pypi/ScriptTest/,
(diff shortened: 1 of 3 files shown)


mdtraj/utils/delay_import.py
```diff
@@ -95,17 +95,6 @@
 
     conda install -c conda-forge openmm
     """,
-    "scripttest": """
```
janash, line 98: Is this removed because we don't use it anymore?

janash (approved):
LGTM thanks Jeremy

mattwthompson (approved):
My opinion, stated somewhere, is that the PyPI and conda build requirements should match; I don't care if netCDF4 is required or not so long as it's consistent. This is a great solution, I think

Issue #1998, opened 2025-03-06
PDBx/MMCIF Reader for MDTraj


janash:
MDTraj currently does not support reading in PDBx/MMCIF. What are thoughts on adding support for that file format?

We could take [the approach that seems to have been used before](https://github.com/mdtraj/mdtraj/tree/main/mdtraj/formats/pdb) for the PDB reader and port the code over from OpenMM.

[OpenMM PDBxFile](https://github.com/openmm/openmm/blob/bda67c677a30e66f433ea48ae69bbc377c25fa2e/wrappers/python/openmm/app/pdbxfile.py#L50)

janash:
@mattwthompson @jeremyleung521 @sukritsingh - thoughts on being able to read PDBx/MMCIF without OpenMM? Should we look into if it would be possible to decouple the parser from OpenMM?

sukritsingh:
I think this is a great idea! MMCIF reading/writing natively will become more and more relevant as people are handling larger and larger systems (like membrane proteins), so reading in a topology from an MMCIF file is going to become more and more common place. I think also having it decoupled from OpenMM is definitely ideal. Thanks for bringing this up (and sorry for the delayed response! Still juggling things)!

mattwthompson:
Topology, trajectory, or both?

sukritsingh:
I presume as a `Topology`, but that comes from the perspective of hoping that nobody is writing out their Trajectories to MMCIF format (or I fear for their disk storage). I guess an MMCIF could be read-in as a single-frame Trajectory object as well, but I don't know what utility that would have off the top of my head (unless someone was doing single structure analysis/manipulation).

jeremyleung521:
There are formats that mdtraj only reads (and not write) like `.arc`. It'll just throw an error (https://github.com/mdtraj/mdtraj/blob/main/mdtraj/formats/arc.py#L382-L394), so we're free to do just a reader or r/w. 

Though OpenMM has a [writing function](https://github.com/openmm/openmm/blob/bda67c677a30e66f433ea48ae69bbc377c25fa2e/wrappers/python/openmm/app/pdbxfile.py#L259-L284) for PDBx/mmCIF, so it's probably best to port that over too. 

There's also the "official" pdbx/mmcif reader from RCSB we can base off of: https://github.com/rcsb/py-mmcif, https://pypi.org/project/mmcif/

mattwthompson:
[...]

PR #2004, opened 2025-03-18
Port PDBx/mmCIF Reader/Writer from OpenMM to MDTraj

cabb99:
This pull request ports the PDBx/mmCIF file reader and writer from OpenMM into MDTraj, and removes the OpenMM requirement for cif file reading and writing. Most of the dependencies, such as `element` and `Topology`, were already present in MDTraj with slight modifications.

This PR corresponds to issue #1998 

### Changes:  
- Introduced `.pdbx`, `.cif`, `.pdbx.gz`, and `.cif.gz` extensions in `mdtraj/core/trajectory.py`.  
- Ported PDBx container and parsing logic from OpenMM:  
  - `PdbxContainers.py` (mmcif container class)
  - `PdbxReader.py` (mmcif file reader)  
  - `pdbx.py` (loading PDBx/mmCIF files and PDBxTrajectoryFile)  
  - `pdbxfile.py` (Reading and writing interface)  
- Removes OpenMM dependency from `tests/test_pdbx.py`


mattwthompson:
Thanks for getting this started!

The licensing of these files seems all over the place; I think most of OpenMM is permissive but we need to make sure we follow the licenses

janash:
@rjdirisio - PDBxFile currently requires OpenMM as a dependency. We've been talking about decoupling this so that OpenMM is not required to be installed to read PDBx files. The observed behavior would not be any different. The PR ports over code and decouples it from OpenMM so that PDBx files can be read without OpenMM (which is an optional dependency of MDTraj)

cabb99:
@rjdirisio The pdbx module has been modified with the following changes:

- Replaced the OpenMM `PDBxFile` dependency by porting the PDBxFiles (`pdbxfile.py`) and the pdbx readers (`PdbxContainers.py` and` PdbxContainers.py`) from OpenMM
- Substituted OpenMM's Topology with MDTraj's Topology.
- Removed the dependency on `openmm.units`

I forgot to remove the original [mdtraj/formats/pdbx.py](https://github.com/mdtraj/mdtraj/blob/main/mdtraj/formats/pdbx.py) file from my previous commit, it was moved to [mdtraj/formats/pdbx/pdbx.py](https://github.com/cabb99/mdtraj/blob/cec8937555a9fe5e6d20ba7c5e8abe0dd71ddeea/mdtraj/formats/pdbx/pdbx.py).

rjdirisio:
Thanks @cabb99 for the update.

And thanks @janash , that provides helpful context. Something I'd like to note: I am aware of some discussion for openeye CIF reads/writes to start to honoring `chem_comp_atom/bond` for parsing bond orders, instead of [struct_conn](https://github.com/mdtraj/mdtraj/issues/1998#issuecomment-2731744895). Apparently, the `chem_comp_bond/atom` spec was agreed upon as the place to grab the bond orders sometime last year.

Unfortunately, I don't have any public correspondence / any recent GH issues discussing this, nor issues in the context of openmm cif parsing. [The best I could find is this old one](https://github.com/openmm/openmm/issues/1443#issuecomment-240534299). All I can assert is that it appears that there may be a coordinated move away from `struct_conn` field parsing for bond orders in future development of OE's toolkit, and perhaps eventually openmm CIF parsing, if everyone gets on the same page with which parts of the spec should be honored.

I'm not trying to say that it's necessary for this port of the cif parsing to honor or support these new specs / features right now, but I would like you all to be aware of this for future road-mapping.

An example MMCIF leveraging the chem_comp spec is [4X2F](https://www.rcsb.org/structure/4X2F) for a macromolecule, and [HEM](https://www.rcsb.org/ligand/HEM) for a small molecule.

cabb99:
What are some typical use case scenarios for the Topology bonds? I think the `chem_comp_bond` is a good source for the topology, but may not be a replacement. `chem_comp_bond` seems to only define bonds within the same aminoacid, while `struct_conn `defines bonds within different aminoacids, like disulfide bridges. Here is an example: [11BA](https://www.rcsb.org/structure/11BA). Also it seems like `chem_comp_bond` doesn't include the connectivity between aminoacids in the backbone.

pechersky:
It's my understanding that `struct_conn` is meant for things like disulfide bonds and other across-residues-but-not-canonical-or-obvious bonds. It is also used for coordination bonds in some places.

According to spec, "canonical bonds" like the ones in the canonical protein residues, or between adjacent residues in the sequence, need not be specified. But I think toolkits would prefer if they were, and take the "make all bonds explicit" stance on writing.

So I agree with you that you can't by default assume that an arbitrary `chem_comp_bond` will have _all_ the bonds in the system. There is no section in the spec that says "here is the section that necessarily defines in-sequence-inter-residue bonds".

cabb99:
There is some information about what type of canonical bond is used in `_chem_comp`. A more general approach for the topology construction then would be:

- Parse the atom definitions from [`chem_comp_atom`](https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Categories/chem_comp_atom.html)
- Add the bonds from [`chem_comp_bond`](https://mmcif.wwpdb.org/dictionaries/mmcif_std.dic/Categories/chem_comp_bond.html) that are defined for around 80% of entries
- Add the bonds from pre-stored tables for common residues where `chem_comp_bond` may not be defined. 
- Adds the bonds from [`struct_conn`](https://mmcif.wwpdb.org/dictionaries/mmcif_std.dic/Categories/struct_conn.html) for disulfides and non-obvious bonds between residues (including phosphorilation, glycosilation, etc.)
- Adding the canonical bonds between protein/DNA/RNA residues, we could use [`_chem_comp`](https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Categories/chem_comp.html) and have predefined methods for each [linkage type](https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_chem_comp.type.html)

I imagine the most common scenario where this may be useful would be to analyze NMR ensembles.

janash:
@cabb99 - this is on my list to review very soon (I hope this week! 🤞🏻)! Would you mind adding a README in your pdbx folder (similar to [the one for PDB](https://github.com/mdtraj/mdtraj/tree/cec8937555a9fe5e6d20ba7c5e8abe0dd71ddeea/mdtraj/formats/pdb) that explains where the code has come from and what changes were made?

jsoerensen:
@rjdirisio 
Some relevant written notes
https://www.wwpdb.org/news/news?year=2023#649f0801d78e004e766a9680
https://www.rcsb.org/news/feature/66fd92157ed93306c9c8068f
https://www.wwpdb.org/news/news?year=2024#662fa1a565d8a8cacaf76f9c

@janash the above protocol looks like the right approach in my view at least.

mdtraj/core/trajectory.py
```diff
@@ -90,6 +90,10 @@
     ".arc",
     ".hdf5",
     ".gsd",
+    ".pdbx"
+    ".cif",
+    ".pdbx.gz",
+    ".cif.gz",
```
janash, line 96: Are `.gz` files actually supported in this current iteration? I do not see that they are supported in `pdbx.py`, which is where `cif` and `pdbx` is registered. Based on my review, support of CIF/PDBX `gz` files is still to be added, so this should be removed from the list until the change is made.

mdtraj/formats/pdb/pdbfile.py
```diff
@@ -507,10 +502,17 @@ def _write_footer(self):
...
+        if not self._mode == "w":
+            raise ValueError("file not opened for writing")
+
+        # Identify bonds that should be listed as CONECT records.
+
         conectBonds = []
         if self._last_topology is not None:
             for atom1, atom2 in self._last_topology.bonds:
-                if atom1.residue.name not in standardResidues or atom2.residue.name not in standardResidues:
+                if atom1.residue.name not in self._standardResidues or atom2.residue.name not in self._standardResidues:
```
janash, line 515: Here, `standardResidues` has been moved from being defined in the footer to being a class attribute on `PDBFile`. This is done so that it is more accessible for PDBx for header and footer reading and writing.

Considering the need to share this across both file types, I'd recommend moving it out of PDBFile entirely and importing the list to both objects. This way, `PDBFile` is not imported to `PDBxFile`. I would recommend putting this in a module in the formats directory and importing it to both PDB and PDBx

tests/test_pdbx.py
```diff
@@ -29,17 +29,6 @@
...
-    import openmm  # noqa: F401
-
-    HAVE_OPENMM = True
-except ImportError:
-    HAVE_OPENMM = False
-
-# special pytest global to mark all tests in this module
-pytestmark = pytest.mark.skipif(not HAVE_OPENMM, reason="test_pdbx.py needs OpenMM.")
-
-
```
janash, line 42: In addition to the existing test for conversion, we should have a test to make sure a CIF file is read correctly. I can work on adding this - would you like me to? I'd recommend adding a CIF to `tests/data` and checking that the number of frames, atoms, unit cell, etc is correct.

janash (changes requested):
Hey Carlos, thanks so much for this PR! I have a few requests

1. For the standard residues, let's move this out of PDBFile to a common location that can be accessed by both file types.
2. The reader does not currently support `cif.gz` or `pdbx.gz` as described. This will fail at the read step because the extensions have not been added to the check (`_parse_topology` in `trajectory.py`.) However, I did try adding them and testing a `.cif.gz` and it was not successfully parsed, meaning the function does not know support compressed files. The mention of these extensions should be removed until this is fixed.
3. More tests need to be added to make sure the CIFs are read correctly.

Let me know if you would like help with any of these points. I could add (3) easily.

mdtraj/formats/pdb/pdbfile.py
janash: I see now that there is something similar occurring for at least one static method as well. I am going to think about this a little more

mdtraj/core/trajectory.py
cabb99: They are supported now

mdtraj/formats/pdbx/pdbx.py
```diff
@@ -192,8 +196,12 @@ def load_pdbx(
     return traj
 
 
-@FormatRegistry.register_fileobject(".pdbx")
 @FormatRegistry.register_fileobject(".cif")
+@FormatRegistry.register_fileobject(".cif.gz")
+@FormatRegistry.register_fileobject(".mmcif")
```
cabb99, line 201: The mmcif and pdbx suffixes may not be as common as the cif suffix.

mattwthompson:
At the risk[^1] of providing two comments that aren't useful with respect to the code changes, some of these files have a lot of messy style such as Python 2 compatibility. Running a linter through it would help, but I'm open to deferring after the first PR (this one) is merged

[^1]: Okay, that's precisely what I'm doing

mattwthompson:
@cabb99 easiest would be to run `pip install pre-commit && pre-commit run -a`, but there might be some manual fixes in file of that age and size.

Alternatively, since this was probably good to go before I introduced this upstream change, I am happy to "fix" everything myself. Just let me know 🙂

cabb99:
@mattwthompson Thank you for the suggestion. I fixed the linting errors.
@janash I also included some tests to check that the bonds for the topology are being correctly written, and it seems it is working correctly with the current implementation from OpenMM.

mdtraj/formats/pdbx/pdbxfile.py
janash, line 1: When comparing this file to the one currently in OpenMM, the major changes seem to be

1. Different handling of bonds (reading from `_chem_comp_bond`, `struct_conn`, and calling `createStandardBonds`) then deduplicating. 
2. Changing handling of unit cell (now records from MMCIF instead of calculation that was present in OpenMM version).  **This one is a bit notable to me, as people using MDTraj for MMCIFs will likely see a different in their unit cells from this change**
3. Removing OpenMM units.
4. Repeat of functions `_loadNameReplacementTables` and `_parseResidueAtoms` and data `_standardResidues`. I agree with the marked TODO to move these to a standard place, though it doesn't necessarily need to be done in this PR.

janash (approved):
Hi @cabb99 - thanks so much for this contribution, it was a big effort! I'm good with approving the PR with some minor comments.

1. I'd like to see some documentation, particularly in `PdbxContainers.py` in the header on the changes. This was done in this file until 2013, and it's more clear and explicit to keep doing so.
2. There are some divergences from the behavior of OpenMM that are notable and we should make sure are clear. These are 
* bond reading and deduplication. For bond reading, we are readng from `_chem_comp_bond`, `struct_conn` and then using `createStandardBonds` and doing a deduplication step (please correct if wrong).
* change in reading unit cell from MMCIF (OpenMM computed it, while we now read it directly from the MMCIF). 

If other maintainers are able (@mattwthompson, @jeremyleung521, @sukritsingh, @Lnaden) another review would be appreciated.

mdtraj/formats/pdbx/pdbxfile.py
cabb99: The handling of the bonds should be similar to the one in OpenMM. The only thing that is new is the deduplication. The change to the handling of the unit cell was done to be consistent with the mdtraj Trajectory.

jeremyleung521 (commented):
Been testing this with [1UBQ](https://www.rcsb.org/structure/1UBQ) and noticed that it wasn't selecting the correct atoms based on residue numbers selection (not resid, which is assigned by MDTraj). Those should be under the `_atom_site.label_seq_id`  data item. I checked the topology itself and the the residues/ atoms have the correct residue number, but `topology.select` is not catching it.

That is:
```
import mdtraj
a = mdtraj.load('1ubq.cif')
a.top.select('residue 1') # empty array
a.top.select('resid 0') # gives me numpy.array([0,1,2,3,4,5,6,7])
```

Using pdb as a topology will return the expected array, so it's definitely a mmCIF topology parsing thing.

Will add more as I dig around.

mdtraj/formats/pdbx/pdbxfile.py
```diff
@@ -0,0 +1,727 @@
...
+        reader = PdbxReader(inputFile)
+        data = []
+        reader.read(data)
+        if ownHandle:
+            inputFile.close()
+        block = data[0]
+
+        # Build the topology.
+
+        atomData = block.getObj("atom_site")
```
jeremyleung521, line 93: I find this section really wordy with a lot of repeating calls (and loads of variables), only later to realize it's all to find the column numbers for each dataset. Maybe it's easier to do this in a dictionary instead? It would also be nice to have a comment saying `# Mapping datasets to respective column numbers` or something

mdtraj/formats/pdbx/pdbxfile.py
```diff
@@ -0,0 +1,727 @@
...
+        self._unitcell_lengths = None
+        self._unitcell_angles = None
+
+        self._loadNameReplacementTables()
+
+        # Load the file.
+
+        inputFile = file
+        ownHandle = False
+        if isinstance(file, str):
```
jeremyleung521, line 81: file could be a Path object too, i think? Also, it might be nice to put this file opening in a context manager or sth.

mdtraj/formats/pdbx/pdbxfile.py
```diff
@@ -0,0 +1,727 @@
...
+        zCol = atomData.getAttributeIndex("Cartn_z")
+        lastChainId = None
+        lastAltChainId = None
+        lastResId = None
+        lastInsertionCode = ""
+        atomTable = {}
+        atomsInResidue = set()
+        models = []
+
+        for row in atomData.getRowList():
```
jeremyleung521, line 133: I'm glad this works, but some of this code is really wordy with many chained if-statements. Seems like it's going to be real difficult to maintain in the future.

At a minimum, I'm sure just spacing things with empty lines will help and adding in-line comments ABOVE the if statements to show what that next block is doing.

mdtraj/formats/pdbx/pdbxfile.py
```diff
@@ -0,0 +1,727 @@
...
+                        resIds[atom2.residue],
+                        atom2.name,
+                    ),
+                    file=file,
+                )
+            print("#", file=file)
+
+        # Write the header for the atom coordinates.
+
+        print("loop_", file=file)
```
jeremyleung521, line 523: I'm sure doing 20+ print statements in a row probably involves opening and closing the file 20+ times. Might as well open the file once in a context manager + write instead.

janash:
Thanks Jeremy! The review is really appreciated!


As far as the code style points you've highlighted, I believe most of these were likely directly from the OpenMM port, particularly Pdbxcontainers.py is an older (but now updated version) of the same file in py-mmcif. I think there is a little bit of a question for me of (1) how to accurately keep up with changes from ported code, especially since we usually squash commits and (2) if the changes should be contributed back to their respective sources.

This may be a wider question for MDTraj in general considering there is a lot of borrowed code :)

mdtraj/formats/pdbx/pdbxfile.py
jeremyleung521: Ah, You're right! I assumed that `file` was the same as the `Pdbxfile.__init__ ` `file` argument. Please disregard...

mattwthompson:
I can't be counted on to spend time thinking deeply about this, especially as a non-expert in this file format, but as a baseline I have relatively high confidence in the quality of the work given that

* it's a port of something that's been used for a while in a stable project
* tests are passing (but no code coverage report? hmm ...)
* most of the "style only" changes are automated
* there has been significant human review in the process.
* there don't appear to be major lingering questions from the discussion to date

I'm not the first person to note that the code isn't written in quite the same way I'm used to reading or writing Python code. This is neither a good nor bad thing overall, but possibly something to discuss in the future. I don't think this PR is the time to worry about that / this ought not hold up clicking the big green button.

Similarly something to discuss elsewhere and not hold this up: how do we, if at all, want to sync this up with upstream changes?

jeremyleung521:
> I can't be counted on to spend time thinking deeply about this, especially as a non-expert in this file format, but as a baseline I have relatively high confidence in the quality of the work given that
> 
>     * it's a port of something that's been used for a while in a stable project
> 
>     * tests are passing (but no code coverage report? hmm ...)
> 
>     * most of the "style only" changes are automated
> 
>     * there has been significant human review in the process.
> 
>     * there don't appear to be major lingering questions from the discussion to date
> 
> 
> I'm not the first person to note that the code isn't written in quite the same way I'm used to reading or writing Python code. This is neither a good nor bad thing overall, but possibly something to discuss in the future. I don't think this PR is the time to worry about that / this ought not hold up clicking the big green button.
> 
> Similarly something to discuss elsewhere and not hold this up: how do we, if at all, want to sync this up with upstream changes?

I think the `ResSeq` bug should at least be fixed since it affects atom selection, but otherwise I'm ok to merge if we're very clear which parts are modified and which parts are copy-pasted verbatim for maintainability. `blame` isn't very clear at this point and it'll be worse with `squash and merge`, as @janash mentioned. We do not want our fixes to be overwritten by copy/pastes in the future.

@cabb99: regarding the bug, probably a `int(row[resNumCol])` right here should fix it (maybe also in the if statement above), but I haven't tested personally.
https://github.com/mdtraj/mdtraj/blob/728e8b2dd2c9ece642d8bc67ac5b43890a691f5e/mdtraj/formats/pdbx/pdbxfile.py#L171

janash:
Hi @cabb99 - if I understand correctly, you made a vendor branch with just the port from OpenMM, then re-added your changes?

This commit history seems too complex to me for merging to the main branch (OpenMM port + changes, back to just OpenMM port, then re-add changes). 37 commits would be a lot to merge for one feature.

I'd prefer to either squash merge this with fully documented changes or to have a PR that has commits only (1) OpenMM port (2) MDTraj changes.

cabb99:
@jeremyleung521 thank you for the review. I will add a test for that.

janash:
Going to squash merge this afternoon. If there are any objections, say them now!

Issue #2001, opened 2025-03-13
Formal charge for atoms in `to_dataframe` and `from_dataframe`


janash:
Formal charge on atoms in the MDTraj topology was added in #1967 . While this information is present on each atom in the topology, it was not added to the methods `to_dataframe` and `from_dataframe`.

* [to_dataframe](https://github.com/mdtraj/mdtraj/blob/2effa667c7d295558bc312a5ee9a025296f9767f/mdtraj/core/topology.py#L498)
* [from_dataframe](https://github.com/mdtraj/mdtraj/blob/2effa667c7d295558bc312a5ee9a025296f9767f/mdtraj/core/topology.py#L552)

I believe that every other attribute of the atom is shown or can be input from these methods, so it is probably most consistent to add the information to the methods (though columns would be `None` for many topologies that do not have this information).

PR #2007, opened 2025-03-30
Add formal charge to the `to_dataframe()` and `from_dataframe()` methods

jchen0506:
This PR is to addressing issue #2001 
1. Atoms with formal charges are now supported by `to_dataframe()` and `from_dataframe()`
2. Unit tests for these two functions are provided

### DIFF
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -648,7 +648,10 @@ class Topology:
             fc = atom["formal_charge"]
             if fc is not None:
                 if pd.notna(fc):
-                    fc = float(fc)
+                    try:
+                        fc = int(fc)
+                    except ValueError:
+                        raise ValueError(f"Formal charge {fc} cannot be cast to an integer.")
                 else:
                     fc = None
                     
@@ -1792,7 +1795,7 @@ class Atom:
         index: int,
         residue: Residue,
         serial: int | None = None,
-        formal_charge: float | None = None,
+        formal_charge: int | None = None,
     ):
         """Construct a new Atom.  You should call add_atom() on the Topology instead of calling this directly."""
         # The name of the Atom
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -645,14 +645,23 @@ class Topology:
                     atom["resSeq"],
                     atom["segmentID"],
                 )
-
+            fc = atom["formal_charge"]
+            if fc is not None:
+                try:
+                    if pd.notna(fc):
+                        fc = int(fc)
+                    else:
+                        fc = None
+                except Exception:
+                    fc = None
+                    
             a = Atom(
                 atom["name"],
                 elem.get_by_symbol(atom["element"]),
                 atom_index,
                 r, # type: ignore
                 serial=atom["serial"],
-                formal_charge=atom["formal_charge"],
+                formal_charge=fc,
             )
             out._atoms[atom_index] = a
             r._atoms.append(a) # type: ignore
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -647,12 +647,9 @@ class Topology:
                 )
             fc = atom["formal_charge"]
             if fc is not None:
-                try:
-                    if pd.notna(fc):
-                        fc = float(fc)
-                    else:
-                        fc = None
-                except Exception:
+                if pd.notna(fc):
+                    fc = float(fc)
+                else:
                     fc = None
                     
             a = Atom(
--- a/mdtraj/core/topology.py
+++ b/mdtraj/core/topology.py
@@ -608,7 +608,7 @@ class Topology:
         try:
             atoms = atoms.astype({"formal_charge": "Int64"})
         except ValueError as e:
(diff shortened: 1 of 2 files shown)


mattwthompson (approved):

mdtraj/core/topology.py
```diff
@@ -641,13 +645,20 @@ def from_dataframe(
                     atom["resSeq"],
                     atom["segmentID"],
                 )
-
+            fc = atom["formal_charge"]
+            if fc is not None:
+                if pd.notna(fc):
+                    fc = float(fc)
```
janash, line 651: Formal charges are integers by the PDB definition. I understand not wanting to cast as `int` in case someone has input a floating point number, but if something like `1` is cast as a float here it will cause problems down the line if you try to write a PDB (ie it will fail because the line is too long).

I think there should be a check here to make sure the charges can be cast as `int`, and then they should be cast as `int` rather than float.

mdtraj/core/topology.py
```diff
@@ -592,6 +593,9 @@ def from_dataframe(
             if col not in atoms.columns:
                 raise ValueError("dataframe must have column %s" % col)
 
+        if "formal_charge" not in atoms.columns:
```
janash, line 596: I'm not a huge fan of how this function modifies the user's dataframe, but there is already a precedent for this with `segmentID`

janash (changes requested):
Sorry! I have a request on how we are casting the charge. It should be cast as an `int` rather than a float. 

It appears to be mislabeled [here](https://github.com/janash/mdtraj/blob/b7c3406f185a3496efbd1935189008bc826342b7/mdtraj/core/topology.py#L1662) (though it is labeled as `int` for the `add_atom` method). That could be fixed in this PR as well.

mdtraj/core/topology.py
jchen0506: Yes, I agree casting to int makes more sense. I've fixed the incorrect type annotation for the formal charge in atom and also added a check to make sure the charge can be  cast as `int`

mdtraj/core/topology.py
```diff
@@ -531,6 +532,7 @@ def to_dataframe(self) -> tuple[pd.DataFrame, NDArray[np.float64] | None]:
                 "resName",
                 "chainID",
                 "segmentID",
+                "formal_charge"
```
janash, line 535: I've been thinking a bit more about this. It seems like we'd want everything in this dataframe to be an integer or object data type (integer for index, object for name). I checked what is coming out now, and `formal_charge` is being upcast to a floating point number. Apparently this happens when you have `NaN` or `None` for some values in a column, but we could ensure that these are returned as the proper type by doing the following *after* dataframe creation (so my comment here isn't quite in the right spot. Using `Int64` with a capital `I` will allow Nan values - it is the [nullable integer type](https://pandas.pydata.org/docs/user_guide/integer_na.html) from pandas.

```
# Using Int64 makes the data type integer
# but allows for NA values in the column
atoms = atoms.astype({
    "formal_charge": "Int64"
})
```

mdtraj/core/topology.py
janash: It was modifying the dataframe before, so I think it's okay to *not* make a copy. It would change the behavior of the function, so let's leave it for now.

mdtraj/core/topology.py
```diff
@@ -641,13 +645,23 @@ def from_dataframe(
                     atom["resSeq"],
                     atom["segmentID"],
                 )
-
+            fc = atom["formal_charge"]
+            if fc is not None:
```
janash, line 649: Instead of having these if statements, you could cast to `Int64` before iterating over the atoms (with a try/except block and `astype` I mentioned above). I'd advise raising a `ValueError` in the `except`. I believe when casting to `Int64` it will ignore NaNs and also raise an error if a value that is not an integer is encountered. It will allow this to be done all at once (probably must faster than checking individual values) and will also be cleaner in the code.

janash (changes requested):

mdtraj/core/topology.py
jchen0506: Thank you! This is much more cleaner. I've changed the code accordingly and add a normalize function in unit test to normalize the formal charge since after casting with `astype("formal_charge": "Int64")`, all None becomes <NA>

mdtraj/core/topology.py
```diff
@@ -592,9 +597,19 @@ def from_dataframe(
...
+
+
         if "segmentID" not in atoms.columns:
             atoms["segmentID"] = ""
 
+        # Ensure formal_charge is of type Int64 (nullable integer)
+        try:
+            atoms = atoms.astype({"formal_charge": "Int64"})
+        except ValueError as e:
+            raise ValueError(f"Error converting formal_charge to Int64: {e}")
```
janash, line 611: Could this say something else about checking to make sure your charges are integer values?

jchen0506: Sure, thanks for spotting that

janash (approved):
Excellent. Thanks Jing! And thank you for your patience in the review.

PR #2011, opened 2025-04-07
Allow mdtraj.rmsd() to calculate without alignment

### LINKED ISSUE #2008: Allow RMSD to calculate without alignment
The algorithm used to calculate RMSD by default aligns/superposes to the selection (inherent to the Theobald QCP method used). There are many cases where you don't want this to happen (Binding RMSD, etc.). This Issue will track the progress as I implement this.

My current idea is that we won't modify the current default behavior, but by passing `(superpose=False)` into `rmsd`, it will execute a new block of code, calculing `rmsd` without alignment. These new blocks will have to be written in cython/C++, a) for maximum speed and b) b/c the rest of the code is already in cython/C++ (might as well, right?).

But effectively, RMSD will be calculated directly via a cythonized version of:
```
def rmsd_no_superpose(traj1, traj2, idx):
    result = traj1.xyz[:,idx,:] - traj2.xyz[:,idx,:]
    result = result ** 2
    result = np.sqrt(np.mean(np.sum(result, axis=2), axis=1))
    return result
```
(Thank @RobertArbon for this snippet, from https://github.com/mdtraj/mdtraj/issues/1994)

Just noting that this is _not_ a new idea. It's been discussed way back in 2016/2017 by J Chodera and Rob McGibbon, among others. After this first pass, (or if someone else is willing), we might try to work out the math described in https://github.com/mdtraj/mdtraj/issues/1188.

Relevant Issues where this is causing confusion/trouble with RMSD calcs:
https://github.com/mdtraj/mdtraj/issues/1797
https://github.com/mdtraj/mdtraj/issues/1846
https://github.com/mdtraj/mdtraj/issues/1994

Previous discussion on this improvement:
https://github.com/mdtraj/mdtraj/issues/1188 (RMSD)
https://github.com/mdtraj/mdtraj/issues/1210 (RMSF)

jeremyleung521:
Resolves #2008 

Users can now run RMSD without aligning. First pass around this, unless someone wants to work out the math described in https://github.com/mdtraj/mdtraj/issues/1188.

Usage:
```
# Suppose you have `traj1` and `ref` md.Trajectory objects, which have different xyz coordinates

# RMSD without aligning
rmsd1 = md.rmsd(traj1, ref, superpose=False)  # RMSD without aligning

# RMSD after aligning manually
traj1.superpose(ref) # Align traj1 to ref, done inplace
rmsd2 = md.rmsd(traj1, ref, superpose=False)  # Calculate without alignment

# Let RMSD/Theobald QCP do the alignment
rmsd3 = md.rmsd(traj1, ref, superpose=True)  
rmsd3 = md.rmsd(traj1, ref)  # These two are equivalent

# Note `rmsd1` != `rmsd2` == `rmsd3`
```

<strike>I'm not sure how much slower this is, but I'll test it in a little bit.</strike> See EDIT 2 below

I also added tests for my additions, and updated `test_rmsd.py` so it no longer rely on the global instance pRNG. 

Annotated output showing the level of cythonization. (remove .txt)
[_rmsd.html.txt](https://github.com/user-attachments/files/19641225/_rmsd.html.txt)

====

EDIT:
I also tried doing a shortcut (similar to the [Theobald QCP code](https://github.com/mdtraj/mdtraj/blob/969b2e1f71f2dfb4f96230392f4925d9649ede11/mdtraj/rmsd/src/theobald_rmsd_generic.h#L74-L80)) where if the xyz are the same it would return 0. It compiles fine if I run `cythonize -a _rmsd.pyx` but it fails to compile via `pip install`. Not sure what's happening here, but any help would be great. I currently removed these lines so it works.

https://github.com/mdtraj/mdtraj/blob/969b2e1f71f2dfb4f96230392f4925d9649ede11/mdtraj/rmsd/_rmsd.pyx#L727-L729

Error:
```
      mdtraj/rmsd/_rmsd.cpp:22733:25: error: invalid operands to binary expression ('__Pyx_memviewslice' and '__Pyx_memviewslice')
       22733 | __pyx_t_1 = (__pyx_t_16 == __pyx_v_ref_xyz_frame);
             |              ~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~
```

=====

EDIT2: Timing
Calculation is faster [9] than using QCP [8], but if you factor in superposing, It's about 4x slower [10]. Most of the run time is from superposing [11].

```
In [8]: timeit.timeit('mdtraj.rmsd(a,a,0)', setup='import mdtraj; a = mdtraj.load("traj.h5")', number=1000)
Out[8]: 0.025098250014707446

In [9]: timeit.timeit('mdtraj.rmsd(a,a,0, superpose=False)', setup='import mdtraj; a = mdtraj.load("traj.h5")', number=1000)
Out[9]: 0.012531082989880815

In [10]: timeit.timeit('a.superpose(a,0); mdtraj.rmsd(a,a,0, superpose=False)', setup='import mdtraj; a = mdtraj.load("traj.h5")', number=
       ⋮ 1000)
Out[10]: 0.12832299998262897

In [11]: timeit.timeit('a.superpose(a,0)', setup='import mdtraj; a = mdtraj.load("traj.h5")', number=1000)
Out[11]: 0.12596291699446738
```

### DIFF
--- a/mdtraj/rmsd/_rmsd.pyx
+++ b/mdtraj/rmsd/_rmsd.pyx
@@ -101,10 +101,14 @@ def rmsd(target, reference, int frame=0, atom_indices=None,
         the trajectory's coordinates, the center and traces will be out of
         date and the RMSDs will be incorrect.
     superpose : bool, default=True
-        Whether to use the Theobald QCP method to calculate RMSD. If True, the 
-        QCP method is used, which inherently superposes the structure based on 
-        the `atom_indices` selection. If False, RMSD is directly calculated with
-        no optimization and no superposition done. The `precentered` option is ignored.
+        Whether to use the Theobald QCP method to calculate RMSD. If True, the
+        QCP method is used, which inherently superposes the structure based on
+        the `atom_indices` selection. If False, the Theolbald QCP method is not used
+        and the RMSD is directly calculated pairwise with no optimization and additional
+        no superposition done.
+        The `precentered` option is ignored. Users are expected to manually superpose
+        and/or image their trajectories using `Trajectory.image_molecules()` and/or
+        `Trajectory.make_molecules_whole()` and/or `Trajectory.superpose()`.
 
     Examples
     --------
@@ -222,6 +226,11 @@ def rmsd(target, reference, int frame=0, atom_indices=None,
         # t2 = time.time()
         # print 'rmsd: %s, centering: %s' % (t2-t1, t1-t0)
     else:
+        if precentered:
+            warnings.warn(
+                'in rmsd(), precentered is ignored when superpose=False',
+                RuntimeWarning)
+
         if parallel:
             for i in prange(target_n_frames, nogil=True):
                 msd = msd_nosuperpose(n_atoms, target_xyz[i], ref_xyz_frame)
--- a/mdtraj/rmsd/_rmsd.pyx
+++ b/mdtraj/rmsd/_rmsd.pyx
@@ -29,6 +29,7 @@ import warnings
 
 import cython
 import numpy as np
+cimport numpy as np
 from mdtraj.utils import ensure_type
 
 from cpython cimport bool
@@ -202,7 +203,7 @@ def rmsd(target, reference, int frame=0, atom_indices=None,
     
         if parallel:
             for i in prange(target_n_frames, nogil=True):
-                msd = msd_nosuperpose(n_atoms, n_atoms, &target_xyz[i, 0, 0], &ref_xyz_frame[0, 0], target_g[i], ref_g, 0, NULL)
+                msd = msd_atom_major(n_atoms, n_atoms, &target_xyz[i, 0, 0], &ref_xyz_frame[0, 0], target_g[i], ref_g, 0, NULL)
                 distances[i] = sqrtf(msd)
         else:
             for i in range(target_n_frames):
@@ -213,12 +214,13 @@ def rmsd(target, reference, int frame=0, atom_indices=None,
         # print 'rmsd: %s, centering: %s' % (t2-t1, t1-t0)
     else:
         if parallel:
+            #pass
             for i in prange(target_n_frames, nogil=True):
-                msd = msd_nosuperpose(n_atoms, &target_xyz[i], &ref_xyz_frame)
+                msd = msd_nosuperpose(n_atoms, target_xyz[i], ref_xyz_frame)
                 distances[i] = sqrtf(msd)
         else:
             for i in range(target_n_frames):
-                msd = msd_nosuperpose(n_atoms, &target_xyz[i], &ref_xyz_frame)
+                msd = msd_nosuperpose(n_atoms, target_xyz[i], ref_xyz_frame)
                 distances[i] = sqrtf(msd)
  
     return np.array(distances, copy=False)
@@ -696,9 +698,10 @@ def getMultipleAlignDisplaceRMSDs_atom_major(float[:, :, ::1] xyz_align1 not Non
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
-def msd_nosuperpose(int n_atoms,
-                    float[:, ::1] xyz_current not None,
-                    float[:, ::1] xyz_ref not None):
+@cython.cdivision(True)
+cdef const float msd_nosuperpose(const int n_atoms,
+                                 const float[:, :] xyz_current,
+                                 const float[:, :] xyz_ref) nogil:
     """
     Private function to directly calculate the MSD of one frame without alignment.
 
@@ -711,15 +714,14 @@ def msd_nosuperpose(int n_atoms,
     xyz_ref : np.ndarray(float), shape = (n_atoms, 3), dtype = np.float32
         The xyz coordinates of the reference frame.
 
-
     Returns
     -------
     msd : float
         The mean square deviation of xyz_current from xyz_ref.
 
     """
-    cdef Py_ssize_t atom_count
-    cdef Py_ssize_t dims_count
+    cdef int atom_count
+    cdef int dims_count
     cdef float msd_temp = 0
 
     # Shortcut out if same xyz.
@@ -729,6 +731,6 @@ def msd_nosuperpose(int n_atoms,
     # Loop through each atom and each dimension to calculate difference from ref 
     for atom_count in range(n_atoms):
         for dims_count in range(3):
-            msd_temp += (xyz_current[atom_count][dims_count]- xyz_ref[atom_count][dims_count])**2
+            msd_temp += (xyz_current[atom_count][dims_count] - xyz_ref[atom_count][dims_count])**2
     return msd_temp / n_atoms
 
--- a/mdtraj/rmsd/_rmsd.pyx
+++ b/mdtraj/rmsd/_rmsd.pyx
@@ -61,8 +61,9 @@ cdef extern from "math.h":
 
 @cython.boundscheck(False)
 def rmsd(target, reference, int frame=0, atom_indices=None,
-         ref_atom_indices=None, bool parallel=True, bool precentered=False):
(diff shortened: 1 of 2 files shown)


rjdirisio:
Nice! I wanted to ask how this handles periodic boundary conditions? I worry that one could get artificially high RMSDs if a molecule passes over the boundary.

jeremyleung521:
> Nice! I wanted to ask how this handles periodic boundary conditions? I worry that one could get artificially high RMSDs if a molecule passes over the boundary.

Hi @rjdirisio. This option essentially delegates all alignment/PBC to the user, so they'd have to run `image_molecules()` and/or `make_molecules_whole()` themselves.

janash:
@jeremyleung521 - I think we may have missed assigning a reviewer to this?

jeremyleung521:
> @jeremyleung521 - I think we may have missed assigning a reviewer to this?

I think @sukritsingh said he was going to look at  this.

janash:
You're right!

mdtraj/rmsd/_rmsd.pyx
```diff
@@ -97,6 +100,11 @@ def rmsd(target, reference, int frame=0, atom_indices=None,
         be unsafe; if you use Trajectory.center_coordinates and then modify
         the trajectory's coordinates, the center and traces will be out of
         date and the RMSDs will be incorrect.
+    superpose : bool, default=True
+        Whether to use the Theobald QCP method to calculate RMSD. If True, the 
+        QCP method is used, which inherently superposes the structure based on 
+        the `atom_indices` selection. If False, RMSD is directly calculated with
+        no optimization and no superposition done. The `precentered` option is ignored.
```
sukritsingh, line 107: This is purely nit-picky, but if we are highlighting that QCP is being used for the superposed, RMSD, it may be worth specifically saying that QCP is not being used if `superpose=False`

mdtraj/rmsd/_rmsd.pyx
```diff
@@ -164,46 +180,57 @@ def rmsd(target, reference, int frame=0, atom_indices=None,
...
-
-    # t1 = time.time()
-
-    cdef float[:] distances = np.zeros(target_n_frames, dtype=np.float32)
-    if parallel:
-        for i in prange(target_n_frames, nogil=True):
-            msd = msd_atom_major(n_atoms, n_atoms, &target_xyz[i, 0, 0], &ref_xyz_frame[0, 0], target_g[i], ref_g, 0, NULL)
-            distances[i] = sqrtf(msd)
+    if superpose:
+        # t0 = time.time()
```
sukritsingh, line 198: As I read it right now, it seems like the code will ignore the options `indices` arguments if `superpose=False`, which makes sense. However, would it be better to print some kind of warning that we are ignoring those provided arguments (or alternatively, throw an error if both indices and `superpose=False` are given?)

mdtraj/rmsd/_rmsd.pyx
```diff
@@ -675,3 +702,38 @@ def getMultipleAlignDisplaceRMSDs_atom_major(float[:, :, ::1] xyz_align1 not Non
...
+
+    """
+    cdef Py_ssize_t atom_count, dims_count
+    cdef float msd_temp = 0
+
+    # Loop through each atom and each dimension to calculate difference from ref 
+    for atom_count in range(n_atoms):
+        for dims_count in range(3):
+            msd_temp += (xyz_current[atom_count][dims_count] - xyz_ref[atom_count][dims_count])**2
+    return msd_temp / n_atoms
```
sukritsingh, line 738: I'm not seeing any checks in here that the atoms provided are the same in both frames (inherently, this would be checked with `superpose=True` to attain superposition). Are we just going to presume then that the user knows what they're doing if they do `superpose=False`? If so might be useful to just log a warning or something so they know. I like the idea of keeping `superpose=True` as the default behavior!

tests/test_rmsd.py
```diff
@@ -103,13 +111,13 @@ def test_superpose_0(get_fn):
 
 def test_superpose_1():
     # make one frame far from the origin
-    reference = md.Trajectory(xyz=np.random.randn(1, 100, 3) + 100, topology=None)
+    reference = md.Trajectory(xyz=rng.standard_normal((1, 100, 3)) + 100, topology=None)
```
sukritsingh, line 114: Why `normal` instead of `randn`? Shouldn't it work either way?

tests/test_rmsd.py
```diff
@@ -103,13 +111,13 @@ def test_superpose_0(get_fn):
...
     reference_xyz = reference.xyz.copy()
 
     for indices in [None, np.arange(90)]:
         # make another trajectory in a similar rotational state
         query = md.Trajectory(
-            xyz=reference.xyz + 0.01 * np.random.randn(*reference.xyz.shape),
+            xyz=reference.xyz + 0.01 * rng.standard_normal(reference.xyz.shape),
```
sukritsingh, line 120: See above about `randn` vs. `normal`?

sukritsingh (commented):
Sorry for the delay in getting to this! Overall everything looks great! Just a couple of nitpicky comments: 

1. I think with `superpose=False` it might be good to have some better warning/logging that we are ignoring the `indices` provided, since we are allowing both `superpose=False` and indices to be passed at the same time without throwing an error. I think this logging will help reduce confusion/inform users in case they expect one behavior but have asked for a different one.
2. Why did we change from `randn()` to `normal` in the tests? This is just from a "Should this change be made at the same time" kind of perspective --- I think `normal` might be better because it gives us more parameters/control?

tests/test_rmsd.py
jeremyleung521: It should work pretty much identically. I'm just following numpy's [general directive](https://numpy.org/doc/2.2/reference/random/generator.html) to move from the global rng instance (`numpy.random`) to independent generators (`rng`/ `numpy.random.default_rng()` instance we created here). It is very easy to mess up the rng seeds and having independent instances means we don't have to worry about what other dependencies/programs are doing. I've had trouble before with other packages playing around with `np.random.seed()` etc. messing up results so I try to move to the "independent rng" way as I see them.

mdtraj/rmsd/_rmsd.pyx
jeremyleung521: The `indices` arguments are __not__ ignored. The specific atoms are sliced in [L193-L195](https://github.com/mdtraj/mdtraj/blob/e87ed0d462ae944a981234ad405ce038f899b5ad/mdtraj/rmsd/_rmsd.pyx#L193-L195), based on provided atom_indices and ref_atom_indices, and the sliced arrays are passed downwards. There are checks above the makes sure len(ref_atom_indices) == len(atom_indices) but they don't have to be the same selection either.

mdtraj/rmsd/_rmsd.pyx
jeremyleung521: These are all checked in the main function above and would error out if len(n_atoms_ref) != len(n_atoms_frame).

mdtraj/rmsd/_rmsd.pyx
sukritsingh: Ahh I see - I've always operated on the presumption that one would want to compute RMSD on the same atoms, but I see your point about that not always being the case. Makes sense then!

mdtraj/rmsd/_rmsd.pyx
sukritsingh: Awesome! Thanks so much! This is great!

tests/test_rmsd.py
sukritsingh: Fair enough! I wanted to make sure there was a citation listed somewhere that we were following since this seemed like a new change to me. I wasn't aware of the directive - that answers my question!

sukritsingh (approved):
This is a fantastic addition to mdtraj.rmsd() that addresses a whole different set of usecases we never were able to support before. Thanks for doing this, @jeremyleung521!
LGTM!

PR #2012, opened 2025-04-08
Fix Topology double-loading for HDF5 and MOL2 (+ remove false-positive UserWarning)

jeremyleung521:
Currently, loading `.h5` and `.mol2` files with `mdtraj.load()` will always return a `UserWarning`, even though no `top` kwarg is passed. This is because the topology is loaded twice for these file types (these loaders don't take the `top` argument), and the topology from the first load triggers the `UserWarning` right before the second successful topology load.

More precisely, the two topology loads happen 
A) once during [topology parsing](https://github.com/mdtraj/mdtraj/blob/main/mdtraj/core/trajectory.py#L398) and 
B) once during this [try/except block](
https://github.com/mdtraj/mdtraj/blob/main/mdtraj/core/trajectory.py#L442-L444) where topology is discarded (because a `top` argument is given) and the topology is reread again. 

This can lead to a large performance hit for large system topologies. In fact there is a huge comment block/TODO dedicated to this: https://github.com/mdtraj/mdtraj/blob/main/mdtraj/core/trajectory.py#L424-L431 . We might want to implement the TODO eventually, but that will require some planning + going through all formats. This is the "quick fix" that alleviates the performance hit, by skipping the first load only for these file types.

====

BEFORE:

```
In [1]: import mdtraj

In [2]: a = mdtraj.load('data/traj.h5')
/Users/user/mdtraj/mdtraj/core/trajectory.py:440: UserWarning: top= kwargs ignored since this file parser does not support it
  warnings.warn("top= kwargs ignored since this file parser does not support it")
In [3]: a = mdtraj.load('data/traj.h5', top='data/traj.h5')
/Users/user/mdtraj/mdtraj/core/trajectory.py:440: UserWarning: top= kwargs ignored since this file parser does not support it
  warnings.warn("top= kwargs ignored since this file parser does not support it")
```

AFTER:
```
In [1]: import mdtraj

In [2]: a = mdtraj.load('data/traj.h5')

In [3]: a = mdtraj.load('data/traj.h5', top='data/traj.h5')
/Users/user/mdtraj/mdtraj/core/trajectory.py:444: UserWarning: top= kwargs ignored since this file parser does not support it
  warnings.warn("top= kwargs ignored since this file parser does not support it")
```

### DIFF
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -398,7 +398,7 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
 
     # These topology formats do not support the 'top' keyword
     # This is to prevent the loader from reading the topology twice.
-    if _get_extension(top) not in ['.h5', '.hdf5', '.mol2']:
+    if isinstance(top, str) and _get_extension(top) not in ['.h5', '.hdf5', '.mol2']:
         kwargs["top"] = _parse_topology(top, **topkwargs)
 
     # get the right loader
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -397,7 +397,7 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
         top = filename_or_filenames[0]
 
     # This is to prevent the loader from reading the topology twice.
-    if extension not in ['.h5', '.hdf5']:
+    if extension not in ['.h5', '.hdf5', '.arc', '.mol2']:
         kwargs["top"] = _parse_topology(top, **topkwargs)
 
     # get the right loader
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -395,7 +395,10 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
     top = topkwargs.pop("top", None)
     if top is None:
         top = filename_or_filenames[0]
-    kwargs["top"] = _parse_topology(top, **topkwargs)
+
+    # This is to prevent the loader from reading the topology twice.
+    if extension not in ['.h5', '.hdf5']:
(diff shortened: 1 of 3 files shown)


mdtraj/core/trajectory.py
```diff
@@ -395,7 +395,11 @@ def load(filename_or_filenames, discard_overlapping_frames=False, **kwargs):
     top = topkwargs.pop("top", None)
     if top is None:
         top = filename_or_filenames[0]
-    kwargs["top"] = _parse_topology(top, **topkwargs)
+
+    # These topology formats do not support the 'top' keyword
+    # This is to prevent the loader from reading the topology twice.
+    if extension not in ['.h5', '.hdf5', '.mol2']:
```
mattwthompson, line 401: Does MDTraj support loading (I feel filthy putting these words together in succession) MOL2 trajectories?

mattwthompson:
I suppose you can't estimate the performance hit without implementing the fix, but are you thinking it's roughly a factor of 2, owing to the unnecessary double-loading, or something else?

jeremyleung521:
> I suppose you can't estimate the performance hit without implementing the fix, but are you thinking it's roughly a factor of 2, owing to the unnecessary double-loading, or something else?

About a factor of two (1.69s vs 3.45s).
```
In [4]: timeit.timeit('mdtraj.load("traj.h5")', setup='import mdtraj', number=1000)
/Users/user//mdtraj/mdtraj/core/trajectory.py:440: UserWarning: top= kwargs ignored since this file parser does not support it
  warnings.warn("top= kwargs ignored since this file parser does not support it")
Out[4]: 3.446037834015442
```

AFTER FIX:
```
In [3]: timeit.timeit('mdtraj.load("traj.h5")', setup='import mdtraj', number=1000)
Out[3]: 1.6921365419984795
```

mattwthompson (approved):

PR #2014, opened 2025-04-17
Preserve chain_id when creating sub-topology

k9cdt:
When extracting part of a structure into a new trajectory, the original code forgets to copy the `chain_id` field to the new `chain` object, resulting in `None` values.

This one-line fix ensures the `chain_id` is properly transferred when creating a sub-topology.


k9cdt:
This problem may be more ubiquitous than I previous expected. It seems that this attribute is also not saved in the HDF5 format. I added two lines to
1. Save it in new HDF5 topologies
2. Load it when reading `.h5` files, while preserving backward compatibility

k9cdt:
Replaced by #2015

PR #2015, opened 2025-04-17
Preserve chain_id when creating topologies

k9cdt:
When extracting part of a structure into a new trajectory, the original code forgets to copy the `chain_id` field to the new `chain` object, resulting in `None` values.

This fix ensures the chain_id is properly set when 
1. Creating a sub-topology.
2. Saving to new HDF5 topologies
3. Reading from HDF5 topologies, while preserving backward compatibility.

Reopen to replace #2014 because I changed my branch name from `master` to `main`


k9cdt:
Hey dev team, just checking—will someone be able to review this PR when they get a chance? Thanks!

janash:
Hi @k9cdt - @Lnaden should be reviewing. This is on our list for the next release as well

mdtraj/formats/hdf5.py
```diff
@@ -337,6 +338,7 @@ def topology(self, topology_object):
                 chain_dict = {
                     "residues": [],
                     "index": int(chain.index),
+                    "chain_id": str(chain.chain_id)
```
Lnaden, line 341: I think this will cause issues from older files which don't have chain ID's properly saved. I'm seeing a case where no `chain_id` information is available, and so all the chains become the string `"None"`, so all the chains are now identified by the same string, even if their indices are different.

It might be good to have a default value generator or something back up at [289](https://github.com/mdtraj/mdtraj/pull/2015/files#diff-153459b06c92988d868bd11560e26cc1cb45d8562a30c8d1d5fcd1d3ddce3850R289) to the effect of

```python
default_chain_id = 0
for chain_dict in sorted(...):
    if chain_id  in chain_dict :
        chain_id = chain_dict[chain_id]
    else:
         chain_id = default_chain_id
         default_chain_id += 1
```
since i dont think you can functionalize the default value of the `dict.get` method. The crude example i have might also have name collisions, but something to think about.

Lnaden (approved):
Looks fine generally. There might be some name collisions with older files im worried about (i.e. all chains being named "None" the str not the obj.). See comment for details.

mdtraj/formats/hdf5.py
k9cdt: What about in line 341, I change it to something like

```python
"chain_id": str(chain.chain_id) if chain.chain_id else None
```

so it will return an actual `None` rather than a string "None." This will be more verbose at least any `"A" == None` kind of comparison will always return `False`

janash:
Ready to merge?

k9cdt:
Yes I just changed one thing according to Lnaden's recommendation. Should be ready to merge now.

Lnaden:
Good enough for now. If it causes issues down the road, we can improve it in a future release

janash:
Thanks! Now it is oddly not mergeable with main. Would you mind fixing the merge conflict? I could also take a look if you prefer not to

janash:
Some of the tests are flaky. It's likely not a true failure

k9cdt:
Great! Otherwise I think it's ready to merge.

janash:
Thanks for the fix!

Issue #1990, opened 2025-02-25
Sphinx version pinning for compatibility with nbsphinx


janash:
Sphinx is [currently pinned](https://github.com/mdtraj/mdtraj/blob/830b04782646a7297fcf14c287ca8ecd61d5df98/docs/requirements.txt#L3) in the docs dependencies to be less than 7.3.

This is due to an incompatibility with the extension nbsphinx with Sphinx v8.2 (see issue https://github.com/spatialaudio/nbsphinx/issues/825)

Once this has been resolved, Sphinx should be unpinned. Notably, we are pinning a lower version than needed currently. I tested locally and Sphinx 8.1 should be fine.

PR #2016, opened 2025-04-19
unpin sphinx

janash:
I believe we can unpin Sphinx now (#1990) - very minor. The CI should tell confirm if this is true :)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -15,9 +15,6 @@
 import os
 import sys
 
-import msmb_theme
-import sphinx_rtd_theme
-
 from mdtraj import __version__
 
 # If extensions (or modules to document with autodoc) are in another directory,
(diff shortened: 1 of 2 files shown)


mattwthompson (approved):
Might be worth keeping upper constraint, just making it higher than 7.3. Sphinx has more regular major version releases than most packages in our ecosystem and each one is liable to break things

janash:
Pinned to Sphinx 8 so that it won't go above this. Seem alright, or were you thinking a different pin?

The problem causing the pin was an incompatibility between `nbspinx` and `sphinx`, so I'm not a sure how to pin to avoid that, but this should keep it going to v9 if that is released.

I also went ahead and removed `runipy` and the `msmb_theme` mentioned in #2017. `msmb_theme` is in `conf.py`, but I think it no longer was having any effect on the docs build and can be safely removed (same for `runipy`)

mattwthompson:
Yup, exactly. If 8.0 works now, it should:tm: work until version 9 is released

PR #2019, opened 2025-04-29
update note on project status

janash:
How does everyone feel about an update to the project status? I've updated it and added it to the end of the README. There are also some slight adjustments to the header levels in the README

@jeremyleung521 @mattwthompson @sukritsingh


jeremyleung521 (approved):
LGTM.

PR #2020, opened 2025-05-05
add py3dmol visualization to pi stacking

janash:
For @rdirisio or anyone with an interest in docs. As I was reviewing the pi stacking code, I found it helpful to visualize the structures. I used VMD, but I've translated it to py3dmol.  There are also some light edits to the notebook structure. I realize this would mean adding another documentation dependency, so am looking to gather thoughts.

Screenshots from locally built documentation, it's also interactive!

image

image

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -18,4 +18,5 @@ pandas
 numpydoc
 wheel
 tables
+py3Dmol
 # python==3.11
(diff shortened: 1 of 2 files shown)


janash:
Thanks @sukritsingh ! It worked last time I built it :) I'll merge

Issue #2028, opened 2025-05-21
Calculating Bravais vectors


jeremyleung521:
So while trying to debug why the CI is breaking, I stumbled upon the `unitcell` module. Occasionally, the CI (throught the RNG) will generate some combination of cell lengths and cell angles that will fail to calculate `unitcell_vectors` (Bravais vectors). I know nothing about [Bravais](https://chem.libretexts.org/Bookshelves/Inorganic_Chemistry/Chemical_Group_Theory_(Miller)/02%3A_Rotational_Symmetry/2.07%3A_Bravais_Lattices_(3-d)) [Lattices](https://en.wikipedia.org/wiki/Bravais_lattice) (too long since my last group theory class and i don't think we even formally touched Bravais lattices).

According to docstrings, the code itself is pulled from Gyroid, case `self.dim==3` within `__create_shape(self)`.
https://pythonhosted.org/gyroid/_modules/gyroid/unitcell.html
The only difference is that Rob added a "patch" by dividing a term in `cy` with `np.sin(gamma)`, which is not part of the Gyroid code.
https://github.com/mdtraj/mdtraj/blame/main/mdtraj/utils/unitcell.py#L87
This extra division will make the value within `np.sqrt()` negative during the calculation of `cz`.

Not sure how appropriate this "patch" is and probably need someone with more expertise to answer this.. In the following reproducible error, the input themselves (`unitcell_lengths` and `unitcell_angles`) look reasonable, but do note that gamma (last angle) is really close to 180˚, which is a no-no. The `np.sin(gamma)` term in the denominator will reach 0 and obviously for a cubic-esque box, we won't ever have angles at 0 or 180, so maybe this is just a computational precision thing? IDK.

Reproducible code:
```
import numpy as np

# from mdtraj.utils.unitcell import lengths_and_angles_to_box_vectors

unitcell_lengths = np.array([2.4055011, 3.174083 , 2.4759393], dtype=np.float32)
unitcell_angles = np.array([126.96979,  54.83003, 178.20027], dtype=np.float32)

alpha = unitcell_angles[-0] * np.pi / 180 

beta = unitcell_angles[1] * np.pi / 180 

gamma = unitcell_angles[2] * np.pi / 180 

c_length = unitcell_lengths[2]

cx = c_length * np.cos(beta)

# This works, equation from gyroid
cy = c_length * (np.cos(alpha) - np.cos(beta) * np.cos(gamma))

[...]

PR #2029, opened 2025-05-21
Update unitcell.py to check if unitcell is valid

jeremyleung521:
#2028

Basically in `conftest.py`/`test_distances.py`, sometimes we generate random numbers that give unitcells with 0 (or imaginary) volume. This PR fixes it so we always generate valid unitcells for tests. Also (typically) when `traj.unitcell_angles` is called, it will throw the user a warning that the unitcell with invalid angles is given.

This should hopefully fix all of our CI flaky problems. (I know I said this multiple times at this point :/ )


tests/conftest.py
```diff
@@ -58,7 +57,14 @@ def gen_random_ptraj(request):
     request.cls.times = np.array([[i, 0] for i in range(N_FRAMES)[::2]], dtype=np.int32)
 
     request.cls.ptraj = md.Trajectory(xyz=xyz, topology=None)
-    request.cls.ptraj.unitcell_vectors = np.ascontiguousarray(
-        rng.standard_normal((N_FRAMES, 3, 3)) + 2 * np.eye(3, 3),
-        dtype=np.float32,
-    )
+
+    while True:
```
janash, line 61: Is there a way to constrain values from RNG to give us valid unit cells (to avoid while loop)?

(not a review, just a question)

jeremyleung521: I'm leaning to say no. Currently, the RNG just draws from the normal distribution to build the unit cell vectors (and then we transform them back to the lengths/angles to check if they are valid), so we'd have to map the domain of the angles to the unit cell vectors which seems like a lot to figure out, especially because our domain is a little complicated (4 constraints).

We could somehow build it the other way round, but then that seems like a lot of rewrite for very little gain.

From running thousands of these test, the max i would get is about 2/200 failures, so the while loop will just run 2-3 times max, if it ever gets an error the first time.

jeremyleung521:
> Strictly speaking, this adds a new function to the public API. This is useful, so I think it's worth being public, but that does mean we want to be more thorough than a little test-only patch.
> 
> Everything looks fine to me on a read through, but I think a few test cases would be beneficial. Here's what I can come up with:
> 
>     * Pass a known (single) bad box to the function, make sure it fails
> 
>     * Pass a single orthogonal box to the function, make sure it passes
> 
>     * Pass a single valid triclinic box to the function, make sure it passes
> 
>     * Pass a "trajectory" of boxes to the function, with a mix of valid and invalid boxes, make sure it fails at some point
> 
> 
> This is based on my understanding that
> 
>     * This function (and most in this submodule) quack similarly for individual boxes or arrays of multiple boxes
> 
>     * It should pass quickly on orthogonal/cubic boxes
> 
>     * Random data fails some % of the time which is hard to estimate and impossible to rely on
> 
>     * If a trajectory has a mix of valid and invalid boxes, we don't really care about which frames have invalid boxes
> 
> 
> Sorry to be such a stickler here!

It's great that you asked for the tests because... there were problems with the previous code.

Those points in your understanding are correct!

Also... what was numpy thinking:
```
In [19]: np.True_ is True
Out[19]: False

In [20]: np.True_ == True
Out[20]: np.True_
```

mattwthompson:
Will try to get to this tomorrow - please ping me if it slips again

Issue #2033, opened 2025-06-01
PDBTrajectoryFile.write: `header=False` omits all MODEL/ENDMDL tags, causing multi‐frame PDBs to concatenate into a single frame


cabb99:
### Summary

When saving a multi-frame `md.Trajectory` to PDB using `header=False`, the resulting file omits `MODEL`/`ENDMDL` delimiters. Instead of producing distinct frames, all atom records are concatenated, which most tools interpret as a single frame with multiplied atoms.

However, `MODEL`/`ENDMDL` lines are not generally considered part of the PDB "header" (i.e., the CRYST1/REMARK/etc. block). Suppressing them under the `header` flag causes misinterpretation when reading the output back into visualization or analysis tools (e.g., VMD) since the MODEL flag separates the frames.

### Reproduction

```python
import mdtraj as md

traj = md.load('2kpv.pdb')  # 2kpv.pdb contains 10 frames
traj.save('with_header.pdb', header=True)
traj.save('no_header.pdb', header=False)
```

Compare:

```bash
diff with_header.pdb no_header.pdb
```

Also, visual inspection in VMD confirms the issue:

* **Without headers** (left) → 1 frame, 10× atom count* 
* **With headers** (right)→ 10 frames
![without headers](https://github.com/user-attachments/assets/32a4c0d5-4e38-4f1e-b72e-07b271c57243)

---

### Cause

In [`PDBTrajectoryFile.write`](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/pdb/pdbfile.py), the `MODEL`/`ENDMDL` markers are currently gated behind the `header` flag:

```python
if header and modelIndex is not None:
    print("MODEL     %4d" % modelIndex, file=self._file)

# ...

if header and modelIndex is not None:
    print("ENDMDL", file=self._file)
```

This behavior was introduced in [commit `c511338`](https://github.com/mdtraj/mdtraj/pull/1928/commits/c511338d89803ed11deae5635fca2e84b53de5f5), likely to make DSSP run properly, since DSSP fails on some multi-model PDBs.

However, suppressing `MODEL` lines globally breaks multi-frame saving and affects correct parsing of trajectories.

---

### Proposed Fix

Instead of disabling `MODEL` blocks via `header=False`, let `MODEL` behavior be controlled by the number of frames.

---

### Related DSSP Behavior

I also traced the motivation for the original change to the DSSP tests. At the time, they failed when `MODEL` was present. However:

[...]

PR #2034, opened 2025-06-01
Removes header option from pdb writer

cabb99:
Proposed change for issue #2033


jeremyleung521 (approved):
LGTM. 

I think it's quite obvious that I didn't think about multi-model pdbs when I proposed the `header` option :P

Issue #2044, opened 2025-06-09
Numpy 2.3 is breaking a lot of our test after np.integer typecasting removal


jeremyleung521:
Numpy 2.3 is breaking a lot of our tests after `np.integers` dtype typecasting removal. We need to pin or fix it before next release.

```
    def _unique_pairs_equal(cls, a_indices: list[int]) -> NDArray[np.integer]:
>       pairs = np.fromiter(
            itertools.chain.from_iterable(
                itertools.combinations(a_indices, 2),
            ),
            dtype=np.integer,
            count=len(a_indices) * (len(a_indices) - 1),
        )
E       TypeError: Converting 'np.integer' or 'np.signedinteger' to a dtype is not allowed
```

https://github.com/mdtraj/mdtraj/actions/runs/15525079161/job/43703525390

Likely this:

> Converting np.complex, np.integer, np.signedinteger,
> np.unsignedinteger, np.generic to a dtype errors (deprecated
> since 1.19)
> 
> (https://github.com/numpy/numpy/pull/28254)

2.3 release notes:
https://github.com/numpy/numpy/releases/tag/v2.3.0

mattwthompson:
Ideally we would have seen this in CI logs? (This is a standard well above what I hold myself to!) I remember checking some within the past two years for fixes that looked easy. Maybe these were only emitted after 2.0 or something

PR #2045, opened 2025-06-09
change np.integers dtype casting to int

jeremyleung521:
#2044 

While `np.integers` previously casted to `np.complex128`, these are mostly atom indices, so I'm switching to `int` (i.e. `np.int_`), which is 32-bit in 32-bit, 64-bit in 64-bit.


janash (approved):
Thanks for catching this @jeremyleung521 !

mattwthompson (approved):
You mean you've never wanted to look at atom number $\sqrt{-5}$?

Issue #1975, opened 2025-02-06
Performance enhancements for `_geometry.pyx`


jeremyleung521:
Recently while building `mdtraj` from source, I'm getting the following "performance hints". I wonder if it's worth implementing these to get better performance.

```
[ 4/10] Cythonizing mdtraj/geometry/src/_geometry.pyx
performance hint: mdtraj/geometry/src/image_molecules.pxi:15:5: Exception check on 'make_whole' will always require the GIL to be acquired.
Possible solutions:
	1. Declare 'make_whole' as 'noexcept' if you control the definition and you're sure you don't want the function to raise exceptions.
	2. Use an 'int' return type on 'make_whole' to allow an error code to be returned.
performance hint: mdtraj/geometry/src/image_molecules.pxi:36:5: Exception check on 'anchor_dists' will always require the GIL to be acquired.
Possible solutions:
	1. Declare 'anchor_dists' as 'noexcept' if you control the definition and you're sure you don't want the function to raise exceptions.
	2. Use an 'int' return type on 'anchor_dists' to allow an error code to be returned.
performance hint: mdtraj/geometry/src/image_molecules.pxi:68:5: Exception check on 'wrap_mols' will always require the GIL to be acquired.
Possible solutions:
	1. Declare 'wrap_mols' as 'noexcept' if you control the definition and you're sure you don't want the function to raise exceptions.
	2. Use an 'int' return type on 'wrap_mols' to allow an error code to be returned.
```

PR #2049, opened 2025-06-17
speedup some cython functions by enforcing nogil for exceptions (add return code)

jeremyleung521:
#1975 

Not sure if it actually speeds up anything or if this actually did anything, but the "performance hints" are gone during compilation.

### DIFF
--- a/mdtraj/geometry/src/image_molecules.pxi
+++ b/mdtraj/geometry/src/image_molecules.pxi
@@ -12,7 +12,7 @@ cdef extern from "math_patch.h" nogil:
     float roundf(float x)
     float floorf(float x)
 
-cdef void make_whole(float[:,::1] frame_positions,
+cdef int make_whole(float[:,::1] frame_positions,
                 float[:,::1] frame_unitcell_vectors,
                 int32_t[:,:] sorted_bonds) nogil:
     # Fix each molecule to ensure the periodic boundary conditions are not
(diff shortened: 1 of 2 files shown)


mdtraj/geometry/src/image_molecules.pxi
```diff
@@ -12,7 +12,7 @@ cdef extern from "math_patch.h" nogil:
     float roundf(float x)
     float floorf(float x)
 
-cdef void make_whole(float[:,::1] frame_positions,
+cdef int make_whole(float[:,::1] frame_positions,
```
janash, line 15: I think this is likely not the right approach because the function doesn't actually return an `int`. I believe this intention is that we return an int (usually 0 for successful or -1 or some other number if not successful). The function doesn't return anything, so the original `void` is more appropriate.

I think the other fix (adding `noexcept`) is more appropriate for this function. ie

```
cdef void make_whole(float[:,::1] frame_positions,
                float[:,::1] frame_unitcell_vectors,
                int32_t[:,:] sorted_bonds) nogil noexcept:
```

For `int` to be right here, I'd suggest adding a `try/except` with return. The only type of error I think would be possible is `IndexError`, so return 0 if succesful, -1 if not. Then int can stay, but I wouldn't add it if the function never actually returns an `int`.

This is mentioned in the [Cython migration docs](https://cython.readthedocs.io/en/stable/src/userguide/migrating_to_cy30.html#:~:text=There%20is%20an,the%20exception%20check), though it would be nice if it was clearer.

janash: `except -1` seems like the best option here.  To be the most consistent, the functions should probably be changed to `return  0` on success.

janash (approved):

PR #2054, opened 2025-06-25
update pypi upload trigger

janash:
I think we want to use "release" rather than "published" for the event type in order to upload to pypi.

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -78,7 +78,7 @@ jobs:
     environment:
       name: pypi
       url: https://pypi.org/p/mdtraj
-    if: github.event_name == 'published'
+    if: github.event_name == 'release'
     permissions:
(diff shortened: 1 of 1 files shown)


jeremyleung521 (approved):
Ha. I came to the same conclusion and was about to push the same PR.  LGTM.

janash:
Perfect!

janash:
I think I might be able to rerun the workflow

jeremyleung521:
> hmm maybe not. Nevermind! Would be great if you could manually upload and if we could figure out a way to dispatch in case we have a problem again :)

I downloaded the artifacts from https://github.com/mdtraj/mdtraj/actions/runs/15851193684/attempts/1 and uploaded them (after untarring for the wheels). I rebuild the source myself just in case since it was a zip instead of a tar.gz. They should all be up now.

Issue #2056, opened 2025-07-13
Amber prmtop reader enforces standard names


CharlieLaughton:
Hi. When loading an AMBER-flavour PDB format file, one has the option to retain non-standard atom and residue names if desired (`standard_names=False`), however when loading an AMBER prmtop file this option is not available (so for example, "HIE" and "HID" are both converted to "HIS", etc.). Is there a reason for this?  If not, I'd be happy to provide a PR.

PR #2057, opened 2025-07-14
Add support for "standard_names" keyword argument to Amber prmtop reader

CharlieLaughton:
Addressing issue #2056 Amber prmtop reader now supports the "standard_names" keyword argument in the same way as the PDB reader does, so users have a choice over whether non-standard protein atom and residues names in Amber .prmtop files are translated into their standard equivalents, or not.


mdtraj/formats/prmtop.py
```diff
@@ -160,8 +160,11 @@ def load_prmtop(filename, **kwargs):
                         raw_data[flag].append(item.strip())
 
     # Add atoms to the topology
-
     pdb.PDBTrajectoryFile._loadNameReplacementTables()
+    if not kwargs.get("standard_names", True):
```
janash, line 164: I have a concern about using `**kwargs` for this purpose. I noticed that in the original version (before the PR), **kwargs was present in the signature but never used in the function body. I assume this may have been an artifact or intended to keep the interface flexible in case extra arguments were passed and ignored.

That said, for something like `standard_names`, I’d prefer it to be a named argument or at the very least documented in the docstring so users know it exists (and especially because it can potentially change function behavior). This would also make it more consistent with the [`load_pdb` function](https://github.com/mdtraj/mdtraj/blob/e5be3a39023057a2db6ed0966ad9a845e6b7cbb8/mdtraj/formats/pdb/pdbfile.py#L84) that you mentioned in your issue. 

I'd prefer the signature be updated to `load_prmtop(filename, standard_names=True, **kwargs)`  It would be clearer to have it documented in the docstring/showing up in IDEs and autocomplete. 

However, if there is a reason not to do this, please let me know!

janash (changes requested):
Thanks so much for the PR! Here is my review. I'd like to avoid adding to the kwargs if possible.

mdtraj/formats/prmtop.py
```diff
@@ -160,8 +163,11 @@ def load_prmtop(filename, **kwargs):
                         raw_data[flag].append(item.strip())
 
     # Add atoms to the topology
-
     pdb.PDBTrajectoryFile._loadNameReplacementTables()
+    if not standard_names:
+        pdb.PDBTrajectoryFile._residueNameReplacements = {}
```
janash, line 168: My apologies again! 

Upon closer review on this, I have to flag this. `PDBTrajectoryFile._residueNameReplacemets` and `PDBTrajectoryFile._atomNameReplacements` are both class attributes in `PDBTrajectoryFile`. This essentially means that changing them would globally change the name replacement tables for this class *and all instances of the class*, which may produce some very unintended side effects. These attributes should not be modified here.

janash (changes requested):
So sorry to send this back to you again, and thank you for your contribution and edit!

I looked more closely at the `pdb.PDBTrajectoryFile._residueNameReplacements`, etc. These are class attributes meaning they are shared across the class and all instances. They should not be modified as this could unintentionally affect other topologies or files.

In the PDBFile, they take the approach of setting self._standard_names (instance level) with a boolean flag and [checking that](https://github.com/mdtraj/mdtraj/blob/99f60b649ae11f7c9c83ac690f621f35e6edeee3/mdtraj/formats/pdb/pdbfile.py#L663) to determine whether to use the tables. That seems like the more appropriate choice here as well.

mdtraj/formats/prmtop.py
```diff
@@ -186,11 +189,11 @@ def load_prmtop(filename, **kwargs):
 
             # check
             res_name = raw_data["RESIDUE_LABEL"][residue_pointer_dict[index]].strip()
-            if res_name in pdb.PDBTrajectoryFile._residueNameReplacements:
+            if res_name in pdb.PDBTrajectoryFile._residueNameReplacements and standard_names:
```
jeremyleung521, line 192: More nitpicky than anything but I would check for `standard_names` first, then do the res_name check. Probably going to be faster, esp. if the dictionary is large.

jeremyleung521 (approved):

Issue #2063, opened 2025-07-25
Python Freethreading in future Python versions (3.13+)


jeremyleung521:
Starting Python 3.14 ([rc1 out 2025-07-22; final out 2025-10-07](https://peps.python.org/pep-0745/#schedule)), free-threading is no longer experimental. That means future infrastructure (`cibuildwheel`, `conda-forge feedstocks` etc.) will probably build the `py3.1*t` wheels and packages automatically.

This begs the following questions:
- [ ] Should we support it?
  - [ ] If yes:
     - [ ] how do we determine whether we currently support it (out-of-box) or not? 
    - [ ] What changes to the code is needed?
  - [ ] If no:
    - [ ] We need to explicitly block these builds from building

Just pushing the question out there.

PR #2066, opened 2025-08-01
Drop numpy1 from testing/support, not building freethreading wheels, build linux-arm64 wheels

jeremyleung521:
Staying in line with SPEC 0, dropping numpy1 testing and support. Should speed up CI too, due to the need for fewer macos runners.

Also blocked freethreading wheels from building, while I'm at it. See #2063 for more information.

Also building wheels for linux-arm64. We're testing it already, why not build them?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -313,7 +313,7 @@ metadata = dict(
     packages=find_packages(),
     cmdclass={**versioneer.get_cmdclass(), "build_ext": build_ext},
     install_requires=[
-        "numpy>=2.0.0,<3",
+        "numpy~=2.0",
         "scipy",
         "pyparsing",
         "packaging",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ requires = [
     "setuptools",
     "versioneer[toml]",
     "wheel",
-    "numpy>=2.0.0,<3",
+    "numpy~=2.0",
     "Cython>=3.0.0",
 ]
 


.github/workflows/build.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
       - name: Build wheels
         uses: pypa/cibuildwheel@v3.1.2
         env:
-          CIBW_SKIP: "pp* *-musllinux* *-muslinux* *-win32"
+          CIBW_SKIP: "pp* *-musllinux* *-muslinux* *-win32 cp31?t-*"
```
mattwthompson, line 32: Could you walk us through the syntax here? Is `?` acting as a wild card? I looked at https://cibuildwheel.pypa.io/en/stable/options/ briefly

setup.py
```diff
@@ -313,7 +313,7 @@ def geometry_extensions():
     packages=find_packages(),
     cmdclass={**versioneer.get_cmdclass(), "build_ext": build_ext},
     install_requires=[
-        "numpy>=1.25.0,<3",
+        "numpy>=2.0.0,<3",
```
mattwthompson, line 316:
```suggestion
        "numpy=2",
```
A little simpler, am I forgetting a reason this wouldn't work? We ought to be well past pre-release versions being pulled down

mattwthompson (approved):
Dropping NumPy 1.x is good to me (https://scientific-python.org/specs/spec-0000/#2025---quarter-3)

I'm a little surprised we haven't gotten user reports about Linux ARM builds. I thought that platform was used at a few clusters. I guess those users would just let `pip` grab the source and build locally?

setup.py
jeremyleung521: Would have to use `~=2.0`. Single equal sign is conda syntax, iirc. Also can't use a single segment version number for that.

https://packaging.python.org/en/latest/specifications/version-specifiers/#examples
https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release

jeremyleung521: `numpy~=2.0` gives us `numpy>=2.0.0,<3` But i suppose the extra `.0` is giving you grief, haha.

mattwthompson: I can live with that!

PR #2083, opened 2025-09-15
Use repeated bonds in PDB CONECT records to increase bond orders (REMAKE of #1848)

### LINKED ISSUE #1847: Bond order not preserved for PDB files

jeremyleung521:
This is a remake of #1848. Closes #1847. 

Essentially, it allows user to duplicate bonds in the pdb `CONECT` section to represent higher bond orders. This is out of regular PDB spec.

### DIFF
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -1624,8 +1624,8 @@ class Trajectory:
                             f.distance_unit,
                         ),
                         unitcell_angles=self.unitcell_angles[i],
-                        bond_orders=bond_orders,
                         ter=ter,
+                        bond_orders=bond_orders,
                     )
                 else:
                     f.write(
@@ -1637,8 +1637,8 @@ class Trajectory:
                         self.topology,
                         modelIndex=i if self.n_frames > 1 else None,
                         bfactors=bfactors[i],
-                        bond_orders=bond_orders,
                         ter=ter,
+                        bond_orders=bond_orders,
                     )
 
     def save_cif(self, filename, force_overwrite=True, bfactors=None, ter=True, header=True):


janash:
Is this ready for a review, or are things still being changed? (I see failing tests)

mdtraj/core/topology.py
```diff
@@ -1905,8 +1905,13 @@ def float_to_bond_type(bond_float: float) -> Singleton | None:
     """
     all_bond_types = [Single, Double, Triple, Aromatic, Amide]
     for bond_type in all_bond_types:
-        if float(cast(Singleton, bond_type)) == float(bond_float):  # explicitly cast to Singleton to avoid mypy error
-            return cast(Singleton, bond_type)
+        try:
+            if float(cast(Singleton, bond_type)) == float(bond_float):
+                # explicitly cast to Singleton to avoid mypy error
+                return cast(Singleton, bond_type)
+        except TypeError:
```
janash: Are we just trying to catch `bond_float=None` here? If so, it may be faster/cleaner to check for this upfront and return `None` instead of going through the loop.

Besides this, this function seems like it would be cleaner by using a dictionary mapping rather than the current loop.

```python
def float_to_bond_type(bond_float: float | None) -> Singleton | None:
    
bond_map: dict[float, Singleton] = {
        float(Single): Single,
        float(Double): Double,
        float(Triple): Triple,
        float(Aromatic): Aromatic,
        float(Amide): Amide,
    }
    try:
        return bond_map.get(float(bond_float), None)
    except (TypeError, ValueError):
        return None
```

mdtraj/formats/pdb/pdbfile.py
```diff
@@ -306,6 +318,7 @@ def write(
         unitcell_lengths=None,
         unitcell_angles=None,
         bfactors=None,
+        bond_orders=False,
```
janash: Same request as above to move `bond_orders` to the end of arguments list.

mdtraj/formats/pdb/pdbfile.py
```diff
@@ -704,13 +727,32 @@ def _read_models(self):
...
                         connectBonds.append((atomByNumber[i], atomByNumber[j]))
+
             if len(connectBonds) > 0:
-                # Only add bonds that don't already exist.
-                existingBonds = {(bond.atom1, bond.atom2) for bond in self._topology.bonds}
-                for bond in connectBonds:
-                    if bond not in existingBonds and (bond[1], bond[0]) not in existingBonds:
-                        self._topology.add_bond(bond[0], bond[1])
-                        existingBonds.add(bond)
+                existingBonds = [(bond.atom1, bond.atom2) for bond in self._topology.bonds]
```
janash: Is there a reason to keep `existingBonds` in both a list and set version (`existinBonds_set`)? I'm not entirely seeing why the list version is needed.

janash (changes requested):
See comments on keyword argument placement, I think it would be best to place the new keyword arugments at the end of functions.

Beyond this, I think there is a bug in building the bond orders from connect records. The `ala_ala_ala.pdb` case seems to work correctly, but `imatinib.pdb` as written would result in a bug. It seems that the bond order between (1,21) should be two based on the CONECT record (line [71](https://github.com/jeremyleung521/mdtraj/blob/5c46c1428aee3ec5311ec70ce820983db61cf7ce/tests/data/imatinib.pdb#L71)), but this bond is repeated later in the PDB as (21, 1) (on line [97](https://github.com/jeremyleung521/mdtraj/blob/5c46c1428aee3ec5311ec70ce820983db61cf7ce/tests/data/imatinib.pdb#L97)) without the repeated CONECT. When adding a bond in the topology, the atoms are always sorted, meaning that this last record would overwrite the previously determined BO 2.

This makes this bond order equal to 1 instead of 2. The logic of counting the conect records currently will overwrite with the "last seen"  bond order rather than the highest one. 

Questions:

* Is that the intended behavior?
* Are there real cases where this could be "asymmetric"?

Small test script:

```python
import mdtraj as md
from pathlib import Path

pdb = Path("tests/data/imatinib.pdb").resolve()

def bond_info(traj):
    for bond in traj.topology.bonds:
        serials = {bond.atom1.serial, bond.atom2.serial}
        if serials == {1, 21}:
            return bond.order, bond.type
    raise ValueError("Bond between serials 1 and 21 not found")

traj_with_orders = md.load_pdb(pdb, bond_orders=True)
traj_without_orders = md.load_pdb(pdb, bond_orders=False)

print("bond_orders=True :", bond_info(traj_with_orders))
print("bond_orders=False:", bond_info(traj_without_orders))
```

Results in:
```
bond_orders=True : (1, Single)
bond_orders=False: (None, None)
```

Based on the line 71 CONECT entry I’d expect `(2, Double)`

mdtraj/core/topology.py
```diff
@@ -1887,7 +1887,7 @@ def __float__(self) -> float:
 Amide: Singleton = Amide()  # type: ignore[no-redef]
 
 
-def float_to_bond_type(bond_float: float) -> Singleton | None:
+def float_to_bond_type(bond_float: float | None) -> Singleton | None:
```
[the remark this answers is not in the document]
jeremyleung521: This was missing before. It would previously error out when `None` is passed via `float_to_bond_type(None)` at `float(None)`. 

https://github.com/jeremyleung521/mdtraj/blame/3b14f0f306f9e2fc36a527d5afb5ca4a22cf46d8/mdtraj/core/topology.py#L1908

jeremyleung521:
> See comments on keyword argument placement, I think it would be best to place the new keyword arugments at the end of functions.
> 
> Beyond this, I think there is a bug in building the bond orders from connect records. The `ala_ala_ala.pdb` case seems to work correctly, but `imatinib.pdb` as written would result in a bug. It seems that the bond order between (1,21) should be two based on the CONECT record (line [71](https://github.com/jeremyleung521/mdtraj/blob/5c46c1428aee3ec5311ec70ce820983db61cf7ce/tests/data/imatinib.pdb#L71)), but this bond is repeated later in the PDB as (21, 1) (on line [97](https://github.com/jeremyleung521/mdtraj/blob/5c46c1428aee3ec5311ec70ce820983db61cf7ce/tests/data/imatinib.pdb#L97)) without the repeated CONECT. When adding a bond in the topology, the atoms are always sorted, meaning that this last record would overwrite the previously determined BO 2.
> 
> This makes this bond order equal to 1 instead of 2. The logic of counting the conect records currently will overwrite with the "last seen" bond order rather than the highest one.
> 
> Questions:
> 
> * Is that the intended behavior?
> * Are there real cases where this could occur?
> 
> Small test script:
> 
> ```python
> import mdtraj as md
> from pathlib import Path
> 
> pdb = Path("tests/data/imatinib.pdb").resolve()
> 
> def bond_info(traj):
>     for bond in traj.topology.bonds:
>         serials = {bond.atom1.serial, bond.atom2.serial}
>         if serials == {1, 21}:
>             return bond.order, bond.type
>     raise ValueError("Bond between 1 and 21 not found")
> 
> traj_with_orders = md.load_pdb(pdb, bond_orders=True)
> traj_without_orders = md.load_pdb(pdb, bond_orders=False)
> 
> print("bond_orders=True :", bond_info(traj_with_orders))
> print("bond_orders=False:", bond_info(traj_without_orders))
> ```
> 
> Results in:
> 
> ```
> bond_orders=True : (1, Single)
> bond_orders=False: (None, None)
> ```
> 
> Based on the line 71 CONECT entry I’d expect `(2, Double)`

Thanks for reviewing! 

The "last seen" behavior indeed is not the intended behavior. PDB spec indicates that you can have a bond listed twice in CONECT ((a,b) and (b,a)) and it should be counted as a single bond. Alternatively, you can also list it once and it should still be counted as a single bond. In that vein, with repeating bonds, I think people could list multiple bonds only once in one order (like in `imatinib.pdb`), or both.

So the current behavior is "largest bond order" priority. If a bond order is listed multiple times, it will be updated based on the most consecutive times you list it. That way we don't over count and accidentally overwrite a higher bond order with 1, like seen in your example.

e.g.
```
CONECT
A B B B
B A A
```
would be read as a triple bond

janash:
> Overall, I think there is an overall problem with `bond_type` and `bond_order` that we should clear up/define clearly. They are obviously related but not always in sync. I'm not sure of the differences in use cases between the two attributes. For example, mol2 reader will only read `bond_type`, leaving `bond_order` 0/None. Here, in this PR, we're rewriting both concurrently from pdb. I'm not sure what the other formats are doing, probably something different.
> 
> The `load_mol2` test stumped me a bit because of the `bond_order`/`bond_type` mismatch between formats. [ec42149#diff-76406bde6a6d499fae412a671dcc9daeb77c1536df2546b5eda2a40814d1635dR45-R48](https://github.com/mdtraj/mdtraj/commit/ec4214943a37cc682a31d77104452ccb1f3de028#diff-76406bde6a6d499fae412a671dcc9daeb77c1536df2546b5eda2a40814d1635dR45-R48)
> 
> Related, I think we forced `bond_order` to be integers in #2007. I think @Lnaden implemented both attributes 8-10 years ago, so if he could clarify his rational/use case back then, that'll be great also.

These are all great points! I think that there could be different use cases for bond order and bond type, and I suppose you might not necessarily be using both at the same time. Bond order should be allowed to be float. The simple bond orders like 1 and 1.5 are from molecular orbital theory, though I've seen float (continuous) bond order from other methods, particularly in QM. I am not sure how relevant those are to MDTraj, but they do exist. I also think it would be simpler to go bond type -> inferred order than the other way around (if someone has continuous bond orders, those don't map neatly to a category, but "single" bond maps easily to a bond order of 1).

As far as forcing bond orders to be `int` in #2007, I see formal charge discussion there, but not bond order. Perhaps I'm missing something. Regardless, bond order should be allowed to be non-integer in the case that someone wants to set this to some value.

janash (approved):
I think this PR is ready to go now with the bugs fixed. A slight comment is that I would like to see the highest order priority tested, but that's not blocking.

janash:
These are all great points Jeremy, and I'm glad you've found all of these subtle issues. It looks like bond order has been cast as int since first implentation(8ish years ago by @Lnaden, as you pointed out). I'd also be curious about that choice (especially given that we have `float_to_bond_type` that maps 1.25 to amide and 1.5 to aromatic). I am totally unsure what downstream effects would be, but it is my intuition that bond order should be allowed to be a float.

As far as `mol2` goes those are really good things to find about the format. When I've used mol2 it's been explicitly because it allows you to store the partial charges. Formal charges and partial charges are different things, and I have thought (especially when adding formal charges) if we should *also* add partial charge "capabilities" to the topology. I would guess that the charges are dropped from the mol2 the same way the formal charges used to be dropped when reading from a PDB.

janash:
Ready for merge?

PR #2088, opened 2025-10-02
Updating macos-13 runner to macos-15-intel

jeremyleung521:
macos-13 runners (for x86-64) are closing down December 4th, 2025
https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/

This pushes all our runners to macos-15-intel, which will be the last macOS x86-64 runners (since Apple dropped support for x86-64). These runners will last until Fall 2027.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -25,7 +25,7 @@ jobs:
       matrix:
         os:
           - macOS-latest  # arm64
-          - macOS-13  # x86-64
+          - macOS-15-intel  # x86-64
           - ubuntu-latest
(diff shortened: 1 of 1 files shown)


mattwthompson:
Not opposed to this, but we need to think about dropping the platform altogether

jeremyleung521:
> Not opposed to this, but we need to think about dropping the platform altogether

Personally, I think we can drop testing when apple drops rosetta 2 support, but that's just me :) I think bug fixes/security patches are until 2028 for intel macs.

mattwthompson (approved):

Issue #1930, opened 2024-09-23
Preserve bond orders when writing out/reading in HDF5 trajectory files


SHZ66:
Hi Developers,

I noticed that the current HDF5 file interface doesn’t appear to support storing or reading bond orders:
[Link to Code](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/hdf5.py#L309-L368)

Are there any plans to add this functionality?

Thank you,
She

jeremyleung521:
Hi She!

We don't have any plans right now on saving bond orders in the HDF5, as that probably requires modifying the HDF5 file specs. We're currently more concerned about getting bugs/getting dependencies up to date (Cython, bugs, numpy2) than features, but if you (or someone) want to contribute, we're happy to review.

We'll add this to the wish list, since preserving bond order information appears to be something people want (see #1848).

PR #2101, opened 2025-11-03
Add bond metadata (order and type) and formal charge to HDF5

janash:
This PR addresses #1930 - it adds **bond order**, **bond type**, and **atom formal charge** to the topology data stored in HDF5. The goal is to enable full round-tripping of a topology: if a user saves a topology to HDF5 and loads it again, they should get a topology with the same information. The current implementation of the HDF5 does not allow this, particulary for bond information.

Currently, the HDF5 format only stores which atoms are bonded (pairs of indices) and does not store the new atom formal charge field. As a result, a user might save an MDTraj trajectory as HDF5, reload it, and get a different topology object (bond order and bond type are lost on write). This PR fixes that by adding a new HDF5 field, `bond_metadata`, which stores metadata for each bond in the same order as the `bonds` list. This keeps the format more backward compatible by keeping the existing `bonds` representation.

For atoms, this PR adds an optional `formal_charge` field. If at least one atom has a formal charge, the field is written; if all formal charges are `None`, it is omitted. This avoids having adding `formal_charge: None` unneccessarily to all files.

When writing HDF5 files, these additions are enabled with the keyword argument `bond_metadata=True` to the `HDF5File` class. When using top level `save` or `load` functions, these fields will now be written by default. Note that the `bond_metadata` field is a special field for the HDF5 file format. Its information is loaded into bond objects on the MDTraj trajectory - it's just use for reconstruction.

**Resulting JSON structure (simplified):**

```
{
  "chains": [
    {
      "residues": [
        {
          "atoms": [
            {"index": …, "name": …, "element": …, **"formal_charge": …**},
            …
          ]
        },
        …
      ]
    },
    …
  ],

  "bonds": [
    [i0, j0],
    [i1, j1],
    …
  ],

  "bond_metadata": [
    {"order": …, "type": …},
    {"order": …, "type": …},
    …
  ]
}
```

### DIFF
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -1507,7 +1507,7 @@ class Trajectory:
         mode : str, default='w'
             The mode in which to save the file. 'w' will overwrite any existing
             file, 'a' will append to an existing file.
-        bond_metadata: bool, default=True
+        bond_metadata : bool, default=True
             Flag to save bond order and type in the HDF5.
         """
         # check if savemode is valid (only "w" or "a" are allowed)


jeremyleung521 (commented):
This is what I tried, which wasn't working. I think we need to expand the `bond_metadata` argument to both the `save()` and `load()` functions for hdf5. I highlighted those places with comments.


```
import mdtraj
a = mdtraj.load('tests/data/4py5.cif')
a.top._bonds  # bond_orders and bond_types exist
a.save('test.h5') 
a.top.bond_metadata # attribute does not exist
a = mdtraj.load('test.h5')
a.top._bonds # bond_orders and bond_types does not exist
```

Problems:
`bond_metadata` is a class attribute of the HDF5TrajectoryFile, which is typically not explicitly accessed/used by the user. It is often discarded after loading/writing.
```
In [3]: a.save('test.h5', bond_metadata=True)
TypeError                                 Traceback (most recent call last)
Cell In[3], line 1
----> 1 a.save('test.h5', bond_metadata=True)

File ~/Documents/Important/Pitt/git/mdtraj/mdtraj/core/trajectory.py:1493, in Trajectory.save(self, filename, **kwargs)
   1486     raise OSError(
   1487         f"Sorry, no saver for filename={filename} (extension={extension}) "
   1488         "was found. I can only save files "
   1489         f"with extensions in {savers.keys()}",
   1490     )
   1492 # run the saver, and return whatever output it gives
-> 1493 return saver(filename, **kwargs)

TypeError: Trajectory.save_hdf5() got an unexpected keyword argument 'bond_metadata'
```

```
In [4]: a.load('test.h5', bond_metadata=True)
TypeError                                 Traceback (most recent call last)
Cell In[4], line 1
----> 1 a.load('test.h5', bond_metadata=True)

File ~/Documents/Important/Pitt/git/mdtraj/mdtraj/core/trajectory.py:1430, in Trajectory.load(filenames, **kwargs)
   1415 @staticmethod
   1416 # im not really sure if the load function should be just a function or a method on the class
   1417 # so effectively, lets make it both?
   1418 def load(filenames, **kwargs):
   1419     """Load a trajectory from disk
   1420
   1421     Parameters
   (...)   1428     As requested by the various load functions -- it depends on the extension
   1429     """
-> 1430     return load(filenames, **kwargs)

File ~/Documents/Important/Pitt/git/mdtraj/mdtraj/core/trajectory.py:444, in load(filename_or_filenames, discard_overlapping_frames, **kwargs)
    434 filename_or_filenames = filename_or_filenames[1:]  # ignore first file
    435 try:
    436     # this is a little hack that makes calling load() more predictable. since
    437     # most of the loaders take a kwargs "top" except for load_hdf5, (since
   (...)    442     # TODO make all the loaders accept a pre parsed topology (top) in order to avoid
    443     # this part and have a more consistent interface and a faster load function
--> 444     t = loader(tmp_file, **kwargs)
    446 except TypeError as e:
    447     # Don't want to intercept legit
    448     # TypeErrors
    449     if "got an unexpected keyword argument 'top'" not in str(e):

TypeError: load_hdf5() got an unexpected keyword argument 'bond_metadata'
```

janash:
@jeremyleung521 - this is ready for another review. I've edited the description for the PR. Here are some important additional notes:

* I did not add `bond_metadata` on `save/load` on the trajectory. The behavior in this PR Is that bond metadata will now be saved automatically when using `save` to HDF5. IMO this is more expected behavior (if you save a topology, all data will be preserved)
* The `save_hdf5` (on topology) has an argument to save bond metadata - it is set to `True` by default.
*  No flag is needed on `load_hdf5` method. It uses `HDF5TrajectoryFile`, which will read `bond_metadata` by default if present. If not present, it just skips. This way, the field will be loaded if present and skipped if not without a flag.

PR #2103, opened 2025-11-06
Addressing some issues with `save_cif`

darianyang:
I recently encountered an error when using `traj[frame].save_cif()`. Using `save_pdb()` there is no problem, but during the header/bond info writing of the `cif` file, I came across `key_errors` for chain and residue IDs.

Tested with `mdtraj-1.11.0`.

My code to reproduce this currently looks something like the following:
``` Python
import mdtraj as md
traj = md.load_dcd("traj.dcd", top="traj.pdb")
traj[0].save_cif("frame.cif")
```

The issue seems to stem from the following dicts: https://github.com/mdtraj/mdtraj/blob/9ca89d82caf6a89d96ca464dba1ca394c041de60/mdtraj/formats/pdbx/pdbxfile.py#L489-L490

When these are indexed later on:
https://github.com/mdtraj/mdtraj/blob/9ca89d82caf6a89d96ca464dba1ca394c041de60/mdtraj/formats/pdbx/pdbxfile.py#L507-L520

I am getting a `key_error` since, e.g. `atom1.residue.chain` is a mdtraj chain object, but for some reason the memory address does not match with either of the chain objects in `chainIds`. A similar issue happens when indexing `resIds`. 

This initial issue lead me to update the following:
* I removed the header arg from the `save_cif` trajectory method, this was not being used anyway and there needs to always be a header to loop the atomic coordinate data being stored.
* In the same `save_cif` method I also exposed and propagated the `keepIds` arg through to the write methods, allowing for user specification.
* With the `keepIds` arg, I noted that when setting to `True`, there was attribute errors when trying to access the topology chain and residue IDs, I updated these attribute calls to what I think are the corrected ones.
    * I also updated the `insertionCode` checking to account for a missing `insertionCode` attribute.
* I then addressed my original `key_error` issue by accessing the chain and res IDs directly from the `atom1` and `atom2` objects, instead of indexing the `chainIds` and `resIds` dicts. This fixes my issue but at cost; now the `keepIds` functionality is not being used in the bond writing part of the header writing in `cif` file saving.

To truly address the issue, some more investigation is needed. Perhaps the `bond.atom.residue.chain` or the topology-based chain objects are getting overridden somewhere? I'm also curious why this issue hasn't come up before, I'm loading in an `OpenMM` generated `dcd` traj file and a `pdb` file for the topology. Perhaps there is something with this specific combination?

Apologies for the unfinished PR, I thought it would be an easy fix but got a bit stuck. If anyone has any thoughts or suggestions here I would greatly appreciate any help.


cabb99:
Hi @jeremyleung521, I could reproduce the error and I have reviewed the code. I also a test based on @darianyang files. The code is in [my fork](https://github.com/cabb99/mdtraj/tree/pr/darianyang/2103). I added a pull request to @darianyang 's branch

jeremyleung521:
Should be good after https://github.com/darianyang/mdtraj/pull/2 is merged. I used the alanine dipeptide files instead since I was finally able to replicate with existing files.

Interesting that the error only occurs when you select a frame and not the whole thing. I was previously trying the latter so couldn't replicate the bug. e.g.:
```
traj[0].save('foo.cif')  # Errors out
traj.save('foo.cif')  # Does not error out
```

jeremyleung521:
> LGTM. I also tested things locally using the ala dipeptide files with same results.
> 
> The frame vs traj saving discrepancy is interesting... it looks like the only difference is just the write loop of one frame vs multiple, so it should work the same, but maybe this is from a difference in memory address when pulling a frame vs a traj?
> 
> Ah, that's it I think. I ran a quick test using the original implementation (before the fixes): image
> 
> You can see that the memory addresses for the `traj` vs `traj[0]` `ChainIds` dicts are different, with the `traj` `ChainIds` dict being correctly indexable.
> 
> The indexing is done using the chain references from `topology.bonds.atom.residue.chain`, which is the same between `traj` and `traj[0]`.
> 
> The `ChainIds` dict is built using `topology.chains` which differs between `traj` and `traj[0]`. This difference is why the error happens only with `traj[n]`.

The [topology is deep-copied](https://github.com/mdtraj/mdtraj/blob/main/mdtraj/core/trajectory.py#L1298) during slice and so that might explain the different memory address. During that copy process, [the topology is effectively remade with new atoms/residues/chains](https://github.com/mdtraj/mdtraj/blob/main/mdtraj/core/topology.py#L248-L269), leading to a mismatch/KeyError.

We might want to double check the topology copy process in another PR/Issue, but that's a little outside the scope of this PR so I'll merge this in right now.

Issue #2116, opened 2026-01-05
Renamed methods in PyParsing 3.x


mattwthompson:
It turns out we've been using aliases for some time. As of November, these are now warnings

> Beginning with 3.3.0, pyparsing emits DeprecationWarning for many pre-PEP8 method names (for example, ParserElement.parseString). The PEP8-compliant names were introduced in 3.0.0 (August 2021); the legacy names remain as aliases but are planned to be removed in pyparsing 4.0 (no earlier than 2026).

https://pyparsing-docs.readthedocs.io/en/latest/whats_new_in_3_3.html#planned-deprecations-and-guidance-for-migration

PR #2117, opened 2026-01-05
Update PyParsing deprecations

mattwthompson:
See #2116


mdtraj/core/selection.py
jeremyleung521:
```suggestion
            return [(kw, klass.n_terms, getattr(OpAssoc, klass.assoc), klass) for kw in kws]
```

mattwthompson:
The web UI won't show me these comments in the file view _or_ let me commit them _or_ let me add them to a patch

PR #2124, opened 2026-01-15
Fix PyPI upload GitHub Action Process

jeremyleung521:
Use `pattern:` instead of `name:`

Bug from #1980. We might be able to manually dispatch this and it should hopefully push the right wheels up. If not we can just manually upload the wheels.

Last release it didn't trigger properly so it never got to this point...

Added an extra conditional so forks won't try to upload either.

EDIT: I manually uploaded the wheels and source.


mattwthompson (approved):
Seems good, mostly guessing since this is hard to properly run tests of

PR #2137, opened 2026-02-10
PDB: take care when CRYST1 is specified but empty fields for unit cell angles/lengths

jeremyleung521:
fixes #2136


janash (commented):
This seems reasonable to me. I pulled down the reference file mentioned in the issue and confirmed that it (1) fails on main and (2) loads successfully with this PR.

The main thing causing the behavior change is the addition of the `try/except` in the `CRYST1` case. 

One thing I *think* I'm not seeing is an associated unit test. Does this have to do with `cassettes.tar.gz` that was changed on this PR?

janash (approved):
Sorry about the delay!

PR #2159, opened 2026-05-26
Guard against empty atom indices producing NaNs in superpose

badisa:
* Minor issue, but it is good to avoid the compute and to raise a meaningful error

### DIFF
--- a/tests/test_rmsd.py
+++ b/tests/test_rmsd.py
@@ -321,16 +321,14 @@ def test_superpose_with_empty_atom_raises_exception():
     n_frames = 5
     n_atoms = 20
 
-    xyz = rng.standard_normal((n_frames, n_atoms, 3)).astype(np.float32)
-    original = md.Trajectory(xyz=xyz, topology=None)
+    xyz = rng.standard_normal((n_frames, n_atoms, 3), dtype=np.float32)
+    traj = md.Trajectory(xyz=xyz, topology=None)
 
-    # Superpose with empty atom_indices list, previously this would lead to NaN positions
-    query = md.Trajectory(xyz=xyz.copy(), topology=None)
     with pytest.raises(ValueError, match="Number of atom indices must be greater than 0"):
-        query.superpose(original, frame=0, atom_indices=[])
+        traj.superpose(traj, frame=0, atom_indices=[])
 
     # Super posing with valid indices should behave as expected afterwards
-    query.superpose(original, frame=0, atom_indices=[0])
+    traj.superpose(traj, frame=0, atom_indices=[0])
 
     # Coordinates should be finite after superposing
-    assert np.all(np.isfinite(query.xyz))
+    assert np.all(np.isfinite(traj.xyz))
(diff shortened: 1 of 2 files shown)


tests/test_rmsd.py
```diff
@@ -314,3 +314,23 @@ def test_rmsd_atom_indices_vs_ref_indices():
...
+def test_superpose_with_empty_atom_raises_exception():
+    # Test that superpose with an empty atom_indices list produces NaN coordinates
+    n_frames = 5
+    n_atoms = 20
+
+    xyz = rng.standard_normal((n_frames, n_atoms, 3)).astype(np.float32)
+    original = md.Trajectory(xyz=xyz, topology=None)
+
+    # Superpose with empty atom_indices list, previously this would lead to NaN positions
+    query = md.Trajectory(xyz=xyz.copy(), topology=None)
```
jeremyleung521: Not sure why you need to make a copy of xyz/traj, unless you meant to run `np.testing.assert_array_equal(original.xyz, query.xyz)` for a stronger test to see if original's xyz is unmodified? Otherwise, You can just superpose the traj on itself.

i.e.
`traj.superpose(traj, frame=0, atom_indices=[])`

tests/test_rmsd.py
```diff
@@ -314,3 +314,23 @@ def test_rmsd_atom_indices_vs_ref_indices():
...
     md.rmsd(trj_2, trj_1, atom_indices=[1], ref_atom_indices=[0])
     # if this don't fail then it's good no matter the result
+
+
+def test_superpose_with_empty_atom_raises_exception():
+    # Test that superpose with an empty atom_indices list produces NaN coordinates
+    n_frames = 5
+    n_atoms = 20
+
+    xyz = rng.standard_normal((n_frames, n_atoms, 3)).astype(np.float32)
```
jeremyleung521:
```suggestion
    xyz = rng.standard_normal((n_frames, n_atoms, 3), dtype=np.float32)
```

[A little bit more efficient](https://github.com/numpy/numpy/blob/cf16ec13dc456c3debccc46578a4f2535be2ea04/numpy/random/_generator.pyx#L1190) by creating the `np.float32` array directly. Instead of casting `np.float64 -> np.float32` with `astype`.

jeremyleung521 (approved):
Welcome and thank you for the contribution. Good catch to make sure `trajectory.xyz` doesn't get spoiled with NaN.

Comments are just some nit-picky things.

tests/test_rmsd.py
badisa: Fixed in https://github.com/mdtraj/mdtraj/pull/2159/commits/481e49ce79f4b2e4831f0fa5eed8c53aeba935af

tests/test_rmsd.py
badisa: Removed the copy in https://github.com/mdtraj/mdtraj/pull/2159/commits/481e49ce79f4b2e4831f0fa5eed8c53aeba935af

badisa:
Is there anything else I need to do for this to be merged?

Issue #2166, opened 2026-06-23
Numpy 2.5.0 is causing instabilities/errors with eigen calculations


jeremyleung521:
Something changed and we're getting errors in the arm-versions of the CI runs.

https://github.com/mdtraj/mdtraj/actions/runs/28000906109

Specifically, stuff that are supposedly 0s are reported as very, very small floats:

> Mismatch at indices:
>  [1, 0]: (2.7078717794979004e-32+0j) (ACTUAL), 0 (DESIRED)
>  [2, 0]: (2.7078717794979004e-32+0j) (ACTUAL), 0 (DESIRED)
>  [3, 2]: (2.7078717794979004e-32+0j) (ACTUAL), 0 (DESIRED)

related to numpy 2.5.0 released yesterday? (EDIT: Confirmed)

mattwthompson:
Let's pin in CI for now

Worth a `.post1` sort of release adding in this constraint?

jeremyleung521:
There's been too many commits since `1.11.1.post1`, so I made a separate branch for `1.11.1.post2` (https://github.com/mdtraj/mdtraj/tree/1.11.1.post2) which contains just the numpy version pin change, branched from the `1.11.1.post1` tag. This branch will freeze after publishing `post2`.

The main branch contains #2168 and already contains the change, so it'll live on.

OR... we could just release `1.11.2` eventually.

ctk3b:
Potential culprit? 
* https://numpy.org/devdocs/release/2.5.0-notes.html#linalg-eig-and-linalg-eigvals-now-always-return-complex-arrays
* https://github.com/mdtraj/mdtraj/blob/main/mdtraj/geometry/order.py#L303

jeremyleung521:
`1.11.1.post2` and the conda-forge 1.11.1 build 2 are out. The pinning of older conda packages will be out when the conda-forge core team approves.

mattwthompson:
This has been out for about three months so I'm going to take a stab at fixing this

PR #2168, opened 2026-07-06
Pin numpy2.5

jeremyleung521:
#2166 

Pinning Numpy to <2.5.0 until we can fix it.


mattwthompson (approved):
LGTM % green checks coming back

PR #2172, opened 2026-07-20
Add bridge_anchors option to image_molecules

jgullingsrud:
image_molecules keeps separate anchor molecules together with a single-closest-contact minimum-image heuristic, which picks the wrong periodic image when the assembled complex spans more than half the box along an axis, splitting elongated multi-chain complexes (e.g. protein dimers) across the periodic boundary.

The new bridge_anchors option treats all anchor molecules as a single connected complex: they are stitched together at their real-space interface contacts (measured on a made-whole reference) so make_whole reassembles the whole complex deterministically, bypassing the per-pair minimum-image heuristic. _sort_bonds now accepts extra_bonds so the bridges are folded into the existing spanning-tree make_whole path.

Defaults to False, so existing behaviour is unchanged.

### DIFF
--- a/mdtraj/core/trajectory.py
+++ b/mdtraj/core/trajectory.py
@@ -2337,13 +2337,19 @@ class Trajectory:
     def _have_unitcell(self):
         return self._unitcell_lengths is not None and self._unitcell_angles is not None
 
-    def _sort_bonds(self):
+    def _sort_bonds(self, extra_bonds=None):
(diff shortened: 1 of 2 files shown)


jgullingsrud:
Hello - does anyone have feedback on this PR?

mattwthompson (approved):
Comments were helpful for reviewing, thanks

